PR #4, opened 2025-06-30
test: add single client test

AndyDai-nv:
Add single client test to test simple downloading functionality and unblock testing until the race condition fix is merged

### DIFF
--- a/model_express_server/src/database.rs
+++ b/model_express_server/src/database.rs
@@ -299,6 +299,60 @@ impl ModelDatabase {
 
         Ok((downloading, downloaded, error))
     }
+
+    /// Atomically attempt to claim a model for downloading using compare-and-swap semantics
+    /// Returns the current status of the model:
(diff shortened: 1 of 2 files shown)


model_express_client/src/bin/test_single_client.rs
```diff
@@ -0,0 +1,140 @@
...
+    println!("Client: Model downloaded in {:?}", start.elapsed());
+
+    println!("Client completed in {:?}", start_time.elapsed());
+    println!("TEST PASSED: Model was downloaded successfully");
+
+    Ok(())
+}
+
+// Function to test provider selection
+#[allow(dead_code)]
```
nicolasnoble: Why are we having this unused function here? Shouldn't it be called alongside `run_fallback_test` from `main` or something? Or maybe we should have this from a CLI switch, to select which test to run?

AndyDai-nv: Updated. We can use separate PR for provider tests once more providers sources are supported

PR #6, opened 2025-07-03
Add Model Express Cache CLI 

KavinKrishnan:


model_express_client/src/lib.rs
```diff
@@ -46,13 +49,79 @@ impl Client {
...
+        
+        // First try to download via server
+        match self.request_model_with_provider(model_name, provider).await {
+            Ok(()) => {
+                info!("Model {} pre-loaded successfully via server", model_name);
+                Ok(())
+            }
+            Err(e) => {
+                // Fallback to direct download
+                info!("Server unavailable, pre-loading model {} directly", model_name);
```
nicolasnoble: Please log the error in `e` for debugging.

model_express_client/src/cache_config.rs
```diff
@@ -0,0 +1,486 @@
...
+        match self.total_size {
+            size if size >= GB => format!("{:.2} GB", size as f64 / GB as f64),
+            size if size >= MB => format!("{:.2} MB", size as f64 / MB as f64),
+            size if size >= KB => format!("{:.2} KB", size as f64 / KB as f64),
+            size => format!("{} bytes", size),
+        }
+    }
+
+    /// Format individual model size as human readable string
+    pub fn format_model_size(&self, model: &ModelInfo) -> String {
```
nicolasnoble: Isn't it just a copy/paste of the above?

Cargo.lock
```diff
@@ -1223,11 +1223,14 @@ dependencies = [
 name = "model_express_client"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
```
nicolasnoble, line 1329: We might want to refactor a lot of the existing code if we start using `anyhow` :) I'm not against it, just that it creates a bit of a imbalance right now.

model_express_client/src/cache_config.rs
```diff
@@ -0,0 +1,486 @@
+use std::path::{Path, PathBuf};
```
nicolasnoble: Been thinking a bit, but shouldn't this be in the model_express_common crate instead? The server eventually should need it too in order to store files in a common location between the server and the client, right?

model_express_client/src/bin/cache_cli.rs
```diff
@@ -0,0 +1,332 @@
+use clap::{Parser, Subcommand};
```
nicolasnoble: Same comment as with the cache_config.rs file. Maybe the CLI tools should all reside in the common crate? Crates can be both libraries and be CLI tools.

model_express_client/src/lib.rs
```diff
@@ -46,13 +49,79 @@ impl Client {
...
+
+    /// Set cache configuration
+    pub fn set_cache_config(&mut self, cache_config: CacheConfig) {
+        self.cache_config = Some(cache_config);
+    }
+
+    /// List cached models
+    pub fn list_cached_models(&self) -> CommonResult<CacheStats> {
+        let cache_config = self.cache_config.as_ref()
+            .ok_or_else(|| model_express_common::Error::Server("Cache not configured".to_string()))?;
```
nicolasnoble: All of these probably should just be `.ok_or` instead of `.ok_or_else`, as there's quite no capture happening there.

KavinKrishnan: Looking into this, the .ok_or adds unnecessary string construction even if error is not hit, which is why .ok_or_else is being used here

model_express_client/src/cache_config.rs
KavinKrishnan: Fixed, ya it was to handle the different model object :) 

Refactored into cleaner struct with re-use

model_express_client/src/cache_config.rs
KavinKrishnan: Refactored

model_express_client/src/bin/cache_cli.rs
KavinKrishnan: Refactored

Cargo.lock
KavinKrishnan: Refactored

model_express_client/src/bin/cache_cli.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    // Check for --cache-path argument
+    let cache_path = args
+        .iter()
+        .position(|arg| arg == "--cache-path")
+        .and_then(|i| args.get(i + 1))
+        .map(|s| PathBuf::from(s));
+
+    // Check for --server-endpoint argument
+    let server_endpoint = args
```
nicolasnoble: This variable is fully unused...

PR #11, opened 2025-07-15
chore: Updating dependencies + preparing further changes

nicolasnoble:
- All dependencies updated to their latest possible version
- Updating performance.rs accordingly, since black_box became deprecated in Criterion
- Adding clap, config, and serde_yaml for future work

### DIFF
--- a/model_express_server/Cargo.toml
+++ b/model_express_server/Cargo.toml
@@ -8,8 +8,8 @@ model_express_common = { path = "../model_express_common" }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 chrono = { workspace = true }
-clap = { version = "4.5", features = ["derive", "env"] }
-config = { version = "0.15", features = ["yaml", "toml", "json"] }
+clap = { workspace = true }
+config = { workspace = true }
 futures = { workspace = true }
 hf-hub = { workspace = true }
 once_cell = { workspace = true }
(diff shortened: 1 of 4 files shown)


model_express_server/Cargo.toml
```diff
@@ -8,23 +8,27 @@ model_express_common = { path = "../model_express_common" }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 chrono = { workspace = true }
+clap = { version = "4.5", features = ["derive", "env"] }
```
nikkon-dev: Why not put this in the workspace's Cargo.toml?

nicolasnoble: Right, I was doing `cargo add` through a vscode extension to avoid having to deal with this myself. What's the preferred method? Should I always edit the cargo.toml manually? What's the best practice here?

nikkon-dev: Unfortunately, the official cargo add command does not support workspaces yet [Issue](https://github.com/rust-lang/cargo/issues/10608).

The best option is to either use [cargo-autoinherit](https://mainmatter.com/blog/2024/03/18/cargo-autoinherit/) or manually add dependencies into the workspace-level Cargo.toml (the dependi vscode/cursor plugin may help here).

nicolasnoble: Right, okay, thanks! That's more clear now to me.

PR #12, opened 2025-07-15
feat: Provides a trait interface for providers.

nicolasnoble:
In order to have future providers being implemented, we provide a trait interface which will allow an easy method to implement future providers.

We also add the "delete model" feature, which will be useful when we need to purge a local cached entry.

This is built on top of #11 for incremental reviewing.

### DIFF
--- a/model_express_server/src/config.rs
+++ b/model_express_server/src/config.rs
@@ -369,11 +369,23 @@ impl ServerConfig {
         }
 
         // Add environment variables (with MODEL_EXPRESS_ prefix)
+        // Enable debug logging for environment variable parsing
+        if std::env::var("RUST_LOG")
+            .unwrap_or_default()
+            .contains("debug")
+        {
+            eprintln!("Debug: Parsing environment variables with MODEL_EXPRESS_ prefix...");
+            for (key, value) in std::env::vars() {
+                if key.starts_with("MODEL_EXPRESS_") {
+                    eprintln!("  Found env var: {key}={value}");
+                }
+            }
+        }
+
         builder = builder.add_source(
             Environment::with_prefix("MODEL_EXPRESS")
                 .try_parsing(true)
-                .separator("_")
-                .list_separator(","),
+                .separator("_"),
         );
 
         // Build the config and merge with defaults
@@ -383,7 +395,37 @@ impl ServerConfig {
                 // Clone the config before trying to deserialize since try_deserialize consumes it
                 match c.clone().try_deserialize::<ServerConfig>() {
                     Ok(full_config) => full_config,
-                    Err(_) => {
+                    Err(deserialize_err) => {
+                        // Provide detailed error information about what went wrong
+                        eprintln!(
+                            "Warning: Configuration deserialization failed with error: {deserialize_err}"
+                        );
+
+                        // Check for common issues and provide better guidance
+                        if deserialize_err.to_string().contains("missing field") {
+                            eprintln!(
+                                "  This is likely due to incomplete environment configuration."
+                            );
+                            eprintln!(
+                                "  The configuration system expects all fields to be present or uses defaults."
+                            );
+                            eprintln!(
+                                "  Falling back to default configuration with manual value extraction..."
+                            );
+                        } else if deserialize_err.to_string().contains("sequence") {
+                            eprintln!(
+                                "  This was likely caused by the previous list separator configuration."
+                            );
+                            eprintln!(
+                                "  This should be fixed now, but if you still see this error,"
+                            );
+                            eprintln!("  please check your environment variable values.");
+                        } else {
+                            eprintln!(
+                                "  Unexpected deserialization error. Falling back to manual configuration."
+                            );
+                        }
+
                         // Start with default and manually merge what we can
                         let mut default_config = ServerConfig::default();
 
@@ -416,8 +458,14 @@ impl ServerConfig {
                     }
                 }
             }
-            Err(_) => {
-                // If building the config fails entirely, just use defaults
+            Err(build_err) => {
+                // If building the config fails entirely, show the error and use defaults
+                eprintln!("Error: Configuration building failed: {build_err}");
+                eprintln!("  This could be due to:");
+                eprintln!("  - Invalid environment variable values");
+                eprintln!("  - Malformed configuration file");
+                eprintln!("  - Environment variables with unexpected types");
+                eprintln!("  Falling back to default configuration...");
                 ServerConfig::default()
             }
         };


model_express_common/src/lib.rs
```diff
@@ -3,6 +3,8 @@ use serde::{Deserialize, Serialize};
 pub mod cache;
 pub mod download;
 pub mod models;
+#[allow(clippy::mod_module_files)]
```
nikkon-dev: Usually, you need to choose self_named_module_files or mod_module_files for the whole project, or not use any of them at all. Specifying the exact module layout for each module would not work well in the long term.

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+        let token = env::var(HF_TOKEN_ENV_VAR).ok();
+        let api = ApiBuilder::new()
+            .with_progress(true)
+            .with_token(token)
+            .build()?;
+        let model_name = model_name.to_string();
+
+        let repo = api.model(model_name.clone());
+
+        let info = match repo.info().await {
```
nikkon-dev:
```
let info = repo.info().await.map_err(|e| e.context("Failed to fetch model '{model_name}' from HuggingFace. Is this a valid HuggingFace ID?")?
```
This way, you preserve the original error in the chain of errors and can output/analyze the whole chain later, or log it with much more context.
If you enable anyhow's backtrace feature, you will also preserve the original call stacks of the errors.

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+            Err(e) => {
+                return Err(anyhow::anyhow!(
+                    "Failed to fetch model '{model_name}' from HuggingFace: {e}. Is this a valid HuggingFace ID?"
+                ));
+            }
+        };
+        info!("Got model info: {info:?}");
+
+        if info.siblings.is_empty() {
+            return Err(anyhow::anyhow!(
```
nikkon-dev: Such constructs may be replaced with `anyhow::bail!("Model....");`

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+        for sib in info.siblings {
+            if HuggingFaceProvider::is_ignored(&sib.rfilename)
+                || HuggingFaceProvider::is_image(Path::new(&sib.rfilename))
+            {
+                continue;
+            }
+
+            match repo.get(&sib.rfilename).await {
+                Ok(path) => {
+                    p = path;
```
nikkon-dev, line 86: Is one of the following true?
- The path of all siblings is the same
- You do not care which sibling's path to use

You are overwriting the path on each iteration and not using it except for the very last one.

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+            None => Err(anyhow::anyhow!("Invalid HF cache path: {}", p.display())),
+        }
+    }
+
+    /// Attempt to delete a model from Hugging Face cache
+    /// Returns Ok(()) if the model was successfully deleted or didn't exist
+    async fn delete_model(&self, model_name: &str) -> Result<()> {
+        info!("Deleting model from Hugging Face cache: {model_name}");
+        let token = env::var(HF_TOKEN_ENV_VAR).ok();
+        let api = ApiBuilder::new().with_token(token).build()?;
```
nikkon-dev: It's a good practice to add `.context` or log with `.inspect_err` before returning the error to the caller.

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+
+        for sib in &info.siblings {
+            if HuggingFaceProvider::is_ignored(&sib.rfilename)
+                || HuggingFaceProvider::is_image(Path::new(&sib.rfilename))
+            {
+                continue;
+            }
+
+            // Try to get the file path from cache first
+            match repo.get(&sib.rfilename).await {
```
nikkon-dev: This is technically 
```
if let Some(cached_path) = repo.get(&sig.rfilename).await {
...
}
```
Your `Err(_)` branch does nothing.

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+                        Ok(_) => {
+                            files_deleted = files_deleted.saturating_add(1);
+                            info!("Deleted cached file: {}", cached_path.display());
+                        }
+                        Err(e) => {
+                            let error_msg = format!(
+                                "Failed to delete cached file '{}': {e}",
+                                cached_path.display()
+                            );
+                            deletion_errors.push(error_msg);
```
nikkon-dev: You could preserve the original errors for context:
```
deletion_errors.push(e.context("Failed to delete...", cached_path.display()));
```

model_express_common/src/providers/huggingface.rs
```diff
@@ -0,0 +1,214 @@
...
+                    continue;
+                }
+            }
+        }
+
+        // Try to remove the empty model directory if all files were deleted
+        if files_deleted > 0 && deletion_errors.is_empty() {
+            // Get any file path to find the model directory
+            for sib in &info.siblings {
+                if let Ok(cached_path) = repo.get(&sib.rfilename).await {
```
nikkon-dev: Are you using Rust 1.88? The let-chains were stabilized there.
```
if let Some(model_dir) = cached_path.parent() 
   && let Ok(mut entries) = std::fs::read_dir() 
   && entities.next().is_none() {
...
}
```

model_express_common/src/lib.rs
nicolasnoble: Right, that's a leftover. I started with an actual mod file, realized this was stupid based off the clippy settings, and switched to self named modules instead.

model_express_common/src/providers/huggingface.rs
nicolasnoble: Right, this was just copy/pasted from https://github.com/ai-dynamo/dynamo/blob/main/lib/llm/src/hub.rs#L43-L52 :)

Thanks.

model_express_common/src/providers/huggingface.rs
nicolasnoble: Hah, again, copy-pasted mindlessly from https://github.com/ai-dynamo/dynamo/blob/main/lib/llm/src/hub.rs#L69-L81 - fun.

nicolasnoble: I'll address this later - right now it matches the behavior of Dynamo.

model_express_common/src/providers/huggingface.rs
nicolasnoble: Got it. I'll keep this in mind. From what I'm reading, `.context` seems a bit more idiomatic?

model_express_common/src/providers/huggingface.rs
nicolasnoble: Ah, I was trying to eat errors, but, yes, I see that now.

model_express_common/src/providers/huggingface.rs
nicolasnoble: Yeah no idea what let-chains are, but I can take a guess from your suggestion :)

model_express_common/src/providers/huggingface.rs
nicolasnoble: How would we return a collection of errors in this case at the end?

Currently trying with

```Rust
            let mut compound_error =
                anyhow::anyhow!("Failed to delete some files for model '{model_name}'");

            for (i, error) in deletion_errors.into_iter().enumerate() {
                compound_error =
                    compound_error.context(format!("Error {}: {:#}", i.saturating_add(1), error));
            }

            return Err(compound_error);
```

But is there something more idiomatic?

PR #13, opened 2025-07-15
feat: Adding layered configuration system.

nicolasnoble:
This overhauls the configuration mechanism of ModelExpress. Also prepares configuration variables for the upcoming cache cleanup mechanism.

Builds on top of #12 for easier reviewing.

### DIFF
--- a/model_express_server/src/config.rs
+++ b/model_express_server/src/config.rs
@@ -1,6 +1,7 @@
+use chrono::Duration;
 use clap::{Parser, ValueEnum};
 use config::{Config, ConfigError, Environment, File};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use std::fmt;
 use std::net::SocketAddr;
 use std::num::NonZeroU16;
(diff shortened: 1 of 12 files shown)


model_express_server/src/bin/config_gen.rs
```diff
@@ -0,0 +1,377 @@
...
+struct ConfigGenArgs {
+    /// Output file path
+    #[arg(short, long, default_value = "model-express.yaml")]
+    output: PathBuf,
+
+    /// Input configuration file to migrate/upgrade
+    #[arg(short, long)]
+    input: Option<PathBuf>,
+
+    /// Output format (yaml only for now - probably ever)
```
nikkon-dev: JFYI, This comment is visible when you run your app with `-h` or `--help`

model_express_server/src/bin/config_gen.rs
```diff
@@ -0,0 +1,377 @@
...
+    if !args.dry_run && args.output.exists() && !args.overwrite {
+        eprintln!(
+            "Error: File {} already exists. Use --overwrite to replace it.",
+            args.output.display()
+        );
+        std::process::exit(1);
+    }
+
+    // Load configuration - either from input file or default
+    let config = if let Some(input_path) = &args.input {
```
nikkon-dev, line 52:
```
let config = &args.input.map_or_else(ServerConfig::default, |p| {load_existing_config(p)?});
```

model_express_server/src/bin/config_gen.rs
```diff
@@ -0,0 +1,377 @@
...
+        for (key, val) in map {
+            if let serde_yaml::Value::String(key_str) = key {
+                match key_str.as_str() {
+                    "host" => {
+                        if let serde_yaml::Value::String(host) = val {
+                            server.host = host.clone();
+                        }
+                    }
+                    "port" => {
+                        if let serde_yaml::Value::Number(n) = val {
```
nikkon-dev: Let-chain is possible here

model_express_server/src/bin/config_gen.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn merge_database_config(
+    database: &mut model_express_server::config::DatabaseSettings,
+    value: &serde_yaml::Value,
+) -> Result<(), Box<dyn std::error::Error>> {
+    if let serde_yaml::Value::Mapping(map) = value {
```
nikkon-dev, line 228: I saw this pattern multiple times here already. Would you be intrested in a special trait that would provide some `.str_map_iter()` function that will return `Iterator<Item=(&str, typeof_var(val))>` ?
Then this code would look like
```
for (key, val) in value.str_map_iter() {
    match key {
        "path" => ...
    }
}
```

model_express_server/src/config.rs
```diff
@@ -0,0 +1,325 @@
...
+        // Validate configuration
+        config.validate()?;
+
+        Ok(config)
+    }
+
+    /// Validate the configuration
+    pub fn validate(&self) -> Result<(), ConfigError> {
+        // Validate log level
+        match self.logging.level.to_lowercase().as_str() {
```
nikkon-dev: You can use clap's `arg(value_enum)` and define enums for logging levels and format. This kind of checks clap may handle on its own.

model_express_server/src/config.rs
```diff
@@ -0,0 +1,325 @@
...
+            _ => {
+                return Err(ConfigError::Message(format!(
+                    "Invalid log format: {}. Must be one of: json, pretty, compact",
+                    self.logging.format
+                )));
+            }
+        }
+
+        // Validate port range
+        if self.server.port == 0 {
```
nikkon-dev: If you set the port type to `NonZero(u16)`, clap will automatically check for this constraint

model_express_server/src/config.rs
```diff
@@ -0,0 +1,325 @@
...
+    pub fn socket_addr(&self) -> Result<SocketAddr, ConfigError> {
+        let addr = format!("{}:{}", self.server.host, self.server.port);
+        addr.parse()
+            .map_err(|e| ConfigError::Message(format!("Invalid server address {addr}: {e}")))
+    }
+
+    /// Get the logging level as a tracing Level
+    pub fn log_level(&self) -> Level {
+        match self.logging.level.to_lowercase().as_str() {
+            "trace" => Level::TRACE,
```
nikkon-dev: You can directly have the Level in the Args structure with (`value_enum`)

model_express_server/src/bin/config_gen.rs
nicolasnoble: This is what happens when you refactor your own code 10 times before publishing it and you already lost initial context :)

model_express_server/src/bin/config_gen.rs
nicolasnoble: Sure - when I was writing this, I was wondering actually if there was a better way to refactor, and while I'd know how in C++, I'm still not there in Rust.

model_express_server/src/bin/config_gen.rs
nicolasnoble: I'm not too sure how to use this exactly?
```
53 |         .map_or_else(ServerConfig::default, |p| load_existing_config(&p)?);
   |                                             ---                         ^ cannot use the `?` operator in a closure that returns `ServerConfig`
   |                                             |
   |                                             this function should return `Result` or `Option` to accept `?`
```

model_express_server/src/config.rs
nicolasnoble: It turns out this was a change that propagated heavily :) But that was worth it, I think.

PR #14, opened 2025-07-15
feat: Implementing cache eviction.

nicolasnoble:
Also using the new configuration system. Built on top of #13 for easier reviewing.

### DIFF
--- a/model_express_client/src/bin/fallback_test.rs
+++ b/model_express_client/src/bin/fallback_test.rs
@@ -1,4 +1,4 @@
-#![allow(clippy::expect_used, clippy::unwrap_used)]
+#![allow(clippy::expect_used)]
 
 use model_express_client::{Client, ClientConfig, ModelProvider};
 use tracing::{error, info};
--- a/model_express_client/src/bin/test_client.rs
+++ b/model_express_client/src/bin/test_client.rs
@@ -1,4 +1,4 @@
-#![allow(clippy::expect_used, clippy::unwrap_used)]
+#![allow(clippy::expect_used)]
 
 use model_express_client::{Client, ClientConfig};
 use model_express_common::models::ModelProvider;
--- a/model_express_common/src/models.rs
+++ b/model_express_common/src/models.rs
@@ -45,7 +45,7 @@ pub struct ModelStatusResponse {
 }
 
 #[cfg(test)]
-#[allow(clippy::unwrap_used)]
+#[allow(clippy::expect_used)]
 mod tests {
     use super::*;
     use serde_json::json;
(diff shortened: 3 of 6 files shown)


model_express_server/src/cache.rs
```diff
@@ -0,0 +1,660 @@
...
+
+/// Configuration for cache eviction policies
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CacheEvictionConfig {
+    /// Whether cache eviction is enabled
+    pub enabled: bool,
+    /// The eviction policy to use
+    pub policy: EvictionPolicyType,
+    /// How often to run the eviction process (in seconds)
+    pub check_interval_seconds: u64,
```
nikkon-dev: If you use something like this, you do not need to limit users artificially; values like "100ms", "2s", "3m" would just naturally work.
```
/// Parse a duration string into a `std::time::Duration`.
/// Used in the [`clap::builder::ValueParser`]
pub(crate) fn parse_duration(value: &str) -> Result<Duration, String> {
    use core::str::FromStr as _;
    use jiff::Span;
    let span = Span::from_str(value).map_err(|err| format!("Invalid duration: {err}"))?;
    Duration::try_from(span).map_err(|err| format!("Invalid duration: {err}"))
}
```

model_express_server/src/cache.rs
```diff
@@ -0,0 +1,660 @@
...
+}
+
+impl Default for CacheEvictionConfig {
+    fn default() -> Self {
+        Self {
+            enabled: true,
+            policy: EvictionPolicyType::Lru(LruConfig::default()),
+            check_interval_seconds: 3600, // Default: check every hour
```
nikkon-dev: Use chrono::Duration type. Not just integer types.

model_express_server/src/cache.rs
```diff
@@ -0,0 +1,660 @@
...
+pub enum EvictionPolicyType {
+    /// Least Recently Used policy
+    Lru(LruConfig),
+}
+
+/// Configuration for LRU eviction policy
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LruConfig {
+    /// Time threshold in seconds before an unused model is eligible for removal
+    pub unused_threshold_seconds: i64,
```
nikkon-dev: Duration as well. Also, i64 - can this be negative?

model_express_server/src/cache.rs
```diff
@@ -0,0 +1,660 @@
...
+            interval(TokioDuration::from_secs(self.config.check_interval_seconds));
+
+        loop {
+            tokio::select! {
+                _ = interval_timer.tick() => {
+                    if let Err(e) = self.run_eviction_cycle().await {
+                        error!("Error during cache eviction cycle: {e}", e = e);
+                    }
+                }
+                _ = &mut shutdown_receiver => {
```
nikkon-dev, line 237: This will not have a chance to do anything until run_eviction_cycle.await is finished above. Is that ok?

model_express_server/src/main.rs
```diff
@@ -1,49 +1,122 @@
...
+    let config = match ServerConfig::load(args) {
+        Ok(config) => config,
+        Err(e) => {
+            eprintln!("Configuration error: {e}");
+            std::process::exit(1);
+        }
+    };
+
+    // If validation was requested, exit after successful validation
+    if std::env::args().any(|arg| arg == "--validate-config") {
```
nikkon-dev: Is the `--validate-config` mutually exclusive with anything else? That's usually handled by clap's `conflicts_with` attribute.

model_express_server/src/cache.rs
nicolasnoble: The fact this was an i64 was for convenience on the duration test by substraction. Flipped it all to Durations.

model_express_server/src/cache.rs
nicolasnoble: Yes, I think it should be fine. An eviction cycle shouldn't take too long.

model_express_server/src/main.rs
nicolasnoble: It shouldn't really be mutually exclusive, no. I'm not sure how I'd use `conflicts_with` here?

PR #15, opened 2025-07-15
chore: Various final cleanups.

nicolasnoble:
- Formatting.
- Removing a bunch of unwraps still.


tests/integration_tests.rs
```diff
@@ -89,7 +99,8 @@ async fn test_integration_small_model_download() {
     let result = Client::download_model_directly(
         "prajjwal1/bert-tiny", // A very small BERT model for testing
         ModelProvider::HuggingFace,
-    ).await;
+    )
+    .await;
```
AndyDai-nv, line 103: why do we use separate lines for `.await` at various places?

PR #19, opened 2025-07-21
build: Creating basic devcontainer environment.

nicolasnoble:


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,69 @@
+# This Dockerfile can be used instead of the image property in devcontainer.json
+# for more advanced customization
+
+FROM ubuntu:24.04
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Configure apt and install packages
+RUN apt-get update \
```
the-david-oy, line 10: If you move this to the end of the Dockerfile and/or separate out the apt-get update/install, your Docker container will be able to reuse its cached layers more often (versus needing to rebuild most of the layers any time you update the package list).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "remoteUser": "vscode",
+    "mounts": [
+        "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind"
+    ],
+    "containerEnv": {
+        "RUST_LOG": "debug",
+        "CARGO_TARGET_DIR": "/tmp/target"
+    }
+}
```
the-david-oy, line 54: End of file newline

.devcontainer/Dockerfile
nicolasnoble: I mean it's pretty much the only thing this Dockerfile does, and the packages are needed for the installation of the Rust toolchain right after :) So there's not really much to re-order here.

.devcontainer/devcontainer.json
nicolasnoble: For some reason visual studio code won't let me add one :D

PR #20, opened 2025-07-21
fix: fix k8s deployment

AndyDai-nv:


k8s-deployment.yaml
```diff
@@ -53,34 +48,7 @@ spec:
...
-metadata:
-  name: model-db-pvc
-spec:
-  accessModes:
-    - ReadWriteOnce
-  resources:
-    requests:
-      storage: 1Gi
-  storageClassName: standard
+          claimName: nnoble-pvc
```
the-david-oy, line 51: Is this intentional? We might not want to have any code specific to one person.

AndyDai-nv: This is the PVC used on our local k8s cluster for testing, previous deployment file does not work, have to set this to something that exists and verify the other parts are all working. Will modify this in future PRs

the-david-oy: Thanks for clarifying!

PR #31, opened 2025-08-13
feat: full k8s config running both modelexpress and aggregated dynamo

KavinKrishnan:
Adds deployment .yaml launching both modelexpress and aggregated dynamo in a single DynamoGraphDeployment .yaml file for Qwen

### DIFF
--- a/examples/aggregated_k8s/agg.yaml
+++ b/examples/aggregated_k8s/agg.yaml
@@ -53,10 +53,14 @@ spec:
           image: model-express:latest
           imagePullPolicy: IfNotPresent
           env:
-            - name: SERVER_PORT
+            - name: MODEL_EXPRESS_SERVER_PORT
               value: "8002"
-            - name: LOG_LEVEL
+            - name: MODEL_EXPRESS_LOGGING_LEVEL
               value: "info"
+            - name: MODEL_EXPRESS_DATABASE_PATH
+              value: "/root/models.db"
+            - name: MODEL_EXPRESS_CACHE_DIRECTORY
+              value: "/root/.model-express/cache"
             - name: MODEL_NAME
               value: "Qwen/Qwen3-0.6B"
             - name: MODEL_CACHE_PATH


examples/aggregated_k8s/agg.yaml
```diff
@@ -0,0 +1,224 @@
...
+            periodSeconds: 10
+            failureThreshold: 60
+          image: nvcr.io/nvidian/nim-llm-dev/vllm-runtime:hzhou-0729-2
+          workingDir: /workspace/components/backends/vllm
+          command:
+            - /bin/sh
+            - -c
+          args:
+            - |
+              echo "Waiting for model to be ready..."
```
julienmancuso, line 220: wouldn't it be better to have an init container that runs from a modelexpress docker image which would provide this script ?
The script could be fed with env variable for model name

examples/aggregated_k8s/agg.yaml
```diff
@@ -0,0 +1,224 @@
...
+            - name: SERVER_PORT
+              value: "8002"
+            - name: LOG_LEVEL
+              value: "info"
+          command:
+            - /bin/sh
+            - -c
+          args:
+            - |
+              echo "Starting Model Express Server..."
```
julienmancuso, line 73: couldn't this script part of the modelexpress docker image instead ?

examples/aggregated_k8s/agg.yaml
KavinKrishnan: Ideally yes, but we are not shipping containers for this release. 

Understand that it is not clean, but for now it seems ok to have this in the example agg.yaml.

We are aiming for 0.5.0 to have release containers to enhance the UX.

cc @nicolasnoble @ganeshku1

examples/aggregated_k8s/agg.yaml
KavinKrishnan: This is the dynamo platform container which we do not have flexibility with.

For 0.5.0 as mentioned in a previous comment, we are targeting to have release containers. During the scoping discussions for that release we should definitely discuss how modelexpress is packaged with Dynamo @ganeshku1

ganeshku1: Yes we are targeting this container release for 0.5.0 release for Dynamo.

examples/aggregated_k8s/agg.yaml
julienmancuso: ok got it, but still even if you use dynamo image, this should be done in a separate init container

PR #34, opened 2025-08-13
ci: Add internal CI trigger workflow

pvijayakrish:
This PR adds the workflow to trigger the internal CI on PRs targeting main and on commits to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,8 @@ env:
 
 # Cancel any previous check runs for the same pull request to avoid redundant workflows.
 concurrency:
-  group: ${{ github.event_name == 'pull_request' && format('{0}-{1}', github.workflow, github.event.pull_request.number) || '' }}
-  cancel-in-progress: true
+  group: ${{ github.event_name == 'pull_request' && format('{0}-{1}', github.workflow, github.event.pull_request.number) || format('{0}-{1}', github.workflow, github.run_id) }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   test:


.github/workflows/ci.yml
```diff
@@ -10,6 +10,11 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
 
+# Cancel any previous check runs for the same pull request to avoid redundant workflows.
+concurrency:
+  group: ${{ github.event_name == 'pull_request' && format('{0}-{1}', github.workflow, github.event.pull_request.number) || '' }}
```
nv-anants: does the rule mean that group will be empty if github event is not a pull request? Will it cancel runs from main and develop as well then?

.github/workflows/trigger_ci.yml
```diff
@@ -0,0 +1,27 @@
+name: Trigger CI for modelexpress
+
+on:
+  push:
+    branches:
+      - main
+      - 'release/*.*.*'
+  pull_request:
```
nv-anants, line 11: are we adding copy-pr-bot here? If yes, we need to use `pull-request/*` branches as trigger, instead of PRs

.github/workflows/ci.yml
pvijayakrish: Yes this is intentionally left to be empty if github event is not a pull request. This will not cancel previous check runs from main/release.

.github/workflows/trigger_ci.yml
pvijayakrish: Not planning to accept external contribution in 0.4.1.

PR #71, opened 2025-09-04
fix: PVC cache directory configuration for Kubernetes deployment

KavinKrishnan:


k8s-deployment.yaml
```diff
@@ -61,6 +69,17 @@ spec:
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: nnoble-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
```
AndyDai-nv, line 85: Same thing here, 10Gi might be too small for model weights files. Do we need these flags here?

KavinKrishnan: I think we keep it, and it should be sufficient for most smaller models. Its on the user to set this according to their target model. Too large and this can lead to PV resource wastage.

PR #73, opened 2025-09-04
feat: add support for name mapping back to HF id

AndyDai-nv:


AndyDai-nv:
> Could you add support for map that keeps track of hf id --> latest snapshot folder sub path

I would keep changes minimal aiming for fixing nvbug only with this PR

> If I understand correctly, this code just converts to hf id and saves to `model_name`

yes, and this what fixing the nvbug needs

modelexpress_common/src/cache.rs
```diff
@@ -208,6 +208,26 @@ impl CacheConfig {
...
+        } else if folder_name.starts_with("datasets--") {
+            // TODO: Handle datasets names conversion
+            folder_name.to_string()
+        } else if folder_name.starts_with("spaces--") {
+            // TODO: Handle spaces names conversion
+            folder_name.to_string()
+        } else {
+            // If it doesn't match the expected pattern, return as-is
+            folder_name.to_string()
+        }
```
nv-hwoo, line 228: maybe just group all the other cases into `else` since they're all returned as-is for now?
```suggestion
        } else {
            // TODO: Handle other prefixes. return as-is for now.
            folder_name.to_string()
        }
```

nv-hwoo (commented):
is it common to use "folder" as varname in a program? i feel like folder is usually used in the context of gui but i guess it doesn't matter much 😆

modelexpress_common/src/cache.rs
AndyDai-nv: These are huggingface folder categories, although we use models for now, I would just keep them as place holders😆

>is it common to use "folder" as varname in a program? i feel like folder is usually used in the context of gui but i guess it doesn't matter much 😆

the original function prints out a list of folder names, and the nvbug asks for hf id, so here are these changes🙃

nv-hwoo (approved):

PR #74, opened 2025-09-05
fix: ensure tilde expands to system home directory

nv-hwoo:
### Summary

This pull request refactors how home directory paths are determined and used throughout the codebase, improving consistency and reliability. It introduces a new `Utils` struct with a `get_home_dir` method to standardize home directory retrieval, updates default cache and config paths to use constants, and adds related unit tests. Additionally, a generic error variant is added for improved error handling.

**Home directory and path handling improvements:**
* Added a `Utils` struct with a `get_home_dir` method to consistently retrieve the user's home directory from environment variables, replacing direct environment variable access throughout the codebase.
* Updated default path construction in `CacheConfig` (including `default`, `auto_detect`, `get_config_path`, and `expand_path` methods) to use `Utils::get_home_dir` and centralized constants for cache and config paths.
* Refactored Hugging Face provider to use `Utils::get_home_dir` and path constants for cache directory fallback. 

**Constants and error handling:**
* Introduced constants for default cache, Hugging Face cache, and config paths in the `constants` module, and updated usages to reference these constants.
* Added a new `Generic` variant to the `Error` enum for more flexible error reporting.

### DIFF
--- a/modelexpress_common/src/cache.rs
+++ b/modelexpress_common/src/cache.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::constants;
+use crate::{Utils, constants};
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 use std::env;
@@ -22,8 +22,9 @@ pub struct CacheConfig {
 
 impl Default for CacheConfig {
     fn default() -> Self {
+        let home = Utils::get_home_dir();
         Self {
-            local_path: PathBuf::from("~/.model-express/cache"),
+            local_path: PathBuf::from(home).join(constants::DEFAULT_CACHE_PATH),
             server_endpoint: format!("http://localhost:{}", constants::DEFAULT_GRPC_PORT),
             timeout_secs: None,
         }
@@ -130,9 +131,10 @@ impl CacheConfig {
 
     /// Auto-detect cache configuration
     pub fn auto_detect() -> Result<Self> {
+        let home = PathBuf::from(Utils::get_home_dir());
         let common_paths = vec![
-            PathBuf::from("~/.model-express/cache"),
-            PathBuf::from("~/.cache/huggingface/hub"),
+            home.join(constants::DEFAULT_CACHE_PATH),
+            home.join(constants::DEFAULT_HF_CACHE_PATH),
             PathBuf::from("/cache"),
             PathBuf::from("/app/models"),
             PathBuf::from("./cache"),
@@ -185,13 +187,9 @@ impl CacheConfig {
 
     /// Get configuration file path
     fn get_config_path() -> Result<PathBuf> {
-        let home = env::var("HOME")
-            .or_else(|_| env::var("USERPROFILE"))
-            .context("Could not determine home directory")?;
+        let home = Utils::get_home_dir();
 
-        Ok(PathBuf::from(home)
-            .join(".model-express")
-            .join("config.yaml"))
+        Ok(PathBuf::from(home).join(constants::DEFAULT_CONFIG_PATH))
     }
 
     /// Expand path with tilde and environment variables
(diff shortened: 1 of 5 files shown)


modelexpress_common/src/cache.rs
```diff
@@ -22,8 +22,9 @@ pub struct CacheConfig {
 
 impl Default for CacheConfig {
     fn default() -> Self {
+        let home = Utils::get_home_dir().unwrap_or_else(|_| ".".to_string());
         Self {
-            local_path: PathBuf::from("~/.model-express/cache"),
```
nicolasnoble, line 26: Ah crap, we had that in there? I totally missed it. Good one.

modelexpress_common/src/lib.rs
```diff
@@ -70,10 +74,26 @@ impl From<tonic::transport::Error> for Box<Error> {
...
+            .map_err(|e| Error::Generic(format!("Failed to get home directory: {e}")).into())
+    }
+}
+
 /// Constants shared between client and server
 pub mod constants {
     use std::num::NonZeroU16;
 
+    pub const DEFAULT_CACHE_PATH: &str = ".model-express/cache";
+    pub const DEFAULT_HF_CACHE_PATH: &str = ".cache/huggingface/hub";
```
nicolasnoble, line 94: I'm actually wondering... this should probably be `".model-express/cache/huggingface"`, right? Or, composing the second string from the first one. Also these probably should be PathBuf objects.

modelexpress_common/src/cache.rs
```diff
@@ -130,9 +131,10 @@ impl CacheConfig {
 
     /// Auto-detect cache configuration
     pub fn auto_detect() -> Result<Self> {
+        let home = Utils::get_home_dir().unwrap_or_else(|_| ".".to_string());
         let common_paths = vec![
-            PathBuf::from("~/.model-express/cache"),
-            PathBuf::from("~/.cache/huggingface/hub"),
+            PathBuf::from(&home).join(constants::DEFAULT_CACHE_PATH),
+            PathBuf::from(&home).join(constants::DEFAULT_HF_CACHE_PATH),
```
nicolasnoble, line 137: I think there's a bit of a mixup here at the end of the day. The cache path folder is where we store all of the model files. But the "hugging face" one needs to be a subdirectory of it.

Basically, the global cache folder is where we'll store all model files for everyone, and we have one subfolder there per provider.

modelexpress_common/src/cache.rs
```diff
@@ -185,23 +187,17 @@ impl CacheConfig {
...
-            .join(".model-express")
-            .join("config.yaml"))
+        Ok(PathBuf::from(home).join(constants::DEFAULT_CONFIG_PATH))
     }
 
     /// Expand path with tilde and environment variables
     fn expand_path(path: &Path) -> Result<PathBuf> {
         let path_str = path.to_string_lossy();
 
         if let Some(stripped) = path_str.strip_prefix("~/") {
```
AndyDai-nv: do we still need these code here to probe prefix `~/`?

modelexpress_common/src/lib.rs
AndyDai-nv: I feel like the reasons here is to keep original `common_paths` vector content in `modelexpress_common/src/cache.rs` unchanged. Unless `.model-express/cache/huggingface` is the new path needed

nv-hwoo: ~~i can change them to pathbufs, but~~ (pathbuf is inherently mutable and is incompatible with const expression) as @AndyDai-nv said, this was just moving the string literals into constant variables to keep them in a single place and reuse them across the codebase. 

> this should probably be ".model-express/cache/huggingface", right?

was this the original intended file hierarchy? which i think makes sense, because i was also a bit confused why we were maintaining two separate cache directories.

modelexpress_common/src/cache.rs
nv-hwoo: good point, let me check if `expand_path` is no longer needed.

AndyDai-nv: Feel free to put it into backlog if this falls out of the scope of this PR

nv-hwoo: i confirmed that `expand_path` is no longer needed and removed it 👍

modelexpress_common/src/cache.rs
nv-hwoo: thanks for clarifying! i think the change you mentioned deserves a separate PR. i'll create a follow up ticket on our backlog.

PR #75, opened 2025-09-10
feat: add `get_model_path` api for dynamo integration

nv-hwoo:
The PR adds a new client api for retrieving the path to a locally cached model snapshot, integrating it into the `ModelProviderTrait` and implementing it for the HuggingFace provider.

### Client API

* Added `get_model_path` to the `Client` struct in `modelexpress_client`, including logic to determine the cache directory and call the provider's method.

### Trait and Provider Implementation

* Added the `get_model_path` async method to the `ModelProviderTrait`, allowing providers to return the path to the latest cached model snapshot or an error if not found.
* Implemented `get_model_path` for `HuggingFaceProvider`, which locates the newest model snapshot directory in the cache and returns its path, with error handling for missing or empty cache directories.

### Testing

* Updated and expanded tests for provider trait implementations and added a new async test for `get_model_path` in the HuggingFace provider, using a temporary directory to verify correct snapshot selection.

### DIFF
--- a/modelexpress_client/src/lib.rs
+++ b/modelexpress_client/src/lib.rs
@@ -120,7 +120,7 @@ impl Client {
         })
     }
 
-    /// TODO: change to using default client/server cache dirs
+    /// TODO: change to using the current client cache dir
     fn get_model_express_cache_dir() -> PathBuf {
         if let Ok(cache_path) = env::var("HF_HUB_CACHE") {
             return PathBuf::from(cache_path);
--- a/modelexpress_client/src/lib.rs
+++ b/modelexpress_client/src/lib.rs
@@ -3,8 +3,9 @@
 
 mod error;
 
+use modelexpress_common::providers::ModelProviderTrait;
 use modelexpress_common::{
-    Result as CommonResult,
+    Result as CommonResult, Utils,
     cache::{CacheConfig, CacheStats},
     client_config::ClientConfig as Config,
     constants, download,
@@ -14,8 +15,11 @@ use modelexpress_common::{
         model::{ModelDownloadRequest, model_service_client::ModelServiceClient},
     },
     models::{ModelStatus, Status},
+    providers::huggingface::HuggingFaceProvider,
 };
 use std::collections::HashMap;
+use std::env;
+use std::path::PathBuf;
 use std::time::Duration;
 use tonic::transport::Channel;
 use tracing::info;
@@ -113,6 +117,35 @@ impl Client {
         cache_config.clear_all().map_err(|e| {
             modelexpress_common::Error::Server(format!("Failed to clear cache: {e}")).into()
         })
+    }
+
+    /// TODO: change to using default client/server cache dirs
+    fn get_model_express_cache_dir() -> PathBuf {
+        if let Ok(cache_path) = env::var("HF_HUB_CACHE") {
+            return PathBuf::from(cache_path);
+        }
+
+        if let Ok(cache_path) = env::var("MODEL_EXPRESS_PATH") {
+            return PathBuf::from(cache_path);
+        }
+        let home = Utils::get_home_dir().unwrap_or_else(|_| ".".to_string());
+
+        PathBuf::from(home).join(".cache/huggingface/hub")
+    }
+
+    pub async fn get_model_path(&self, model_name: &str) -> anyhow::Result<PathBuf> {
+        let cache_dir = Client::get_model_express_cache_dir();
+
+        info!("### CACHE DIR: {:?}", cache_dir);
+
+        let model_path = HuggingFaceProvider
+            .get_model_path(model_name, cache_dir)
+            .await
+            .map_err(|e| anyhow::anyhow!(format!("Failed to get model path: {e}")))?;
+
+        info!("### MODEL PATH: {:?}", model_path);
+
+        Ok(model_path)
     }
 
     /// Pre-download model to cache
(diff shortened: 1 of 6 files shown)


modelexpress_common/src/providers/huggingface.rs
```diff
@@ -218,6 +221,33 @@ impl ModelProviderTrait for HuggingFaceProvider {
...
+            anyhow::bail!("Model snapshots for '{model_name}' not found in cache");
+        }
+
+        let mut files: Vec<fs::DirEntry> = fs::read_dir(path)?.filter_map(Result::ok).collect();
+        files.sort_by_key(|e| {
+            e.metadata()
+                .and_then(|m| m.created().or_else(|_| m.modified()))
+                .unwrap_or(std::time::SystemTime::UNIX_EPOCH)
+        });
+        files.reverse(); // newest first
```
nv-hwoo: A bit hacky way to figure out the latest snapshot (if multiple exists) since system time is not an ideal way to compare time. we probably want to come up with more robust approach (e.g. timestamping for every model download maybe?) but i think it deserves a separate discussion.

modelexpress_client/src/lib.rs
```diff
@@ -115,6 +120,31 @@ impl Client {
         })
     }
 
+    /// TODO: change to using the current client cache dir
+    fn get_model_express_cache_dir() -> PathBuf {
+        if let Ok(cache_path) = env::var("HF_HUB_CACHE") {
+            return PathBuf::from(cache_path);
+        }
+
+        if let Ok(cache_path) = env::var("MODEL_EXPRESS_PATH") {
```
AndyDai-nv: Are the env vars supposed to be different from what currently being used in mx client or server?

AndyDai-nv (commented):
Overall LGTM since it already works, left one question.

modelexpress_client/src/lib.rs
nv-hwoo: good catch! i actually just moved this function straight from dynamo codebase so didn't realize but i think you're right, this should be `MODEL_EXPRESS_CACHE_PATH`.

AndyDai-nv: Ahh I see, there will be cleanup needed for all these anyway

nv-hwoo: yeah for this one tho, I just updated to correct env var name :)

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -218,6 +221,54 @@ impl ModelProviderTrait for HuggingFaceProvider {
...
+
+        for file in &files {
+            if file.file_name().display().to_string() == info.sha {
+                return Ok(file.path());
+            }
+        }
+
+        warn!(
+            "Existing model snapshots do not match the latest commit hash '{0}'. \
+            Returning the best-effort, latest model snapshot.",
```
AndyDai-nv: I think the following code is returning the latest local commit,
```suggestion
            Returning the best-effort, latest local model snapshot.",
```

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -218,6 +221,54 @@ impl ModelProviderTrait for HuggingFaceProvider {
...
+
+        // A bit hacky way to figure out the latest snapshot
+        files.sort_by_key(|e| {
+            e.metadata()
+                .and_then(|m| m.created().or_else(|_| m.modified()))
+                .unwrap_or(std::time::SystemTime::UNIX_EPOCH)
+        });
+        files.reverse();
+
+        // Check against the latest commit hash from HF
```
AndyDai-nv, line 249: Most of the time, HF repos are updated for nit like README, License, description changes, which will be different commits from users' local repos, so whether checking for latest remote commit or not is your call😄, I am ok with both

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -218,6 +221,54 @@ impl ModelProviderTrait for HuggingFaceProvider {
...
+        let token = env::var(HF_TOKEN_ENV_VAR).ok();
+        let api = ApiBuilder::from_env().with_token(token).build()?;
+        let repo = api.model(model_name.to_string());
+        let info = repo.info().await.map_err(|e| {
+            anyhow::anyhow!("Failed to fetch model '{model_name}' from HuggingFace: {e}")
+        })?;
+
+        for file in &files {
+            if file.file_name().display().to_string() == info.sha {
```
AndyDai-nv, line 258: If we sort the vector before, could we check only the first one?

AndyDai-nv (approved):
LGTM, just some open discussion, feel free to merge without making further changes.😄

modelexpress_common/src/providers/huggingface.rs
nv-hwoo: that's good point! i maybe just overthinking here but also didn't want to ignore the case where weights are actually updated. I still think it's not harmful since we're just letting the user know about it 😄

modelexpress_common/src/providers/huggingface.rs
nv-hwoo: that would be true only if the sorting using system time was robust. one of the reason why i added checking remote hf repo was to add more confidence to the returned path by checking with the ground truth source.

modelexpress_common/src/providers/huggingface.rs
nv-hwoo: yeah good point, updated!

PR #76, opened 2025-09-19
fix: fix yaml config serialization bug

AndyDai-nv:


README.md
```diff
@@ -128,15 +128,48 @@ Create a configuration file (supports YAML):
...
+
+**Basic Configuration (`model-express.yaml`):**
+
+```yaml
+server:
+  host: 0.0.0.0
+  port: 8002
+
+database:
+  path: ./models2.db
```
nv-hwoo: what is the number 2?

modelexpress_common/src/config.rs
```diff
@@ -61,6 +60,16 @@ impl fmt::Display for DurationConfig {
     }
 }
 
+// Serialize as just the number of seconds (not as a struct)
+impl Serialize for DurationConfig {
```
nv-hwoo, line 64: was it the missing `Serialize` trait that was causing the problem? (im not really familiar with serde in rust)

modelexpress_common/src/config.rs
```diff
@@ -249,31 +261,24 @@ where
...
-            "/etc/model-express/config.yml",
-        ];
-
-        for config_path in &default_configs {
-            if PathBuf::from(config_path).exists() {
-                builder = builder.add_source(File::with_name(config_path).required(false));
-                break;
-            }
-        }
+        return Err(ConfigError::Message(
```
nv-hwoo, line 277: maybe i'm confused but are we requiring config file? or is it optional?

README.md
AndyDai-nv: Oops, copied this from nvbug repro steps for testing. Will update.

modelexpress_common/src/config.rs
AndyDai-nv: Yes, basically. The custom yaml doesn't work, server always uses default settings(see nvbug logs)

modelexpress_common/src/config.rs
AndyDai-nv: You are right, it is optional. But this function is not always called unless there is yaml provided. So this error can be only shown when user provides a yaml but somehow the code comes to this `else`

nv-hwoo: ah makes sense. thanks for clarifying!

modelexpress_common/src/config.rs
```diff
@@ -249,31 +261,24 @@ where
...
-        } else {
+        if !config_path.exists() {
             return Err(ConfigError::Message(format!(
                 "Configuration file not found: {}",
                 config_path.display()
             )));
         }
+        builder = builder.add_source(File::from(config_path.clone()));
+    } else if let Some(default_path) = discover_default_config() {
+        println!("Using default config: {}", default_path.display());
```
nv-hwoo: just saw this now. we should probably use `info!` to be consistent with other loggings.

PR #77, opened 2025-09-22
feat: add ignore_weights parameter

nv-hwoo:


modelexpress_common/src/providers/huggingface.rs
```diff
@@ -284,6 +289,87 @@ mod tests {
...
+    /// us to simulate a real model download without external network access.
+    struct MockHFServer {
+        /// WireMock instance; keeps the server alive for the lifetime of the test
+        _server: MockServer,
+        /// Temporary HF cache root that tests pass to `ApiBuilder::with_cache_dir`
+        pub cache_path: PathBuf,
+    }
+
+    impl MockHFServer {
+        /// Start a WireMock server and configure stubs compatible with hf_hub's download flow.
```
AndyDai-nv, line 309: are the following comments necessary?

nv-hwoo: just wanted to add more explanation to the mockserver in case it's not clear. do you find it less helpful?

AndyDai-nv: I am fine with both options, your call, approved😁

nv-hwoo: gotcha :) Iet's keep it and if it ends up being less helpful, we can remove it.

PR #90, opened 2025-10-03
fix: Cherry Pick Rename Helm Chart to Modelexpress (#89)

KavinKrishnan:


ganeshku1:
Going forward it should be mandatory  in all MX PR's to enforce how it was tested, unless they are part of CI testing.
cc: @AndyDai-nv @nicolasnoble @KavinKrishnan 
Until we have CI in place some of reference to manual testing should be placed in PR description.

PR #105, opened 2025-10-17
fix: Update K8s Examples to Latest Operator

KavinKrishnan:


examples/aggregated_k8s/agg.yaml
```diff
@@ -23,48 +22,51 @@ metadata:
...
             - name: MODEL_EXPRESS_LOGGING_LEVEL
               value: "info"
             - name: MODEL_EXPRESS_DATABASE_PATH
               value: "/root/models.db"
             - name: MODEL_EXPRESS_CACHE_DIRECTORY
-              value: "/root/.model-express/cache"
+              value: "/model/.model-express/cache"
             - name: MODEL_NAME
-              value: "Qwen/Qwen3-0.6B"
+              value: "Qwen/Qwen2.5-0.5B-Instruct"
```
grahamking: Why?

KavinKrishnan: No particular reason, was testing 3.3 70B --> quickly switched to smaller model to debug something

KavinKrishnan: Switched back to Qwen3

examples/aggregated_k8s/agg.yaml
```diff
@@ -23,143 +22,94 @@ metadata:
...
             - |
               echo "Starting Model Express Server..."
               ./modelexpress-server &
+              
               SERVER_PID=$!
               echo "Server started with PID: $SERVER_PID"
 
               echo "Setting up Model Express configuration..."
               mkdir -p $MODEL_CACHE_PATH
+              mkdir -p /root/.model-express
```
nv-hwoo: is this intended? the cache dir seems to be under `/model` but we're creating config under `/root`

examples/aggregated_k8s/agg.yaml
```diff
@@ -23,143 +22,94 @@ metadata:
...
-              else
-                echo "Warning: No snapshot directory found for $MODEL_CACHE_DIR"
-              fi
-
               echo "Server PID: $SERVER_PID"
               wait $SERVER_PID
-      pvc:
+      pvcs:
         name: model-cache-pvc
         mountPoint: /root
```
nv-hwoo: does this need to be `/model`?

examples/aggregated_k8s/agg.yaml
```diff
@@ -201,38 +151,35 @@ spec:
           value: "Qwen/Qwen3-0.6B"
         - name: MODEL_CACHE_PATH
           value: "/model/.model-express/cache"
+        - name: HF_HUB_CACHE
+          value: "/model/.model-express/cache"
+        - name: MODEL_EXPRESS_URL
+          value: "http://vllm-agg-modelexpress:8000"
```
nv-hwoo, line 160: where did this address come from?

examples/aggregated_k8s/agg.yaml
```diff
@@ -201,38 +151,35 @@ spec:
...
-              ls -la $MODEL_CACHE_PATH/$MODEL_CACHE_DIR/snapshots/
-              python3 -m dynamo.vllm --model $MODEL_CACHE_PATH/$MODEL_CACHE_DIR/snapshots/latest/ | tee /tmp/vllm.log
-      pvc:
+              python3 -m dynamo.vllm --model $MODEL_NAME --tensor-parallel-size 1 | tee /tmp/vllm.log
+          resources:
+            requests:
+              ephemeral-storage: "50Gi"
+            limits:
+              ephemeral-storage: "50Gi"
+      pvcs:
```
nv-hwoo: I'm not expert in k8s CRD, but is this the right syntax for DGD? The [api reference](https://github.com/ai-dynamo/dynamo/blob/main/docs/kubernetes/api_reference.md#dynamographdeploymentspec) seems to define `pvcs` along the `services` level so I used to define it like following:
```
spec:
  pvcs:
    - name: model-cache-pvc
  services:
    SomeService:
      ...
      volumeMounts:
        - name: model-cache-pvc
          mountPoint: /model
```
maybe both works. Wanted to make sure we're both doing it right.
cc @biswapanda @julienmancuso

examples/aggregated_k8s/agg.yaml
```diff
@@ -23,143 +22,94 @@ metadata:
...
         tcpSocket:
-          port: 8002
+          port: 8000
         initialDelaySeconds: 10
         periodSeconds: 5
         timeoutSeconds: 3
         failureThreshold: 3
       dynamoNamespace: vllm-agg
-      componentType: main
+      componentType: frontend
```
nv-hwoo, line 45: what does setting this to frontend do?

examples/aggregated_k8s/agg.yaml
```diff
@@ -23,143 +22,94 @@ metadata:
...
         mainContainer:
-          image: nvcr.io/nvstaging/ai-dynamo/vllm-runtime:0.4.0
-          workingDir: /workspace/components/backends/vllm
-          command:
-            - /bin/sh
-            - -c
-          args:
-            - "python3 -m dynamo.frontend --http-port 8000"
-      pvc:
+          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:my-tag
```
nv-hwoo:
```suggestion
          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:my-tag  # set the tag here
```
we should probably add a comment here so it's easier for people to notice that they have to set it to some valid tag?

examples/aggregated_k8s/agg.yaml
```diff
@@ -201,38 +151,35 @@ spec:
...
           startupProbe:
             httpGet:
               path: /health
               port: 9090
-            periodSeconds: 10
-            failureThreshold: 60
-          image: nvcr.io/nvstaging/ai-dynamo/vllm-runtime:0.4.0
+            periodSeconds: 30
+            failureThreshold: 240
+          image: nvcr.io/nvidia/ai-dynamo/vllm-runtime:my-tag
```
nv-hwoo: same as above

nv-hwoo (commented):
Nice work! Left a few quick comments

examples/aggregated_k8s/agg.yaml
KavinKrishnan: moved to /model as we caught in benchmarking - good catch

examples/aggregated_k8s/agg.yaml
KavinKrishnan: moved to /model as we caught in benchmarking - good catch

examples/aggregated_k8s/agg.yaml
KavinKrishnan: this is a workaround from instruction with the operator team, it allows creating a k8s svc for this pod (which I reference in another comment you asked)

examples/aggregated_k8s/agg.yaml
KavinKrishnan: Done - good suggestion

examples/aggregated_k8s/agg.yaml
KavinKrishnan: this is the service account that gets created by setting MX as a "front end",  its the target dynamo will use to interact with MX

examples/aggregated_k8s/agg.yaml
KavinKrishnan: Done

examples/aggregated_k8s/agg.yaml
hutm: that does seem right.

nv-hwoo: hi @hutm, are you referring to my example in the comment or Kavin's agg.yaml in the PR?

nv-hwoo (approved):
LGTM!

PR #107, opened 2025-10-24
chore: update devcontainer.json

nv-hwoo:
- set remote user to ubuntu (following [dynamo](https://github.com/ai-dynamo/dynamo/blob/main/.devcontainer/vllm/devcontainer.json) example) to avoid permission errors caused by UID and GID setup in Dockerfile.
- removed default github copilot extension for cursor users
- set default shell to bash


nicolasnoble (approved):

nv-hwoo:
Note: Security Audit is failing due to `cargo-audit` is requiring rust version >=1.89, while we have rust version 1.88.

PR #112, opened 2025-12-03
fix: Addressing changes with dynamo 0.7.0's handling of PVCs

nicolasnoble:
Additionally:
- let's not force people to use microk8s
- let's not force people to use RWX PVCs
- let's not use internal NVidia repositories...

### DIFF
--- a/examples/aggregated_k8s/agg.yaml
+++ b/examples/aggregated_k8s/agg.yaml
@@ -151,7 +151,6 @@ spec:
         requests:
           cpu: "1"
           memory: "2Gi"
-        limits:
         mainContainer:
           image: nvcr.io/ai-dynamo/vllm-runtime:0.7.0
           workingDir: /workspace/examples/backends/vllm
--- a/examples/aggregated_k8s/README.md
+++ b/examples/aggregated_k8s/README.md
@@ -1,6 +1,5 @@
 # Model Express with Aggregated Dynamo Inference on K8s
 
-
 ## Prerequisites
 
 1. **Install Dynamo Cloud on your Cluster**
@@ -10,18 +9,18 @@
 2. **Build and Push Model Express Image**
    - **IMPORTANT**: You must build the Model Express image and push it to a Docker registry that your Kubernetes cluster can access
    - The deployment expects the image to be available at `localhost:5000/model-express:latest` (for local development) or your registry URL
-   
+
    ```bash
    # Build the Model Express image
    docker build -t model-express:latest .
-   
+
    # Tag for your registry (replace with your registry)
    docker tag model-express:latest your-registry/model-express:latest
-   
+
    # Push to your registry
    docker push your-registry/model-express:latest
    ```
-   
+
    **Registry Options:**
    - **Docker Hub**: `docker tag model-express:latest yourusername/model-express:latest`
    - **Local Registry**: Use `localhost:5000/model-express:latest` (requires local registry setup)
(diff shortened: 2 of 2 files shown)


examples/aggregated_k8s/agg.yaml
```diff
@@ -143,25 +144,22 @@ spec:
...
       componentType: main
       replicas: 1
+      volumeMounts:
+        - name: model-cache-pvc
+          mountPoint: /model
       resources:
         requests:
           cpu: "1"
           memory: "2Gi"
         limits:
```
nv-hwoo: should this be removed as well?

examples/aggregated_k8s/agg.yaml
```diff
@@ -8,20 +8,21 @@ metadata:
...
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
       storage: 100Gi
     limits:
       storage: 100Gi
-  storageClassName: microk8s-hostpath
-
+  storageClassName: standard
```
biswapanda, line 17: `standard` is better than `microk8s-hostpath`. 

But `standard` StorageClass is not guaranteed in all Kubernetes clusters, and its presence and behavior depend on how the cluster/storageclasses were installed and configured.

Should we add a note in the README to adjust the storage class name? 

a README section like ~ 
```
## Storage Configuration

Update the storageClassName in examples/aggregated_k8s/agg.yaml to match appropriate stoargeclass available in your cluster:

# Find your storage class name
kubectl get storageclass

# Edit the examples/aggregated_k8s/agg.yaml file and update:
# spec:
#   storageClassName: "your-actual-storage-class"
```

examples/aggregated_k8s/agg.yaml
```diff
@@ -41,6 +42,9 @@ spec:
       dynamoNamespace: vllm-agg
       componentType: main
       replicas: 1
+      volumeMounts:
+        - name: model-cache-pvc
+          mountPoint: /root
```
biswapanda, line 47: curious why we are using `/root` is this needed for some reason due to `root` runtime user ? 

its it ok to use something more generic like `/opt/dynamo/model-express`?

examples/aggregated_k8s/agg.yaml
```diff
@@ -50,7 +54,7 @@ spec:
           memory: "256Mi"
       extraPodSpec:
         mainContainer:
-          image: model-express:latest
+          image: nvcr.io/ai-dynamo/modelexpress-server:0.2.1
```
biswapanda, line 57: great!

examples/aggregated_k8s/agg.yaml
biswapanda: non-blocker question. please feel free to answer and resolve

examples/aggregated_k8s/agg.yaml
biswapanda: non-blocker. please feel free to resolve and follow up

examples/aggregated_k8s/agg.yaml
nv-hwoo: IIUC, I don't think there was any reason to set this to `/root`, so we should be able to change this to your suggestion. Correct me if I'm wrong here @KavinKrishnan @nicolasnoble

nicolasnoble: Yeah this was so we can map straight into the .huggingface's cache folder, without trying too hard, but this can be remapped. We can punt this for further cleanup.

examples/aggregated_k8s/agg.yaml
nicolasnoble: Yes... this is just an empty list.

saturley-hall:
Closed as unnecessary

PR #113, opened 2025-12-03
chore: Removing / refactoring dead code.

nicolasnoble:
- Removed all of the code I could find that's completely unused at the moment.
- Refactored the two very-similar tests clients into a single codebase.

### DIFF
--- a/modelexpress_client/src/lib.rs
+++ b/modelexpress_client/src/lib.rs
@@ -119,6 +119,12 @@ impl Client {
     /// Get the cache directory path, using client's cache config if available,
     /// otherwise discovering from environment variables and defaults.
     fn get_cache_dir(&self) -> PathBuf {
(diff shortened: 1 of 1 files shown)


modelexpress_client/src/lib.rs
```diff
@@ -120,22 +116,22 @@ impl Client {
         })
     }
 
-    /// TODO: change to using the current client cache dir
-    fn get_model_express_cache_dir() -> PathBuf {
-        if let Ok(cache_path) = env::var("HF_HUB_CACHE") {
-            return PathBuf::from(cache_path);
-        }
```
nv-hwoo, line 127: Is this logic handled by the new function? I don't think cache config checks for this env variable.

nicolasnoble: Right! Sorry, I was nuking duplicated environment variable checks, this one needs to stay.

PR #123, opened 2025-12-18
fix: properly flattening CLI arguments from the common structure

nicolasnoble:
This makes the CLI arguments work properly in the client CLI, like they do on the server.


nv-hwoo:
Thanks for the fixes. Do we know why some checks are failing?

nv-hwoo (approved):
LGTM

PR #125, opened 2025-12-23
fix: preload now properly follow shared_storage parameter

nicolasnoble:
When shared_storage is false in cache config, files are now streamed properly from the server to the client after preload completes.


nicolasnoble:
Right, so... untangling these will be difficult honestly. The breakages happened as tooling got updated, and when we rolled over to the new year. It'd be a lot easier to make a separate PR to fix them. All of these are red at head too.

PR #127, opened 2025-12-23
feat: adding k8s test for no-shared storage

nicolasnoble:
When all the other pull requests have been merged, we can then ensure there will be no further regression by enabling this test. Hopefully it should fail right now, and then get green after everything else has been merged.

## Release Notes

* **Tests**
  * Added Kubernetes-based gRPC transfer testing to CI pipeline, validating model file transfers with both CLI and environment variable configurations.

* **Chores**
  * Enhanced CI workflow to include new gRPC transfer test job as a release dependency.

### DIFF
--- a/test_grpc_transfer_k8s.sh
+++ b/test_grpc_transfer_k8s.sh
@@ -17,7 +17,7 @@ RELEASE_NAME="modelexpress-test"
 NAMESPACE="modelexpress-test"
 IMAGE_NAME="modelexpress"
 IMAGE_TAG="test"
-TEST_MODEL="google-t5/t5-small"
+TEST_MODEL="hf-internal-testing/tiny-random-gpt2"
 TIMEOUT_SECONDS=600
 CLEANUP=true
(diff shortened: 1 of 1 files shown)


test_grpc_transfer_k8s.sh
```diff
@@ -0,0 +1,425 @@
...
+  name: grpc-transfer-test-env
+spec:
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: client
+        image: ${IMAGE_NAME}:${IMAGE_TAG}
+        imagePullPolicy: IfNotPresent
```
nv-hwoo, line 305: Should we change this to `Always` so that we don't accidentally run outdated container?

nicolasnoble: Since the script builds the image locally and loads it into kind (no registry), `IfNotPresent` is the right choice here - `Always` would cause a pull failure since there's nowhere to pull from.

AndyDai-nv (approved):

PR #128, opened 2026-01-22
fix: respect HF_HUB_OFFLINE environment variable

KavinKrishnan:
When HF_HUB_OFFLINE is set to 1/ON/YES/TRUE, modelexpress now:
- Skips network requests to HuggingFace
- Uses only locally cached models
- Returns clear error if model not in cache

Added integration tests for offline mode behavior.

```
=== Testing Offline Mode with Uncached Model ===

Attempting to download a model NOT in cache while HF_HUB_OFFLINE=1...
(This should fail with a clear error message)

2026-01-22T20:19:36.861238Z ERROR Model download failed for some/nonexistent-model: Server returned error: Direct download failed: Model 'some/nonexistent-model' not found in cache and HF_HUB_OFFLINE is set. Cannot download model in offline mode. Please ensure the model is cached locally or disable offline mode.
2026-01-22T20:19:36.861259Z ERROR Command execution failed: Server returned error: Direct download failed: Model 'some/nonexistent-model' not found in cache and HF_HUB_OFFLINE is set. Cannot download model in offline mode. Please ensure the model is cached locally or disable offline mode.
Error: Server returned error: Direct download failed: Model 'some/nonexistent-model' not found in cache and HF_HUB_OFFLINE is set. Cannot download model in offline mode. Please ensure the model is cached locally or disable offline mode.
Model Download
  Model: some/nonexistent-model
  Provider: HuggingFace
  Strategy: Direct

❌ FAILED
  Failed to download model 'some/nonexistent-model': Server returned error: Direct download failed: Model 'some/nonexistent-model' not found in cache and HF_HUB_OFFLINE is set. Cannot download model in offline mode. Please ensure the model is cached locally or disable offline mode.
command terminated with exit code 1
```

### DIFF
--- a/modelexpress_common/src/providers/huggingface.rs
+++ b/modelexpress_common/src/providers/huggingface.rs
@@ -318,10 +318,15 @@ impl ModelProviderTrait for HuggingFaceProvider {
 mod tests {
     use super::*;
     use serde_json::json;
+    use std::sync::Mutex;
     use tempfile::TempDir;
     use tokio::time::Duration;
     use wiremock::matchers::{method, path_regex};
     use wiremock::{Mock, MockServer, ResponseTemplate};
+
+    /// Mutex to serialize access to HF_HUB_OFFLINE environment variable across tests.
+    /// This prevents race conditions when tests run in parallel.
+    static ENV_MUTEX: Mutex<()> = Mutex::new(());
 
     /// Minimal mock of the Hugging Face Hub used by tests.
     ///
@@ -491,6 +496,7 @@ mod tests {
 
     #[test]
     fn test_is_offline_mode() {
+        let _guard = ENV_MUTEX.lock().expect("Failed to acquire env mutex");
         unsafe {
             env::set_var(HF_HUB_OFFLINE_ENV_VAR, "1");
             assert!(is_offline_mode());
@@ -504,7 +510,9 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn test_download_model_offline_mode_with_cache() {
+        let _guard = ENV_MUTEX.lock().expect("Failed to acquire env mutex");
         let temp_dir = TempDir::new().expect("Failed to create temporary directory");
         let snapshots_path = temp_dir
             .path()
@@ -530,7 +538,9 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn test_download_model_offline_mode_without_cache() {
+        let _guard = ENV_MUTEX.lock().expect("Failed to acquire env mutex");
         let temp_dir = TempDir::new().expect("Failed to create temporary directory");
 
         unsafe {
--- a/workspace-tests/tests/integration_tests.rs
+++ b/workspace-tests/tests/integration_tests.rs
@@ -5,9 +5,14 @@
 
 use modelexpress_client::{Client, ClientConfig};
 use modelexpress_common::{constants, models::ModelProvider};
+use std::sync::Mutex;
 use std::time::Duration;
 use tokio::time::timeout;
 use tracing::{info, warn};
+
+/// Mutex to serialize access to HF_HUB_OFFLINE environment variable across tests.
+/// This prevents race conditions when tests run in parallel.
+static ENV_MUTEX: Mutex<()> = Mutex::new(());
 
 #[tokio::test]
 #[ignore = "Ignore by default since it requires a running server"]
@@ -136,7 +141,9 @@ async fn test_integration_client_config_validation() {
 }
 
 #[tokio::test]
+#[allow(clippy::await_holding_lock)]
 async fn test_integration_offline_mode_without_cache() {
+    let _guard = ENV_MUTEX.lock().expect("Failed to acquire env mutex");
     // Test that HF_HUB_OFFLINE mode properly fails when model is not cached
     unsafe {
         std::env::set_var("HF_HUB_OFFLINE", "1");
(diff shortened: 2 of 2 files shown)


modelexpress_common/src/providers/huggingface.rs
```diff
@@ -49,26 +58,78 @@ impl HuggingFaceProvider {
...
+                .and_then(|m| m.created().or_else(|_| m.modified()))
+                .unwrap_or(std::time::SystemTime::UNIX_EPOCH)
+        });
+        files.reverse();
+
+        let latest_snapshot = files[0].path();
+        info!(
+            "Using cached model snapshot for '{model_name}' in offline mode: {}",
+            latest_snapshot.display()
+        );
```
grahamking: We should only have one log message, so either this one...

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -49,26 +58,78 @@ impl HuggingFaceProvider {
...
         let cache_dir = get_cache_dir(cache_dir);
         std::fs::create_dir_all(&cache_dir).map_err(|e| {
             anyhow::anyhow!("Failed to create cache directory {:?}: {}", cache_dir, e)
         })?;
 
+        // Check if offline mode is enabled
+        if is_offline_mode() {
+            info!(
+                "HF_HUB_OFFLINE is set. Attempting to use cached model for '{model_name}' without network requests."
+            );
```
grahamking: ... or this log message, but preferably not both.

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -244,7 +305,18 @@ impl ModelProviderTrait for HuggingFaceProvider {
 
     /// Get the full path to the latest model snapshot if it exists
     /// Returns the path if found, or an error if not found
+    ///
+    /// When HF_HUB_OFFLINE is set, this function will skip the API call to check
+    /// for the latest commit hash and return the most recent local snapshot instead.
     async fn get_model_path(&self, model_name: &str, cache_dir: PathBuf) -> Result<PathBuf> {
+        // In offline mode, skip network requests entirely
```
grahamking: We need an Agents.md to tell AI to chill out with the comments. Can you read and tidy them up, so there isn't the repetition?

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -244,7 +305,18 @@ impl ModelProviderTrait for HuggingFaceProvider {
...
+        if is_offline_mode() {
+            info!(
+                "HF_HUB_OFFLINE is set. Returning cached model path for '{model_name}' without network validation."
```
grahamking: .. and don't need this log statement either. However you do it, only say it to the user once.

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -470,4 +542,167 @@ mod tests {
...
+            env::set_var(HF_HUB_OFFLINE_ENV_VAR, "");
+            assert!(!is_offline_mode());
+
+            // Clean up
+            env::remove_var(HF_HUB_OFFLINE_ENV_VAR);
+        }
+
+        // Test when variable is not set
+        assert!(!is_offline_mode());
+    }
```
grahamking, line 510: You have 50 lines of tests for a 3 line function. Is that really necessary?

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -470,4 +542,167 @@ mod tests {
...
+
+        // Create cache directory structure
+        let snapshots_path = cache_path
+            .join("models--test--model")
+            .join("snapshots")
+            .join("abc1234");
+        std::fs::create_dir_all(&snapshots_path).expect("Failed to create directory");
+
+        let provider = HuggingFaceProvider;
+        let result = provider.get_model_path_offline("test/model", cache_path);
```
grahamking: Why did you put `provider` in a variable here?

modelexpress_common/src/providers/huggingface.rs
KavinKrishnan: have started an Agents.md in a seperate MR, will post soon

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -49,26 +58,71 @@ impl HuggingFaceProvider {
     fn is_subdirectory_file(filename: &str) -> bool {
         Path::new(filename).components().count() > 1
     }
+
+    /// Get the model path in offline mode without making any network requests.
+    /// Returns the most recent local snapshot if available, or an error if not found.
+    fn get_model_path_offline(&self, model_name: &str, cache_dir: PathBuf) -> Result<PathBuf> {
```
nv-hwoo: Why do we need this separate `get_model_path_*`? Can't we just use existing `get_model_path`?

For example, it seems it's enough to just
1. skip the downloading if `HF_OFFLINE` is set in `download_model`
2. run plain `get_model_path` using the cache dir (set by the user)?

AndyDai-nv (approved):
Thanks for fixing the core bug, minor problems can be addressed later

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -53,22 +62,26 @@ impl HuggingFaceProvider {
...
-
-        // Get cache directory and ensure it exists
         let cache_dir = get_cache_dir(cache_dir);
         std::fs::create_dir_all(&cache_dir).map_err(|e| {
             anyhow::anyhow!("Failed to create cache directory {:?}: {}", cache_dir, e)
         })?;
 
+        if is_offline_mode() {
+            info!("HF_HUB_OFFLINE is set, using cached model for '{model_name}'");
+            return self.get_model_path(model_name, cache_dir).await;
```
nv-hwoo, line 80: I know the `download_model` return signature includes Pathbuf but I don't think we're actually returning the path in the code, so i think we can just immediately return `Ok()` here.

For example, if you look at how we call in dynamo: https://github.com/ai-dynamo/dynamo/blob/38fbb1db448926f3185b7c1b0f09bd786c040983/lib/llm/src/hub.rs#L40, we run download and getting model path separately.

KavinKrishnan: Lets fix in separate PR as this would involve significant changes across calls, signatures, etc.

nv-hwoo: maybe I wasn't clear enough. what i'm saying is 
```suggestion
            return Ok();
```
should be enough. I don't think the current function actually returns the pathbuf (and we're not expecting it in the code where this is called).

PR #136, opened 2026-01-28
feat: Model Streamer initial support

nicolasnoble:
We provide a method to download from S3 using Model Streamer. Given Model Streamer's API, this ends up as a Python sidecar process for which we add a Rust interface so this becomes completely transparent for the end user, where we only have a new provider that they can use.

This also adds a small test in the form of a minio storage.

I'm not 100% certain of what I'm doing in Python here, so adding @ajcasagrande for reviewing this part too.

## Release Notes

* **New Features**
  * Added Model Streamer provider for downloading models from S3, GCS, or S3-compatible storage
  * Introduced Python sidecar service for cloud-based model management
  * Added configuration support for cloud storage credentials (S3, GCS, MinIO)

* **Chores**
  * Enhanced Docker deployment with multi-service orchestration (server, sidecar, MinIO)
  * Added GitHub Actions pipeline for integration testing with cloud storage


sidecar/src/modelexpress_sidecar.egg-info/dependency_links.txt
nv-hwoo, line 1: is `.egg-info` directory suppose to be part of the PR?

sidecar/src/modelexpress_sidecar/api/__init__.py
nv-hwoo, line 1: how is the sidecar process intended to be used? is there any example of complete workflow that involves both sidecar and MX server?

sidecar/src/modelexpress_sidecar.egg-info/dependency_links.txt
ajcasagrande: typically no, shouldnt be

sidecar/src/modelexpress_sidecar/api/routes.py
```diff
@@ -0,0 +1,88 @@
...
+        result = await downloader.delete_model(model_id)
+        return result
+    except FileNotFoundError:
+        raise HTTPException(status_code=404, detail=f"Model '{model_id}' not found")
+    except Exception as e:
+        logger.exception("Failed to delete model: %s", model_id)
+        return DeleteResponse(
+            success=False,
+            error=str(e),
+        )
```
ajcasagrande: hmm, interesting that it returns 2 different things. marking for now until i confirm what deleteresponse does

sidecar/src/modelexpress_sidecar/api/schemas.py
```diff
@@ -0,0 +1,83 @@
...
+    success: bool = Field(..., description="Whether the download was successful")
+    local_path: Optional[str] = Field(
+        None, description="Local path where the model was downloaded"
+    )
+    files: Optional[list[str]] = Field(None, description="List of downloaded files")
+    total_size: Optional[int] = Field(
+        None, description="Total size of downloaded files in bytes"
+    )
+    error: Optional[str] = Field(None, description="Error message if download failed")
+    error_code: Optional[str] = Field(None, description="Error code for categorization")
```
ajcasagrande, line 63: was there a specific reason to use these over http error responses?

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+def is_ignored_file(filename: str) -> bool:
+    """Check if a file should be ignored."""
+    return filename in IGNORED_FILES
+
+
+def is_image_file(filename: str) -> bool:
+    """Check if a file is an image."""
+    return any(filename.lower().endswith(ext) for ext in IMAGE_EXTENSIONS)
```
ajcasagrande: small nit, but this is calling `.lower()` for every iteration in the loop, so recommend temp variable or something:


```suggestion
    filename = filename.lower()
    return any(filename.endswith(ext) for ext in IMAGE_EXTENSIONS)
```

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+# Files to ignore during download
+IGNORED_FILES = {".gitattributes", ".gitignore", "README.md"}
+
+# Image file extensions to ignore
+IMAGE_EXTENSIONS = {".png", ".jpg", ".jpeg", ".gif", ".webp", ".svg", ".ico", ".bmp", ".tiff", ".tif"}
+
+
+def is_weight_file(filename: str) -> bool:
+    """Check if a file is a model weight file."""
+    return any(filename.endswith(ext) for ext in WEIGHT_EXTENSIONS)
```
ajcasagrande, line 43: does this need the .lower() check like the images one?

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+        # Convert model_id back to a path
+        parts = model_id.split("/", 2)
+        if len(parts) < 2:
+            raise FileNotFoundError(f"Invalid model_id format: {model_id}")
+
+        scheme = parts[0]
+        bucket = parts[1]
+        path = parts[2] if len(parts) > 2 else ""
+
+        local_path = self.default_cache_dir / "model-streamer" / scheme / bucket / path
```
ajcasagrande: hmmm, this hard-codes `self.default_cache_dir`, but the other functions allow specifying a custom cache_dir. sounds like a potential for mismatch?

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+        if parsed.scheme in ("s3", "gs"):
+            bucket = parsed.netloc
+            path = parsed.path.lstrip("/")
+            return parsed.scheme, bucket, path
+        elif parsed.scheme in ("s3+http", "s3+https"):
+            # For MinIO: s3+http://endpoint:port/bucket/path
+            # The netloc is the endpoint, first path component is bucket
+            parts = parsed.path.lstrip("/").split("/", 1)
+            bucket = parts[0]
+            path = parts[1] if len(parts) > 1 else ""
```
ajcasagrande: can be simplified to
```python
bucket, _, path = parsed.path.lstrip("/").partition("/")
```

> **str.partition(sep, /)**
> Split the string at the first occurrence of sep, and return a 3-tuple containing the part before the  separator, the separator itself, and the part after the separator. If the separator is not found, return a 3-tuple containing the string itself, followed by two empty strings.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+            return "s3", bucket, path
+        else:
+            raise ValueError(f"Unsupported URI scheme: {parsed.scheme}")
+
+    def _get_cache_path(self, model_path: str, cache_dir: str) -> Path:
+        """Get the local cache path for a model."""
+        scheme, bucket, path = self._parse_model_uri(model_path)
+        cache_base = Path(cache_dir) if cache_dir else self.default_cache_dir
+        return cache_base / "model-streamer" / scheme / bucket / path
```
ajcasagrande: `"model-streamer"` is used 3 times in this file. Should it be a constant?

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+    return any(filename.lower().endswith(ext) for ext in IMAGE_EXTENSIONS)
+
+
+class ModelDownloader:
+    """Service for downloading models using Model Streamer."""
+
+    def __init__(self):
+        """Initialize the downloader."""
+        self.default_cache_dir = Path(
+            os.environ.get("MODEL_EXPRESS_CACHE_DIRECTORY", "/app/cache")
```
ajcasagrande, line 63: do you need to verify the path exists and raise an exception if it doesn't? although I do see later on you are creating it if it doesnt exist, so maybe not

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+            # Calculate total size
+            total_size = sum(
+                (local_path / f).stat().st_size
+                for f in downloaded_files
+                if (local_path / f).exists()
+            )
+
+            logger.info(
+                "Successfully downloaded %d files (%d bytes) to %s",
+                len(downloaded_files),
```
ajcasagrande: i mean technically you filter out non existent files, but this one doesn't. it also returns them later on. if non existing files were a concern, may want to filter them once so all 3 spots make use of it.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        downloaded_files = []
+
+        try:
+            # Try to import Model Streamer
+            from runai_model_streamer import SafetensorsStreamer
+
+            # For now, we'll use boto3 for actual file downloads
+            # Model Streamer is better suited for streaming to GPU
+            # In phase 2, we can use Model Streamer's streaming capabilities
```
ajcasagrande, line 244: Image

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+            downloaded_files = await self._download_with_boto3(
+                model_path=model_path,
+                local_path=local_path,
+                ignore_weights=ignore_weights,
+            )
+
+        except ImportError:
+            logger.warning("Model Streamer not available, falling back to boto3")
+            downloaded_files = await self._download_with_boto3(
```
ajcasagrande: both of these paths lease to boto3, but what about gcs?

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+        # List objects in the bucket with the given prefix
+        paginator = s3_client.get_paginator("list_objects_v2")
+        downloaded_files = []
+
+        for page in paginator.paginate(Bucket=bucket, Prefix=prefix):
+            for obj in page.get("Contents", []):
+                key = obj["Key"]
+                # Get relative path from prefix
+                rel_path = key[len(prefix):].lstrip("/") if key.startswith(prefix) else key
+                filename = os.path.basename(rel_path)
```
ajcasagrande: Recommend using pathlib:
```python
# add import to top of function
from pathlib import PurePosixPath

filename = PurePosixPath(rel_path).name
```

sidecar/requirements.txt
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ajcasagrande: you shouldn't need a seperate rquirements.txt. all of this can be in the `pyproject.toml`, which it looks like most already are

.gitignore
```diff
@@ -49,3 +49,6 @@ logs/
 
 # Models database
 models.db
+
+# Python testing
+.venv/
```
ajcasagrande, line 56: should probably add the `__pycache__` and egg info and pyc and such

sidecar/src/modelexpress_sidecar/services/downloader.py
noa-neria: Model streamer has such methods [here](https://github.com/run-ai/runai-model-streamer/blob/329949165651cc686ef8a6583abc75df9fbb97cc/py/runai_model_streamer/runai_model_streamer/safetensors_streamer/safetensors_streamer.py#L36) which were created for downloading the model configuration files. This function `pull_files` supports **both s3 and gcs** (and soon Azure). It is implemented using boto3 etc. 
You can use it here too instead of boto3.

However, using a single boto3 client or gcs client will not download as fast as other tools like s5cmd which download in multi parts. For our use case it was not an issue, because configuration files are tiny.

Having said that, the model streamer itself can be leveraged to download to local disk instead of loading to memory. That will require some work in the Model Streamer code.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+        """Get information about a cached model."""
+        # Convert model_id back to a path
+        # model_id format: s3/bucket/path or gs/bucket/path
+        parts = model_id.split("/", 2)
+        if len(parts) < 2:
+            raise FileNotFoundError(f"Invalid model_id format: {model_id}")
+
+        scheme = parts[0]
+        bucket = parts[1]
+        path = parts[2] if len(parts) > 2 else ""
```
nv-hwoo: I see these parsing logic more than once (with some variants too). Should we create a function dedicated to parsing them so the parsing logic is handled in a single place?

sidecar/src/modelexpress_sidecar/services/downloader.py
noa-neria: Model streamer `pull_files` [here](https://github.com/run-ai/runai-model-streamer/blob/329949165651cc686ef8a6583abc75df9fbb97cc/py/runai_model_streamer/runai_model_streamer/safetensors_streamer/safetensors_streamer.py#L36) supports both s3 and gcs (and soon Azure). It is implemented using boto3 etc.
You can use it here too instead of boto3.

nicolasnoble:
A couple of items from the review feedback that we're deferring to a follow-up PR:

1. **Use model streamer `pull_files` instead of raw boto3** (from @noa-neria) - this would give us GCS and Azure support for free instead of reimplementing with boto3. Agreed this is the right direction, just want to keep this PR focused.

2. **Replace `success=False` + HTTP 200 with proper HTTP status codes** (from @ajcasagrande) - this also covers the "delete route returns 2 different response types" comment. The plan is to use appropriate 4xx/5xx codes with a JSON body for error details, rather than returning 200 with `success=False`. This touches the response schemas, route handlers, downloader, and the Rust client, so it's better as its own PR.

sidecar/src/modelexpress_sidecar.egg-info/dependency_links.txt
nicolasnoble: Removed, and added to .gitignore.

sidecar/tests/test_s3_integration.py
```diff
@@ -0,0 +1,153 @@
...
+            "/api/v1/download",
+            json={
+                "model_path": model_path,
+                "cache_dir": tmpdir,
+                "s3_credentials": {
+                    "access_key_id": os.environ.get("AWS_ACCESS_KEY_ID", "minioadmin"),
+                    "secret_access_key": os.environ.get("AWS_SECRET_ACCESS_KEY", "minioadmin"),
+                    "region": os.environ.get("AWS_REGION", "us-east-1"),
+                    "endpoint_url": endpoint,
+                },
```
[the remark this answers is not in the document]
nicolasnoble: Fixed, changed to 'endpoint' in all three test payloads.

modelexpress_common/src/providers/model_streamer.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    /// Encode a model URI as a path-safe identifier
+    /// E.g., "s3://bucket/path/to/model" -> "s3/bucket/path/to/model"
+    fn encode_model_id(model_path: &str) -> String {
+        model_path
+            .replace("s3://", "s3/")
+            .replace("gs://", "gs/")
+            .replace("s3+http://", "s3-http/")
+            .replace("s3+https://", "s3-https/")
```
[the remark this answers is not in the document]
nicolasnoble: Fixed, Rust now strips the endpoint and normalizes to s3/ to match the Python sidecar.

modelexpress_common/src/providers/model_streamer.rs
```diff
@@ -0,0 +1,491 @@
...
+        assert!(ModelStreamerProvider::is_weight_file("weights.bin"));
+        assert!(!ModelStreamerProvider::is_weight_file("config.json"));
+    }
+
+    #[test]
+    fn test_is_ignored() {
+        assert!(ModelStreamerProvider::is_ignored("README.md"));
+        assert!(ModelStreamerProvider::is_ignored(".gitignore"));
+        assert!(!ModelStreamerProvider::is_ignored("model.safetensors"));
+    }
```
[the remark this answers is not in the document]
nicolasnoble: Implemented both functions on ModelStreamerProvider, matching the Python sidecar lists.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+
+        return original_env
+
+    def _restore_credentials(self, original_env: dict[str, str]) -> None:
+        """Restore original environment variables."""
+        for key, value in original_env.items():
+            if value:
+                os.environ[key] = value
+            elif key in os.environ:
+                del os.environ[key]
```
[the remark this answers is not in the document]
nicolasnoble: Added an asyncio.Lock around credential setup/restore.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,357 @@
...
+
+                # Download the file
+                local_file = local_path / rel_path
+                local_file.parent.mkdir(parents=True, exist_ok=True)
+
+                logger.info("Downloading: %s -> %s", key, local_file)
+                s3_client.download_file(bucket, key, str(local_file))
+                downloaded_files.append(rel_path)
+
+        return downloaded_files
```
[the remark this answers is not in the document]
nicolasnoble: Made _download_with_boto3 sync and wrapped calls with asyncio.to_thread.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Fixed, .lower() is called once upfront now.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Added, both functions call .lower() once now.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Fixed, delete_model now accepts an optional cache_dir like get_model.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Done, using partition now.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Extracted to a CACHE_SUBDIR constant.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Fixed, we filter once after download and use the filtered list everywhere.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Done, using PurePosixPath(rel_path).name now.

sidecar/requirements.txt
nicolasnoble: Removed, Dockerfile now installs from pyproject.toml directly.

.gitignore
nicolasnoble: Added __pycache__/, *.pyc, *.pyo, and *.egg-info/ patterns.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Added a _resolve_model_id helper that both get_model and delete_model use now.

sidecar/src/modelexpress_sidecar/api/routes.py
nicolasnoble: Acknowledged, deferring to a follow-up PR. Left a note in the PR comments.

sidecar/src/modelexpress_sidecar/api/schemas.py
nicolasnoble: Acknowledged, deferring to a follow-up PR. Left a note in the PR comments.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Great suggestion, deferring to a follow-up PR. Left a note in the PR comments.

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Same as above, will tackle in a follow-up.

nicolasnoble: Acknowledged, related to the pull_files suggestion. Deferring to a follow-up.

nicolasnoble:
Replying to a few threads below and resolving the ones that are fine to defer.

sidecar/src/modelexpress_sidecar/api/__init__.py
nicolasnoble: The sidecar runs alongside the main Rust server inside the same container. `entrypoint.sh` starts the Python sidecar (FastAPI on port 8002), waits for its health check, then starts the Rust server. The Rust server talks to the sidecar over HTTP when it needs to download models from S3/GCS.

For a complete workflow example, `docker-compose.yml` in this PR shows the full setup: ModelExpress server (with sidecar), MinIO for S3-compatible storage, and a bucket initializer.

We can add more docs on this in a follow-up if that would help!

sidecar/src/modelexpress_sidecar/services/downloader.py
nicolasnoble: Right, the path gets created downstream in `download()` via `local_path.mkdir(parents=True, exist_ok=True)`. No need to validate existence here since this is just resolving the ID to a path - the caller decides what to do with it.

.github/workflows/ci.yml
```diff
@@ -213,6 +213,62 @@ jobs:
         files: lcov.info
         fail_ci_if_error: false
 
+  sidecar-test:
+    name: Sidecar Python Tests
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
nicolasnoble: The existing `k8s-grpc-test` job also does not gate the release job, so this is consistent. We can add sidecar-test to the release gate in a follow-up once the test suite is more mature.

docker-compose.yml
```diff
@@ -9,8 +9,59 @@ services:
...
+      - minio
+    entrypoint: >
+      /bin/sh -c "
+      sleep 5;
+      mc alias set myminio http://minio:9000 minioadmin minioadmin;
+      mc mb myminio/models --ignore-existing;
+      mc anonymous set download myminio/models;
+      echo 'MinIO initialization complete';
+      exit 0;
+      "
```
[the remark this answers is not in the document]
nicolasnoble: Fair point, but this docker-compose.yml is a dev/testing convenience - the hardcoded creds match the defaults. Using `depends_on` with health condition is a nice improvement though. Filed as a follow-up.

Dockerfile
```diff
@@ -23,11 +23,22 @@ FROM nvcr.io/nvidia/base/ubuntu:noble-20250619 AS runtime
...
+        python3 \
+        python3-pip \
+        python3-venv && \
     rm -rf /var/lib/apt/lists/*
 
+# Create Python virtual environment and install sidecar package
+COPY sidecar/pyproject.toml /app/sidecar/
+COPY sidecar/src /app/sidecar/src
+RUN python3 -m venv /app/venv && \
+    /app/venv/bin/pip install --no-cache-dir /app/sidecar
```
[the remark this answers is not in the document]
nicolasnoble: Good catch - it is technically redundant since the package is pip-installed into the venv. Keeping it as belt-and-suspenders for now since it is harmless and helps if someone runs Python outside the venv for debugging. Can clean up later.

modelexpress_common/src/download.rs
```diff
@@ -2,16 +2,31 @@
...
+                Ok(p) => Box::new(p),
+                Err(e) => {
+                    // Log the error and panic - this is a configuration error
+                    tracing::error!("Failed to initialize ModelStreamer provider: {}", e);
+                    panic!("Failed to initialize ModelStreamer provider: {e}");
+                }
+            }
+        }
     }
 }
```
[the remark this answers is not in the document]
nicolasnoble: This is intentional - `get_provider` is called at startup, and a bad ModelStreamer config means the server cannot function. The panic is documented with a `# Panics` doc comment. Changing the return type to `Result` would push error handling to every call site for something that is fundamentally a "fix your config" situation. Happy to revisit if we add runtime provider switching later.

modelexpress_common/src/download.rs
```diff
@@ -96,11 +111,17 @@ mod tests {
...
+    fn test_get_provider_huggingface() {
         let provider = get_provider(ModelProvider::HuggingFace);
         assert_eq!(provider.provider_name(), "Hugging Face");
     }
 
+    #[test]
+    fn test_get_provider_model_streamer() {
+        let provider = get_provider(ModelProvider::ModelStreamer);
+        assert_eq!(provider.provider_name(), "Model Streamer");
+    }
```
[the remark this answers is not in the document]
nicolasnoble: The integration tests already use `@pytest.mark.skipif(not minio_available(), ...)` to skip when MinIO is not running. The CI workflow sets up a real MinIO instance before running these tests, so this is working as intended.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,364 @@
...
+
+        return original_env
+
+    def _restore_credentials(self, original_env: dict[str, str]) -> None:
+        """Restore original environment variables."""
+        for key, value in original_env.items():
+            if value:
+                os.environ[key] = value
+            elif key in os.environ:
+                del os.environ[key]
```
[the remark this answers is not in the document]
nicolasnoble: Agreed this is not ideal long-term. The `asyncio.Lock` serializes credential setup correctly for now - the sidecar is expected to handle one download at a time in practice (model loading is sequential). The boto3 call is offloaded to a thread but the lock covers the full download lifecycle including credential restore.

Passing credentials directly to the boto3 session (instead of env vars) would be cleaner. Deferring to a follow-up since it works correctly for the current single-download use case.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,364 @@
...
+
+                # Download the file
+                local_file = local_path / rel_path
+                local_file.parent.mkdir(parents=True, exist_ok=True)
+
+                logger.info("Downloading: %s -> %s", key, local_file)
+                s3_client.download_file(bucket, key, str(local_file))
+                downloaded_files.append(rel_path)
+
+        return downloaded_files
```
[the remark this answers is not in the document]
nicolasnoble: Good observation. In practice, a failed download will return `success: false` to the Rust server, which will retry the whole operation. The orphaned partial files get overwritten on retry. Adding cleanup-on-failure is a nice hardening improvement but not blocking for the initial implementation.

entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+    echo "Shutting down..."
+    if [ -n "$SIDECAR_PID" ]; then
+        kill "$SIDECAR_PID" 2>/dev/null || true
+    fi
+}
+trap cleanup EXIT INT TERM
+
+# Start the main server (blocking)
+echo "Starting ModelExpress server..."
+exec /app/modelexpress-server "$@"
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. Dropped the `exec` so the shell stays alive and the cleanup trap can fire to kill the sidecar. The server now runs in the background with `wait` to block until it exits.

modelexpress_common/src/config.rs
```diff
@@ -370,6 +370,111 @@ impl ConnectionConfig {
...
+
+    /// AWS secret access key
+    pub secret_access_key: Option<String>,
+
+    /// AWS region
+    pub region: Option<String>,
+
+    /// Custom S3 endpoint (for MinIO or other S3-compatible storage)
+    pub endpoint: Option<String>,
+}
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. Replaced derived `Debug` with a manual impl that redacts `access_key_id` and `secret_access_key` while keeping `region` and `endpoint` visible for debugging.

modelexpress_common/src/config.rs
```diff
@@ -370,6 +370,111 @@ impl ConnectionConfig {
...
+
+/// GCS credentials configuration for Model Streamer
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct GcsCredentials {
+    /// Path to GCS credentials JSON file
+    pub credentials_file: Option<String>,
+
+    /// Base64-encoded credentials JSON (alternative to file)
+    pub credentials_json: Option<String>,
+}
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c alongside S3Credentials. Manual `Debug` impl redacts `credentials_json` while keeping `credentials_file` (just a path) visible.

modelexpress_common/src/providers/model_streamer.rs
```diff
@@ -0,0 +1,527 @@
...
+
+/// Request to download a model from the sidecar
+#[derive(Debug, Serialize)]
+struct DownloadRequest {
+    model_path: String,
+    cache_dir: String,
+    ignore_weights: bool,
+    s3_credentials: Option<S3CredentialsPayload>,
+    gcs_credentials: Option<GcsCredentialsPayload>,
+}
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. `DownloadRequest` now has a manual `Debug` impl that redacts the credential payloads entirely, while keeping `model_path`, `cache_dir`, and `ignore_weights` visible.

sidecar/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+    "pytest-asyncio>=0.21",
+    "httpx>=0.25.0",
+]
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
+[tool.pytest.ini_options]
+asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. Bumped `pytest-asyncio` floor from `>=0.21` to `>=0.24` to match the `asyncio_default_fixture_loop_scope` config option.

sidecar/src/modelexpress_sidecar/api/routes.py
```diff
@@ -0,0 +1,91 @@
...
+        result = await downloader.get_model(model_id, cache_dir)
+        return result
+    except FileNotFoundError:
+        raise HTTPException(status_code=404, detail=f"Model '{model_id}' not found")
+    except Exception as e:
+        logger.exception("Failed to get model: %s", model_id)
+        return GetModelResponse(
+            exists=False,
+            error=str(e),
+        )
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. Both `FileNotFoundError` handlers now use `raise HTTPException(...) from err`.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,364 @@
...
+        parts = model_id.split("/", 2)
+        if len(parts) < 2:
+            raise FileNotFoundError(f"Invalid model_id format: {model_id}")
+
+        scheme = parts[0]
+        bucket = parts[1]
+        path = parts[2] if len(parts) > 2 else ""
+
+        cache_base = Path(cache_dir) if cache_dir else self.default_cache_dir
+        return cache_base / CACHE_SUBDIR / scheme / bucket / path
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. `_resolve_model_id` now resolves the path and checks it stays within the cache directory. If the resolved path escapes the cache base, it raises a `ValueError`.

sidecar/tests/test_s3_integration.py
```diff
@@ -0,0 +1,153 @@
...
+                    "region": os.environ.get("AWS_REGION", "us-east-1"),
+                    "endpoint": endpoint,
+                },
+            },
+        )
+        assert response.status_code == 200
+        data = response.json()
+        # Download may succeed but return empty files, or fail entirely
+        # depending on S3 behavior with non-existent prefixes
+        assert "success" in data
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in cb50a5c. The test now asserts `data["success"] is False` and checks for an `error` key. A non-existent bucket should always fail with boto3.

sidecar/src/modelexpress_sidecar/services/downloader.py
```diff
@@ -0,0 +1,370 @@
...
+        scheme, bucket, prefix = self._parse_model_uri(model_path)
+
+        if scheme != "s3":
+            raise ValueError(f"boto3 download only supports S3, got: {scheme}")
+
+        # Create S3 client
+        endpoint_url = os.environ.get("AWS_ENDPOINT_URL")
+        config = Config(signature_version="s3v4")
+
+        s3_client = boto3.client(
```
erezzarum, line 287: To improve concurrency it could be beneficial to provide a TransferConfig and have user configure it via environment variables - https://docs.aws.amazon.com/boto3/latest/reference/customizations/s3.html#s3-transfers

zhengluo-nv:
Close outdated ModelStreamer PR

PR #137, opened 2026-01-29
feat: Kubernetes Metadata Management

KavinKrishnan:
Adds support for Kubernetes CRD & ConfigMap backed metadata persistence, as an alternative to current Redis backed storage.

Additionally, default configuration is an in-memory cache on model express server - while allowing for optional write-through to Redis or CRD for persistence. 

```
                         MX Server
                            |
                    +-----------------+
                    |   gRPC Service  |
                    |  (P2pService)   |
                    +-----------------+
                            |
                    +-----------------+
                    |  In memory cache |
                    +-----------------+
                    |         \
             +-------+    +----------+
             | Redis |    |   K8s   |
             | (HA)  |    |  CRDs   |
             +-------+    +----------+
              optional       optional
         write-through   write-through

  Source vLLM                              Target vLLM
  +------------+                           +------------+
  | Load model |   NIXL/RDMA              | Query MX   |
  | Register   |--------------------->.   | Receive    |
  | Publish    |   GPU-to-GPU             | Load into  |
  | metadata   |   112+ Gbps              | vLLM model |
  +------------+                            +------------+
       |                                    |
       +--- PublishMetadata (gRPC) ---------+
       +--- PublishReady (gRPC) ------------+
       +--- GetMetadata (gRPC) -------------+
       +--- GetReady (gRPC) ----------------+
```

### DIFF
--- a/modelexpress_client/python/modelexpress/vllm_loader.py
+++ b/modelexpress_client/python/modelexpress/vllm_loader.py
@@ -116,7 +116,6 @@ class MxSourceModelLoader(DefaultModelLoader):
     """
 
     def __init__(self, load_config: LoadConfig):
-        logger.debug("MxSourceModelLoader.__init__ called!")
         # Map mx-source to auto format internally for weight loading
         # We keep our custom loader behavior but use standard weight loading
         import copy
@@ -136,8 +135,6 @@ class MxSourceModelLoader(DefaultModelLoader):
         self, vllm_config: VllmConfig, model_config: ModelConfig
     ) -> nn.Module:
         """Load model with NIXL registration before weight processing."""
-        logger.info("MxSourceModelLoader.load_model() STARTING")
-
         device_config = vllm_config.device_config
         load_config = vllm_config.load_config
         load_device = (
@@ -397,7 +394,6 @@ class MxTargetModelLoader(DummyModelLoader):
     """
 
     def __init__(self, load_config: LoadConfig):
-        logger.debug("MxTargetModelLoader.__init__ called!")
         # Map mx-target to dummy format internally for weight initialization
         import copy
         modified_config = copy.copy(load_config)
@@ -416,9 +412,6 @@ class MxTargetModelLoader(DummyModelLoader):
     ) -> nn.Module:
         """Load model with RDMA transfer before weight processing."""
         load_start = time.perf_counter()
-        logger.info("=" * 60)
-        logger.info("MxTargetModelLoader.load_model() STARTING")
-        logger.info("=" * 60)
 
         device_config = vllm_config.device_config
         load_config = vllm_config.load_config
@@ -430,39 +423,31 @@ class MxTargetModelLoader(DummyModelLoader):
 
         with set_default_torch_dtype(model_config.dtype):
             with target_device:
-                t0 = time.perf_counter()
+                init_start = time.perf_counter()
                 logger.info("[TIMING] Initializing model structure...")
                 model = initialize_model(
                     vllm_config=vllm_config, model_config=model_config
                 )
-                logger.info(f"[TIMING] Model structure initialized in {time.perf_counter() - t0:.2f}s")
+                logger.info(f"[TIMING] Model structure initialized in {time.perf_counter() - init_start:.2f}s")
 
-            t0 = time.perf_counter()
+            dummy_start = time.perf_counter()
             logger.info("[TIMING] Creating dummy weights...")
             self.load_weights(model, model_config)
-            logger.info(f"[TIMING] Dummy weights created in {time.perf_counter() - t0:.2f}s")
+            logger.info(f"[TIMING] Dummy weights created in {time.perf_counter() - dummy_start:.2f}s")
 
-            # HOOK: Receive RAW tensors via RDMA before processing
-            t0 = time.perf_counter()
-            logger.info("=" * 60)
-            logger.info("[TIMING] === HOOK: Receiving raw tensors via RDMA ===")
-            logger.info("=" * 60)
+            rdma_start = time.perf_counter()
+            logger.info("[TIMING] Receiving raw tensors via RDMA...")
             self._receive_raw_tensors(model, target_device)
-            rdma_time = time.perf_counter() - t0
-            logger.info(f"[TIMING] === RDMA reception complete in {rdma_time:.2f}s ===")
+            rdma_time = time.perf_counter() - rdma_start
+            logger.info(f"[TIMING] RDMA reception complete in {rdma_time:.2f}s")
 
-            # Now run FP8 processing (transforms weight_scale_inv -> weight_scale)
-            # This will produce IDENTICAL results to source since we have same raw data
-            t0 = time.perf_counter()
+            process_start = time.perf_counter()
             logger.info("[TIMING] Processing weights (FP8 transformation)...")
             process_weights_after_loading(model, model_config, target_device)
-            logger.info(f"[TIMING] Weight processing complete in {time.perf_counter() - t0:.2f}s")
+            logger.info(f"[TIMING] Weight processing complete in {time.perf_counter() - process_start:.2f}s")
 
         total_time = time.perf_counter() - load_start
-        logger.info("=" * 60)
-        logger.info("[TIMING] MxTargetModelLoader.load_model() COMPLETE")
-        logger.info(f"[TIMING] Total load time: {total_time:.2f}s")
-        logger.info("=" * 60)
+        logger.info(f"[TIMING] MxTargetModelLoader.load_model() complete in {total_time:.2f}s")
         return model.eval()
 
     def _receive_raw_tensors(
@@ -491,7 +476,7 @@ class MxTargetModelLoader(DummyModelLoader):
             return
 
         # Collect target tensors (these have dummy data)
-        t0 = time.perf_counter()
+        collect_start = time.perf_counter()
         target_tensors: dict[str, torch.Tensor] = {}
         for name, param in model.named_parameters():
             if param.is_cuda:
@@ -500,7 +485,7 @@ class MxTargetModelLoader(DummyModelLoader):
         device_id = _get_worker_rank(device)
         scale_count = sum(1 for n in target_tensors if "scale_inv" in n.lower())
         total_size = sum(t.numel() * t.element_size() for t in target_tensors.values())
-        logger.debug(f"[Worker {device_id}] [TIMING] Collected {len(target_tensors)} tensors in {time.perf_counter() - t0:.3f}s")
+        logger.debug(f"[Worker {device_id}] [TIMING] Collected {len(target_tensors)} tensors in {time.perf_counter() - collect_start:.3f}s")
 
         logger.info(
             f"[Worker {device_id}] Target has {len(target_tensors)} tensors "
@@ -520,7 +505,7 @@ class MxTargetModelLoader(DummyModelLoader):
             logger.debug(f"[Worker {device_id}]   '{name}': shape={t.shape}, dtype={t.dtype}, checksum={checksum}")
 
         # Initialize NIXL manager
-        t0 = time.perf_counter()
+        nixl_start = time.perf_counter()
         agent_name = f"mx-target-worker{device_id}-{uuid.uuid4().hex[:8]}"
         logger.debug(f"[Worker {device_id}] [TIMING] Initializing NIXL manager with agent_name={agent_name}")
         self._nixl_manager = NixlTransferManager(
@@ -528,13 +513,13 @@ class MxTargetModelLoader(DummyModelLoader):
             device_id=device_id,
         )
         self._nixl_manager.initialize()
-        nixl_init_time = time.perf_counter() - t0
+        nixl_init_time = time.perf_counter() - nixl_start
         logger.info(f"[Worker {device_id}] [TIMING] NIXL manager initialized in {nixl_init_time:.3f}s")
 
-        t0 = time.perf_counter()
+        reg_start = time.perf_counter()
         logger.debug(f"[Worker {device_id}] [TIMING] Registering target tensors with NIXL...")
         self._nixl_manager.register_tensors(target_tensors)
-        reg_time = time.perf_counter() - t0
+        reg_time = time.perf_counter() - reg_start
         logger.info(f"[Worker {device_id}] [TIMING] Target tensors registered in {reg_time:.3f}s")
 
         # COORDINATION: Wait for NIXL ready flag before initiating transfer
@@ -665,7 +650,6 @@ class MxTargetModelLoader(DummyModelLoader):
                     transfer_time = time.perf_counter() - transfer_start
 
                     bandwidth_gbps = (bytes_transferred * 8) / (transfer_time * 1e9) if transfer_time > 0 else 0
-                    logger.info("=" * 60)
                     logger.info(
                         f"[Worker {device_id}] [TIMING] RDMA TRANSFER COMPLETE:"
                     )
@@ -673,7 +657,6 @@ class MxTargetModelLoader(DummyModelLoader):
                     logger.info(f"[Worker {device_id}] [TIMING]   Data: {bytes_transferred / 1e9:.2f} GB")
                     logger.info(f"[Worker {device_id}] [TIMING]   Time: {transfer_time:.3f}s")
                     logger.info(f"[Worker {device_id}] [TIMING]   Bandwidth: {bandwidth_gbps:.1f} Gbps")
-                    logger.info("=" * 60)
 
                     # Sync CUDA to ensure transfer is visible
                     torch.cuda.synchronize()
@@ -746,14 +729,12 @@ class MxTargetModelLoader(DummyModelLoader):
 
             # Final timing summary
             total_receive_time = time.perf_counter() - receive_start
-            logger.info("=" * 60)
             logger.info(f"[Worker {device_id}] [TIMING] _receive_raw_tensors SUMMARY:")
             logger.info(f"[Worker {device_id}] [TIMING]   Total time: {total_receive_time:.2f}s")
             logger.info(f"[Worker {device_id}] [TIMING]   NIXL init: {nixl_init_time:.3f}s")
             logger.info(f"[Worker {device_id}] [TIMING]   Tensor reg: {reg_time:.3f}s")
             logger.info(f"[Worker {device_id}] [TIMING]   Wait for source: {wait_time:.2f}s")
             logger.info(f"[Worker {device_id}] [TIMING]   RDMA transfer: {transfer_time:.3f}s")
-            logger.info("=" * 60)
 
         except Exception as e:
             import traceback


examples/p2p_transfer_k8s/deploy/persistence/README.md
```diff
@@ -0,0 +1,81 @@
...
+│   └────────────────────┬────────────────────────┘    │
+│                        │ write-through (optional)     │
+│              ┌─────────┴──────────┐                  │
+│              ▼                    ▼                   │
+│   ┌──────────────────┐ ┌────────────────────┐        │
+│   │   Redis Backend  │ │ Kubernetes CRD     │        │
+│   │   (HA, fast)     │ │ Backend (native)   │        │
+│   └──────────────────┘ └────────────────────┘        │
+└──────────────────────────────────────────────────────┘
+```
```
nv-hwoo, line 25: nit: align the box lines

modelexpress_client/python/modelexpress/metadata_client.py
```diff
@@ -0,0 +1,552 @@
...
+Supports multiple backends:
+- Redis: Original implementation, stores JSON in Redis keys
+- Kubernetes: Uses CRDs and ConfigMaps for native K8s integration
+- grpc / memory: Uses ModelExpress server gRPC (PublishReady/GetReady)
+
+The backend is selected via the MX_METADATA_BACKEND environment variable:
+- "redis" (default): Use Redis backend
+- "kubernetes", "k8s", "crd": Use Kubernetes CRD backend
+- "memory", "grpc": Use gRPC to ModelExpress server (for in-memory server)
+"""
```
nicolasnoble: Heads up, the Python client here defaults to `"redis"`:
```python
os.environ.get("MX_METADATA_BACKEND", "redis")
```

but the Rust server (`BackendConfig::from_env()` in `metadata_backend.rs`) defaults to `"memory"`. So if someone deploys without explicitly setting `MX_METADATA_BACKEND`, the server will be using in-memory while the client tries to talk to Redis. Would it make sense to align these, maybe default to `"memory"` or `"grpc"` here too?

nicolasnoble: Heads up, the Python client here defaults to `"redis"`:
```python
os.environ.get("MX_METADATA_BACKEND", "redis")
```

but the Rust server (`BackendConfig::from_env()` in `metadata_backend.rs`) defaults to `"memory"`. So if someone deploys without explicitly setting `MX_METADATA_BACKEND`, the server will be using in-memory while the client tries to talk to Redis. Would it make sense to align these, maybe default to `"memory"` or `"grpc"` here too?

examples/p2p_transfer_k8s/deploy/vllm-target.yaml
```diff
@@ -35,32 +38,48 @@ spec:
...
               value: "1"
+            - name: MX_EXPECTED_WORKERS
+              value: "8"
+            # Metadata backend: "redis" (default) or "kubernetes" (CRD-based)
+            # Set to "kubernetes" to use native K8s CRDs instead of Redis
+            - name: MX_METADATA_BACKEND
+              value: "redis"
+            # Redis backend configuration
+            - name: MX_REDIS_HOST
+              value: "modelexpress-server"
```
nicolasnoble: The file header (line 4) says _"Uses Kubernetes CRD backend for metadata (default)"_, and the comment just above says _"Set to \"kubernetes\" to use native K8s CRDs instead of Redis"_, but the actual value here is `"redis"`. Is this a copy-paste leftover? If this file is meant to be the CRD variant, this should probably be `"kubernetes"` and the `MX_REDIS_HOST` below could be dropped too.

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -0,0 +1,401 @@
...
+                    ..Default::default()
+                },
+                spec: ModelMetadataSpec {
+                    model_name: model_name.to_string(),
+                    expected_workers: 8, // Default
+                },
+                status: None,
+            };
+
+            api.create(&PostParams::default(), &new_cr).await?;
```
nicolasnoble: Did you think about the case where two workers call `publish_metadata` at the same time? Both could see `existing.is_none() == true`, and the second `api.create()` would get a 409, which isn't caught here (unlike the ConfigMap upsert at line ~108 which handles 409 nicely).

There's also a subtler issue with the read-modify-write on lines 226-270: two concurrent workers both read the same worker list, both merge in their data, and the second patch wins, quietly dropping the first worker's metadata.

The Redis backend handles this with a Lua script for atomicity. For the K8s path, a couple options:
1. Catch 409 on CR create (like the ConfigMap path already does)
2. Use `resourceVersion`-based conflict detection + retry for the status patch

Or server-side apply could sidestep the create race entirely.

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -0,0 +1,401 @@
...
+                            cr_name.clone(), // Use sanitized name for label value
+                        );
+                        labels
+                    }),
+                    ..Default::default()
+                },
+                spec: ModelMetadataSpec {
+                    model_name: model_name.to_string(),
+                    expected_workers: 8, // Default
+                },
```
nicolasnoble, line 227: Is `expected_workers: 8` meant to be a placeholder? Different deployments will have different TP sizes, so this might want to come from config (e.g. `MX_EXPECTED_WORKERS` env var) or be passed as a parameter.

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -0,0 +1,401 @@
...
+        let tensors = tensor_descs
+            .into_iter()
+            .map(|t| TensorRecord {
+                name: t.name,
+                addr: t.addr.parse().unwrap_or(0),
+                size: t.size.parse().unwrap_or(0),
+                device_id: t.device_id,
+                dtype: t.dtype,
+            })
+            .collect();
```
nicolasnoble:
```rust
addr: t.addr.parse().unwrap_or(0),
size: t.size.parse().unwrap_or(0),
```

Did you consider what happens if these parses actually fail? An address of `0` would be a null pointer and RDMA would fail in a way that's pretty hard to track down. Might be safer to propagate the error here instead of silently falling back to 0.

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -0,0 +1,401 @@
...
+            workers,
+            published_at,
+        }))
+    }
+
+    async fn remove_metadata(&self, model_name: &str) -> MetadataResult<()> {
+        let cr_name = sanitize_model_name(model_name);
+
+        // Delete the CR (ConfigMaps will be orphaned, could add owner references)
+        let api = self.model_metadata_api();
```
nicolasnoble, line 438: The comment here mentions ConfigMaps will be orphaned. Did you get a chance to look into setting `ownerReferences` on them? If the server crashes before the cleanup loop at lines 368-380 finishes, those ConfigMaps would stick around forever. With owner references pointing to the parent `ModelMetadata` CR, K8s garbage collection would clean them up automatically.

modelexpress_client/python/modelexpress/metadata_client.py
```diff
@@ -0,0 +1,552 @@
...
+        worker_id: int,
+    ) -> WorkerReadyInfo | None:
+        try:
+            response = self._get_client().get_ready(model_name, worker_id)
+            if response.found and response.ready:
+                return WorkerReadyInfo(
+                    session_id=response.session_id,
+                    nixl_ready=True,
+                    stability_verified=response.stability_verified,
+                    metadata_hash=response.metadata_hash or None,
```
nicolasnoble: I don't think `GetReadyResponse` has a `stability_verified` field. The proto has `found`, `ready`, `session_id`, `metadata_hash`, and `timestamp`. Python protobuf will just silently return `False` for the missing attribute, so `WorkerReadyInfo.stability_verified` will always be `False` here.

It's currently masked because `wait_for_ready()` goes through a different path, but it'd bite anyone calling `get_ready_signal()` directly.

modelexpress_client/python/modelexpress/metadata_client.py
```diff
@@ -0,0 +1,552 @@
...
+                    metadata_hash=info.get("metadata_hash"),
+                    timestamp=info.get("timestamp"),
+                )
+        except Exception as e:
+            logger.warning(f"[Worker {worker_id}] Error getting ready signal: {e}")
+        
+        return None
+    
+    def wait_for_ready(
+        self,
```
nicolasnoble:
```python
return True, None, None
```

This returns "ready" when the backend is unavailable, which means the target would go ahead without actually knowing if the source is ready. Could that lead to serving with dummy weights?

Would it be safer to return `(False, None, None)` or raise here? Looks like `KubernetesBackend.wait_for_ready()` might have the same pattern if it exists.

examples/p2p_transfer_k8s/deploy/persistence/rbac-modelmetadata.yaml
```diff
@@ -0,0 +1,91 @@
...
+#
+# RBAC resources for ModelExpress CRD-based metadata
+# Required when using MX_METADATA_BACKEND=kubernetes
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: modelexpress
+  namespace: kavin
+  labels:
```
nicolasnoble, line 12: Looks like `kavin` is a personal namespace that snuck in here. It shows up on lines 12, 21, 64, and 75. Want to swap it for something generic like `modelexpress` or `default`?

examples/p2p_transfer_k8s/deploy/persistence/rbac-modelmetadata.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - modelmetadatas/status
+    verbs:
+      - get
+      - update
+      - patch
+  # ConfigMaps for tensor descriptors
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
```
nicolasnoble, line 52: This grants full CRUD on all ConfigMaps in the namespace, not just the ModelExpress ones. That could be a concern in shared namespaces. Worth at least calling it out with a comment, or maybe mentioning it in the README?

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -55,6 +55,11 @@
 logger = logging.getLogger("modelexpress.vllm_loader")
 logger.setLevel(logging.DEBUG)
 
+def _log(msg: str, level: str = "INFO") -> None:
```
nicolasnoble: Any reason to use raw `print()` here instead of the `logger` already defined on line 55? The comment says "k8s visibility" but you could get the same thing by configuring the logger with a `StreamHandler` to stdout. With `print()` you lose log level filtering, aggregation framework integration, and structured parsing, plus you're duplicating the timestamp/level formatting that logging already gives you.

modelexpress_client/python/modelexpress/metadata_client.py
```diff
@@ -0,0 +1,552 @@
...
+- "redis" (default): Use Redis backend
+- "kubernetes", "k8s", "crd": Use Kubernetes CRD backend
+- "memory", "grpc": Use gRPC to ModelExpress server (for in-memory server)
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import logging
```
nicolasnoble: Nit: `base64` (line 22) and `Any` from `typing` (line 28) look unused.

examples/p2p_transfer_k8s/deploy/persistence/redis-standalone.yaml
```diff
@@ -0,0 +1,73 @@
...
+    metadata:
+      labels:
+        app: redis
+    spec:
+      containers:
+        - name: redis
+          image: redis:7-alpine
+          args: ["--appendonly", "yes", "--dir", "/data"]
+          ports:
+            - containerPort: 6379
```
nicolasnoble, line 58: This deploys Redis without authentication, so any pod in the namespace can connect. Is that intentional for dev/testing, or would it be worth adding `--requirepass` with a Secret reference?

examples/p2p_transfer_k8s/deploy/persistence/redis-standalone.yaml
```diff
@@ -0,0 +1,73 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: redis-data
+  labels:
+    app: redis
+spec:
+  accessModes:
+    - ReadWriteOnce
+  storageClassName: csi-mounted-fs-path-sc
```
nicolasnoble: `csi-mounted-fs-path-sc` is pretty environment-specific. Most clusters won't have it. Maybe drop this line to use the cluster default, or add a comment noting it needs to be changed?

modelexpress_client/python/modelexpress/metadata_client.py
```diff
@@ -0,0 +1,552 @@
...
+            cr = api.get_namespaced_custom_object(
+                group="modelexpress.nvidia.com",
+                version="v1alpha1",
+                namespace=self._namespace,
+                plural="modelmetadatas",
+                name=cr_name,
+            )
+            
+            # Update worker status
+            workers = cr.get("status", {}).get("workers", [])
```
nicolasnoble: Same concurrency question as the Rust side. Did you consider what happens when two workers call `publish_ready_signal()` at the same time? The second patch would overwrite the first since there's no `resourceVersion` check. You could grab the `resourceVersion` from the GET and include it in the PATCH, then retry on 409.

modelexpress_client/python/modelexpress/metadata_client.py
```diff
@@ -0,0 +1,552 @@
...
+    
+    @staticmethod
+    def _sanitize_model_name(model_name: str) -> str:
+        """Convert model name to valid K8s resource name."""
+        return (
+            model_name.lower()
+            .replace("/", "-")
+            .replace("_", "-")
+            .strip("-")
+        )
```
nicolasnoble: The Rust version of this in `k8s_types.rs` also filters out any character that isn't `[a-z0-9\-.]`:

```rust
.chars()
.filter(|c| c.is_ascii_alphanumeric() || *c == '-' || *c == '.')
```

Without that step here, a model name like `meta-llama/Llama@3.1+8B` would give you `meta-llama-llama3.18b` on the server but `meta-llama-llama@3.1+8b` on the client. The CRD lookup would silently miss. Worth adding the equivalent filter?

```python
import re
re.sub(r"[^a-z0-9\-.]", "", ...)
```

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -361,12 +371,12 @@ def _publish_metadata_to_server(
...
             else:
-                logger.error(f"[Worker {device_id}] FAILED to publish metadata")
+                _log(f"[Worker {device_id}] FAILED to publish metadata", "ERROR")
 
         except Exception as e:
             import traceback
-            logger.error(f"[Worker {device_id}] EXCEPTION publishing metadata: {e}")
-            logger.error(f"[Worker {device_id}] Traceback: {traceback.format_exc()}")
+            _log(f"[Worker {device_id}] EXCEPTION publishing metadata: {e}", "ERROR")
+            _log(f"[Worker {device_id}] Traceback: {traceback.format_exc()}", "ERROR")
```
nicolasnoble: This exception is logged but not re-raised, so the source finishes loading normally even though no target will ever be able to discover it. Targets will sit there polling for up to 2 hours before timing out.

Could this re-raise instead? Or maybe add some retry logic? The `_receive_raw_tensors` path on line 680 does re-raise, so there's already precedent in this file for not swallowing these.

modelexpress_server/src/state.rs
```diff
@@ -167,52 +30,83 @@ pub struct ReadyRecord {
...
+        // Need to connect
         self.connect().await?;
 
-        let guard = self.redis.read().await;
+        let guard = self.backend.read().await;
         guard
             .as_ref()
             .cloned()
-            .ok_or_else(|| "Redis connection not available".into())
+            .ok_or_else(|| "Backend not available".into())
```
nicolasnoble: Did you consider the case where two callers hit this at the same time? Both would read `None`, both drop the read lock, and both call `connect()`, creating duplicate backends.

The usual fix is a write-lock with double-check:

```rust
async fn get_backend(&self) -> MetadataResult<Arc<dyn MetadataBackend>> {
    {
        let guard = self.backend.read().await;
        if let Some(backend) = guard.as_ref() {
            return Ok(backend.clone());
        }
    }
    let mut guard = self.backend.write().await;
    // Double-check after acquiring write lock
    if let Some(backend) = guard.as_ref() {
        return Ok(backend.clone());
    }
    let backend = create_backend(self.config.clone()).await?;
    *guard = Some(backend.clone());
    Ok(backend)
}
```

Same pattern applies to `RedisBackend::get_conn()` in `redis.rs` if it has the same structure.

nicolasnoble (commented):
So, I think this looks good in general, but I ran into a few things that I think are worth looking at before merging. Most are around edge cases and concurrency, plus a couple of config bugs. See inline comments for context on each one.

Also:

- No tests for kubernetes.rs or metadata_client.py
- Layered backend test only covers memory-only mode (no write-through or hydration)
- test_backend_config_from_env uses unsafe env manipulation that could interfere with parallel tests

modelexpress_client/python/modelexpress/metadata_client.py
nv-hwoo, line 1: Where and how is this metadata client used exactly?

examples/p2p_transfer_k8s/deploy/vllm-target.yaml
```diff
@@ -69,7 +88,7 @@ spec:
             - name: NIXL_LOG_LEVEL
               value: "INFO"
             - name: UCX_LOG_LEVEL
-              value: "INFO"
+              value: "WARN"
```
nv-hwoo:
```suggestion
              value: "INFO"
```

examples/p2p_transfer_k8s/deploy/vllm-target.yaml
```diff
@@ -35,32 +38,48 @@ spec:
...
             # OPTIMIZATION: Synchronized start - wait for all workers before transferring
             # This maximizes RDMA parallelism by ensuring all workers start simultaneously
             - name: MX_SYNC_START
               value: "1"
+            - name: MX_EXPECTED_WORKERS
+              value: "8"
+            # Metadata backend: "redis" (default) or "kubernetes" (CRD-based)
+            # Set to "kubernetes" to use native K8s CRDs instead of Redis
+            - name: MX_METADATA_BACKEND
+              value: "redis"
```
nv-hwoo: The source node is deployed with kubernetes and target node is deployed as redis. Is this intended?

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
```diff
@@ -0,0 +1,170 @@
...
+              value: "1"
+            # Server address for metadata publishing
+            - name: MX_SERVER_ADDRESS
+              value: "modelexpress-server:8001"
+            # OPTIMIZATION: Synchronized publish - wait for all workers before publishing
+            # This reduces the staggered window and allows targets to start simultaneously
+            - name: MX_SYNC_PUBLISH
+              value: "1"
+            - name: MX_EXPECTED_WORKERS
+              value: "8"
```
nv-hwoo, line 73: Do we need to manually specify the num of expected worker? (meaning does the user have to change this everytime the config changes?)

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
```diff
@@ -0,0 +1,170 @@
...
+              value: "modelexpress-server"
+            # Baseline: individual tensor registration (stable and tested)
+            # Set to "1" for contiguous regions (experimental - see docs/CONTIGUOUS_CONTEXT.md)
+            - name: MX_CONTIGUOUS_REG
+              value: "0"
+            # Logging (set to DEBUG for troubleshooting)
+            - name: NIXL_LOG_LEVEL
+              value: "INFO"
+            - name: UCX_LOG_LEVEL
+              value: "WARN"
```
nv-hwoo:
```suggestion
              value: "INFO"
```

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
```diff
@@ -0,0 +1,170 @@
...
+              session_id = str(uuid.uuid4())
+              client = redis.Redis(host=host, port=6379, decode_responses=True)
+              workers = int(os.environ.get("MX_EXPECTED_WORKERS", "8"))
+              for w in range(workers):
+                  key = f"mx:nixl_ready:{model}:worker:{w}"
+                  value = json.dumps({"session_id": session_id, "nixl_ready": True, "stability_verified": True})
+                  client.setex(key, 14400, value)
+                  print(f"Published ready for worker {w}")
+              '
+              ) &
```
nv-hwoo, line 128: Do we need this check here? Why can't this ready signal be handled internally?

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -0,0 +1,401 @@
...
+            }
+        };
+
+        // Reconstruct workers from status + ConfigMaps
+        let mut workers = Vec::new();
+        for worker_status in status.workers {
+            // Decode NIXL metadata
+            let nixl_metadata = BASE64
+                .decode(&worker_status.nixl_metadata)
+                .unwrap_or_default();
```
nv-hwoo: `unwrap_or_default()` could silently swallow corrupted NIXL metadata. Instead we should propagate the error.

modelexpress_client/python/modelexpress/vllm_loader.py
nv-hwoo: +1. Also I was able to see these logs using logger

docs/metadata.md
```diff
@@ -0,0 +1,520 @@
...
+### Custom Resource Definitions
+
+#### ModelMetadata CRD
+
+```yaml
+apiVersion: modelexpress.nvidia.com/v1alpha1
+kind: ModelMetadata
+metadata:
+  name: deepseek-ai-deepseek-v3  # Sanitized model name
+  namespace: kavin
```
nv-hwoo: let's remove hardcoded namespace

modelexpress_server/src/metadata_backend/memory.rs
```diff
@@ -0,0 +1,234 @@
...
+        Self::new()
+    }
+}
+
+#[async_trait]
+impl MetadataBackend for InMemoryBackend {
+    async fn connect(&self) -> MetadataResult<()> {
+        info!("In-memory metadata backend initialized");
+        Ok(())
+    }
```
nv-hwoo, line 47: this looks like an empty function

examples/p2p_transfer_k8s/deploy/persistence/rbac-modelmetadata.yaml
KavinKrishnan: Added a comment to the top of this file

modelexpress_client/python/modelexpress/vllm_loader.py
KavinKrishnan: using logger now

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: removed

examples/p2p_transfer_k8s/deploy/persistence/redis-standalone.yaml
KavinKrishnan: Yes it intentional, I also added a comment in this file as well

examples/p2p_transfer_k8s/deploy/vllm-target.yaml
KavinKrishnan: resolved

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: grpc now

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan: we accept expected_workers as a parameter  now

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan: Now if two workers call publish_metadata simultaneously and the second one gets a 409 on api.create(), it catches the error and falls through to the update path instead of crashing

we now have a status patch which now reads the current resourceVersion before patching. If a concurrent update causes a 409 conflict, it retries up to 5 times with exponential backoff

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan: we propagate the parse error instead of defaulting to 0.

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan: Added ownerReferences on ConfigMaps when creating them, referencing the parent ModelMetadata CR

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: Now return (False, None, None)

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: the stability_verified field was removed from WorkerReadyInfo and all references

examples/p2p_transfer_k8s/deploy/persistence/rbac-modelmetadata.yaml
KavinKrishnan: replaced

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: Included resourceVersion check from the GET in the PATCH, retry on 409.

examples/p2p_transfer_k8s/deploy/persistence/redis-standalone.yaml
KavinKrishnan: Removed the storageClassName line

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: added re.sub(r"[^a-z0-9\-.]", "", ...) to the Python sanitizer

modelexpress_client/python/modelexpress/vllm_loader.py
KavinKrishnan: Re-raised the exception after logging

modelexpress_server/src/state.rs
KavinKrishnan: Implement this

examples/p2p_transfer_k8s/deploy/persistence/README.md
KavinKrishnan: fixed

modelexpress_client/python/modelexpress/metadata_client.py
KavinKrishnan: removed, it was initially used for manually testing readiness

examples/p2p_transfer_k8s/deploy/vllm-target.yaml
KavinKrishnan: changed

examples/p2p_transfer_k8s/deploy/vllm-target.yaml
KavinKrishnan: This was a copy-paste error, fixed above for Nicos comment

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
KavinKrishnan: Currently yes, MX_EXPECTED_WORKERS must match --tensor-parallel-size. The source loader auto-detects from vLLM's TP world size at runtime (see _get_expected_workers() in vllm_loader.py), but the deploy script needs it for the ready-publish loop

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
KavinKrishnan: The ready signal requires: (1) vLLM health endpoint returning 200, (2) a warmup inference request, (3) a grace period for stability. The source loader finishes before vLLM is fully healthy (it runs during weight loading), so it cannot publish ready at that point. A cleaner approach would be a vLLM post-startup hook or a sidecar, but the bash script is the pragmatic solution for now

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
KavinKrishnan: fixed

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan: Replaced unwrap_or_default() with map_err(...)

docs/metadata.md
KavinKrishnan: Replaced all namespace: kavin with namespace: default

modelexpress_server/src/metadata_backend/memory.rs
KavinKrishnan: This is intentional. The MetadataBackend trait requires a connect() method for backends that need initialization

examples/p2p_transfer_k8s/deploy/persistence/vllm-source-redis.yaml
nv-hwoo: Followup question: 

Why do we need to wait until vLLM warm up? The current vllm loader doesn't wait until the end: https://github.com/ai-dynamo/modelexpress/blob/ce148089f34a5ffdcc2f9c84bc562d71a836c88b/modelexpress_client/python/modelexpress/vllm_loader.py#L347-L362
It publishes ready as soon as it publishes metadata, and target loader can immediate use these metadata.

KavinKrishnan: ahh I misread your comment, good catch

removed it from the 3 .yaml scripts

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -0,0 +1,475 @@
...
+                    ))
+                    .await;
+                    continue;
+                }
+                Err(e) => return Err(e.into()),
+            }
+        }
+
+        let total_tensors: usize = worker_records.iter().map(|w| w.tensors.len()).sum();
+        info!(
```
nicolasnoble, line 353: The retry loop silently succeeds if all 5 attempts are exhausted. After the `for` loop ends, execution falls through to the success log and `Ok(())`. If we burned through 5 consecutive 409s, the metadata wasn't actually published but the caller thinks it was. Did you consider adding an error return after the loop? Something like:

```rust
// after the for loop
return Err("Failed to update status after max retries due to conflicts".into());
```

And make the `Ok(_) => break` path the only way to exit successfully.

modelexpress_server/src/metadata_backend/redis.rs
```diff
@@ -0,0 +1,349 @@
...
+        }
+        drop(guard);
+
+        // Need to reconnect
+        self.connect().await?;
+
+        let guard = self.redis.read().await;
+        guard
+            .as_ref()
+            .cloned()
```
nicolasnoble: This still has the same race I flagged on the old `P2pStateManager::get_conn()`. Two callers can both read `None`, both drop the read lock, and both call `connect()`. The double-check locking fix was applied to `P2pStateManager::get_backend()` in `state.rs` but not mirrored here. Same pattern would work:

```rust
async fn get_conn(&self) -> MetadataResult<ConnectionManager> {
    {
        let guard = self.redis.read().await;
        if let Some(conn) = guard.as_ref() {
            return Ok(conn.clone());
        }
    }
    let mut guard = self.redis.write().await;
    if let Some(conn) = guard.as_ref() {
        return Ok(conn.clone());
    }
    let client = redis::Client::open(self.redis_url.as_str())?;
    let conn = ConnectionManager::new(client).await?;
    *guard = Some(conn.clone());
    Ok(conn)
}
```

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan: Added status_updated flag; if all 5 retries are exhausted without success, returns an explicit error. The Ok(_) path is now the only way to exit successfully.

modelexpress_server/src/metadata_backend/redis.rs
KavinKrishnan: Applied the same double-checked locking pattern as state.rs::get_backend(): read lock first, if None, acquire write lock, double-check, then create connection. No more duplicate ConnectionManager on concurrent access

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -477,12 +484,14 @@ def _receive_raw_tensors(
...
 
+        logger.debug(f"Model: {model_name}")
+
         if not model_name:
             logger.warning("MODEL_NAME not set, skipping transfer")
             return
 
         # Collect target tensors (these have dummy data)
-        collect_tensors_start = time.perf_counter()
+        t0 = time.perf_counter()
```
nv-hwoo: can we revert these back? I don't see much value in using `t0` as var name for all counters

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -133,11 +130,14 @@ def __init__(self, load_config: LoadConfig):
...
         self._raw_tensors: dict[str, torch.Tensor] = {}
         self._mx_client = MxClient()
+        logger.debug("MxSourceModelLoader initialized successfully")
 
     def load_model(
         self, vllm_config: VllmConfig, model_config: ModelConfig
     ) -> nn.Module:
         """Load model with NIXL registration before weight processing."""
+        logger.info("MxSourceModelLoader.load_model() STARTING")
+
```
nv-hwoo: I think this is unnecessary log..

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -120,6 +116,7 @@ class MxSourceModelLoader(DefaultModelLoader):
     """
 
     def __init__(self, load_config: LoadConfig):
+        logger.debug("MxSourceModelLoader.__init__ called!")
```
nv-hwoo: I don't think this debug is much useful

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -396,6 +397,7 @@ class MxTargetModelLoader(DummyModelLoader):
     """
 
     def __init__(self, load_config: LoadConfig):
+        logger.debug("MxTargetModelLoader.__init__ called!")
```
nv-hwoo: same as above

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -407,12 +409,16 @@ def __init__(self, load_config: LoadConfig):
...
+        logger.debug("MxTargetModelLoader initialized successfully")
 
     def load_model(
         self, vllm_config: VllmConfig, model_config: ModelConfig
     ) -> nn.Module:
         """Load model with RDMA transfer before weight processing."""
         load_start = time.perf_counter()
+        logger.info("=" * 60)
+        logger.info("MxTargetModelLoader.load_model() STARTING")
+        logger.info("=" * 60)
```
nv-hwoo: same as above

PR #139, opened 2026-02-02
fix: handle empty files and skip dotfiles when downloading from HF

nv-hwoo:
### Summary
Skip dotfiles and handle empty files gracefully when downloading from HuggingFace.

### Problem
Downloads fail with HTTP 416 (Range Not Satisfiable) for empty files like lock.lock or .gitkeep
because range requests are invalid on 0-byte content.

### Changes
 - `providers.rs`: `is_ignored()` now skips all dotfiles (files starting with `.`)
 - `huggingface.rs`: Catch HTTP 416 errors and skip empty files instead of failing
 - Added `test_download_ignores_dotfiles` to verify dotfiles are not downloaded

## Release Notes

* **Bug Fixes**
  * Enhanced error handling when downloading models containing empty files; HTTP errors are now managed gracefully, allowing downloads to continue instead of halting.
  * Improved file filtering during model downloads to properly ignore dotfiles (files beginning with a dot) while ensuring regular configuration files are processed normally.

### DIFF
--- a/modelexpress_common/src/providers.rs
+++ b/modelexpress_common/src/providers.rs
@@ -35,7 +35,11 @@ pub trait ModelProviderTrait: Send + Sync {
         Self: Sized,
     {
         const DEFAULT_IGNORED: [&str; 1] = ["README.md"];
-        filename.starts_with('.') || DEFAULT_IGNORED.contains(&filename)
+        let name = std::path::Path::new(filename)
+            .file_name()
+            .and_then(|s| s.to_str())
+            .unwrap_or(filename);
+        name.starts_with('.') || DEFAULT_IGNORED.contains(&name)
     }
 
     /// Check if a file is an image file that should be ignored
@@ -104,8 +108,13 @@ mod tests {
         assert!(HuggingFaceProvider::is_ignored(".gitkeep"));
         assert!(HuggingFaceProvider::is_ignored(".hidden"));
 
+        // Dotfiles in subdirectories
+        assert!(HuggingFaceProvider::is_ignored("subdir/.gitkeep"));
+        assert!(HuggingFaceProvider::is_ignored("a/b/.hidden"));
+
         // Explicit files
         assert!(HuggingFaceProvider::is_ignored("README.md"));
+        assert!(HuggingFaceProvider::is_ignored("subdir/README.md"));
 
         // (Not Ignored) Regular files
         assert!(!HuggingFaceProvider::is_ignored("model.bin"));


modelexpress_common/src/providers.rs
```diff
@@ -29,13 +29,13 @@ pub trait ModelProviderTrait: Send + Sync {
...
-    /// Default implementation ignores common repository metadata files
+    /// Default implementation ignores dotfiles and common repository metadata files
     fn is_ignored(filename: &str) -> bool
     where
         Self: Sized,
     {
-        const DEFAULT_IGNORED: [&str; 3] = [".gitattributes", ".gitignore", "README.md"];
-        DEFAULT_IGNORED.contains(&filename)
+        const DEFAULT_IGNORED: [&str; 1] = ["README.md"];
+        filename.starts_with('.') || DEFAULT_IGNORED.contains(&filename)
```
AndyDai-nv: coderabbit's point looks valid here:)

nv-hwoo: We skip all subdirectories so we won't really encounter this, but I agree it's a good practice :) Applied the fix.

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -132,6 +132,15 @@ impl ModelProviderTrait for HuggingFaceProvider {
...
                 Err(e) => {
+                    // HTTP 416 (Range Not Satisfiable) occurs for empty files (0 bytes)
+                    // since range requests are invalid on empty content. Skip gracefully.
+                    if e.to_string().contains("416") {
+                        warn!(
+                            "Skipping empty file '{}' from model '{}': {}",
+                            sib.rfilename, model_name, e
+                        );
+                        continue;
+                    }
```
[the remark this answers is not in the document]
grahamking: @nv-hwoo This sounds like great advice. Are you able to incorporate it?

modelexpress_common/src/providers/huggingface.rs
```diff
@@ -132,6 +132,17 @@ impl ModelProviderTrait for HuggingFaceProvider {
                     files_downloaded = true;
                 }
                 Err(e) => {
+                    // HTTP 416 (Range Not Satisfiable) occurs for empty files (0 bytes)
+                    // since range requests are invalid on empty content. Skip gracefully.
+                    if let ApiError::RequestError(req_err) = &e
+                        && req_err.status().is_some_and(|s| s.as_u16() == 416)
+                    {
+                        warn!(
+                            "Skipping empty file '{}' from model '{}': {}",
```
grahamking, line 141: Here you have `{}", var`, but below you do `{var}`. Could you make it consistent? `{var}` is the more modern construction.

PR #143, opened 2026-02-10
fix: disable automountServiceAccountToken by default

nicolasnoble:
The server does not need Kubernetes API access, so there is no reason to mount the service account token into pods. Setting automount to false reduces attack surface, resolving S6865.

## Summary by CodeRabbit

* **Chores**
  * Changed default service account token auto-mounting from enabled to disabled.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -21,7 +21,7 @@ serviceAccount:
   create: true
   annotations: {}
   name: ""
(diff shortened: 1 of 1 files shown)


nicolasnoble:
This already has one approval. Could use a second look to get it merged. @nv-hwoo @AndyDai-nv

PR #154, opened 2026-02-27
fix: prevent downloads in delete_model

andrewpaprotsky:
**Problem**
- `delete_model` used `repo.get(...)` to find cache paths.
- In `hf_hub`, `repo.get(...)` downloads on cache miss.
- The cleanup loop could call `repo.get(...)` again after deletion, re-creating files.

**Fix**
- Switch delete path resolution to cache-only lookups via `hf_hub::Cache`.
- Track parent directories from files actually deleted in cache.
- Perform empty-directory cleanup only on those tracked local directories.

## Summary by CodeRabbit

* **New Features**
  * Model deletion now requires specifying a cache directory for enhanced control over cached file cleanup.

* **Bug Fixes**
  * Model deletion operations are now cache-aware, eliminating unnecessary network requests during deletion.
  * Improved error handling and automatic cleanup of empty cached directories.

### DIFF
--- a/modelexpress_common/src/providers.rs
+++ b/modelexpress_common/src/providers.rs
@@ -18,7 +18,7 @@ pub trait ModelProviderTrait: Send + Sync {
 
     /// Delete a model from the provider's cache
     /// Returns Ok(()) if the model was successfully deleted or didn't exist
-    async fn delete_model(&self, model_name: &str, cache_dir: Option<PathBuf>) -> Result<()>;
+    async fn delete_model(&self, model_name: &str, cache_dir: PathBuf) -> Result<()>;
 
     /// Get the full path to the latest model snapshot if it exists
     /// Returns the path if found, or an error if not found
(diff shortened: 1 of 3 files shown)


modelexpress_common/src/providers.rs
```diff
@@ -18,7 +18,7 @@ pub trait ModelProviderTrait: Send + Sync {
 
     /// Delete a model from the provider's cache
     /// Returns Ok(()) if the model was successfully deleted or didn't exist
-    async fn delete_model(&self, model_name: &str) -> Result<()>;
+    async fn delete_model(&self, model_name: &str, cache_dir: Option<PathBuf>) -> Result<()>;
```
nicolasnoble: Did you consider making `cache_dir` a required `PathBuf` here, like `get_model_path` does? `download_model` takes `Option` because it handles auto-detection, but `delete_model` operates on already-downloaded models where the path is always known. Keeping it optional reintroduces the env-var fallback that caused the original bug.

PR #157, opened 2026-03-05
feat: Add TransferEngine backend to P2P metadata

ishandhanani:
## Summary
- Extend `WorkerMetadata` in `p2p.proto` with `oneof backend_metadata` supporting both NIXL (bytes) and TransferEngine (session_id string)
- Update all metadata backends (memory, redis, kubernetes, layered) to handle the new oneof variant
- Regenerate Python protobuf stubs for protobuf 5.x compatibility

## Context
This enables SGLang's Mooncake TransferEngine to use ModelExpress for metadata coordination, alongside the existing NIXL support. The seed instance publishes its TransferEngine session ID + tensor descriptors to MX, and target instances query MX to discover the seed and perform RDMA reads.

## Design: typed `oneof` vs opaque bytes

We use a typed `oneof` rather than opaque `bytes backend_metadata` + `string backend_type`. Today MX acts as a metadata store, but the plan is for the MX client to own the transfer pipeline (compute transfer plan, execute RDMA, apply fixups). When MX owns transfers, it needs to interpret the backend metadata to select the right transfer backend -- so the `oneof` gives us compile-time type safety for that future.

## Changes
- `p2p.proto`: `WorkerMetadata.backend_metadata` oneof with `nixl_metadata` (field 2) and `transfer_engine_session_id` (field 10)
- `metadata_backend.rs`: New `BackendMetadataRecord` enum (Nixl/TransferEngine/None) with `from_flat()` helper for deserialization
- `k8s_types.rs`: Added `transfer_engine_session_id: Option<String>` to `WorkerStatus`
- `redis.rs`: Added `transfer_engine_session_id` to `WorkerRecordJson`
- All backend `From` conversions updated
- Python stubs regenerated with `grpcio-tools<=1.66.2`

## Test plan
- [x] 170 Rust tests pass, 0 clippy warnings
- [x] E2E roundtrip: publish metadata with transfer_engine_session_id, get it back with correct oneof
- [x] Backward compat: existing NIXL metadata path unaffected
- [x] SGLang E2E: Qwen3-235B-A22B-FP8, TP4 EP4, 22x speedup (0.74s RDMA vs 16.43s disk)

## Summary by CodeRabbit

* **New Features**
  * Added support for TransferEngine session IDs as an alternative metadata option for worker configuration.

* **Refactor**
  * Restructured worker metadata to support multiple backend types as mutually exclusive options.

### DIFF
--- a/modelexpress_server/src/state.rs
+++ b/modelexpress_server/src/state.rs
@@ -289,7 +289,10 @@ mod tests {
             &record.backend_metadata,
             BackendMetadataRecord::TransferEngine(sid) if sid == "192.168.1.10:12345"
         ));
-        assert_eq!(record.backend_metadata.backend_type_str(), "transfer_engine");
+        assert_eq!(
+            record.backend_metadata.backend_type_str(),
+            "transfer_engine"
+        );
 
         let back: WorkerMetadata = record.into();
         assert_eq!(back.worker_rank, meta.worker_rank);
@@ -299,7 +302,11 @@ mod tests {
     #[test]
     fn test_backend_metadata_from_flat_with_discriminator() {
         // Explicit backend_type takes precedence
-        let te = BackendMetadataRecord::from_flat(Vec::new(), Some("10.0.0.1:5000".into()), Some("transfer_engine"));
+        let te = BackendMetadataRecord::from_flat(
+            Vec::new(),
+            Some("10.0.0.1:5000".into()),
+            Some("transfer_engine"),
+        );
         assert!(matches!(te, BackendMetadataRecord::TransferEngine(ref s) if s == "10.0.0.1:5000"));
 
         let nixl = BackendMetadataRecord::from_flat(vec![1, 2, 3], None, Some("nixl"));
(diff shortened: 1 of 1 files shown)


modelexpress_server/src/metadata_backend/redis.rs
```diff
@@ -118,15 +118,24 @@ impl From<TensorRecordJson> for TensorRecord {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 struct WorkerRecordJson {
     pub worker_rank: u32,
+    #[serde(default)]
     pub nixl_metadata: Vec<u8>,
+    #[serde(default)]
+    pub transfer_engine_session_id: Option<String>,
     pub tensors: Vec<TensorRecordJson>,
 }
```
KavinKrishnan, line 130: Likely need to include `backend_type` here now

modelexpress_server/src/metadata_backend/kubernetes.rs
KavinKrishnan, line 1: Similar to redis, need to ensure there is a backend_type field here for the CRD and likely need to update schema in `examples/p2p_transfer_k8s/deploy/persistence/crd-modelmetadata.yaml`

KavinKrishnan:
Look pretty good overall

Some questions (maybe can be disregarded if this PR is just to setup subsequent followup for SGL support):

1.  How are you handling configuration of which backend to use (NIXL or TE)?  IIUC we are always using NIXL currently
2.  How are we ensuring target node and source are using the same backend (and respective routing logic)?

Some recommendations:

1. Env variables for backend configuration
2. Maybe have a dedicated TransferEngineManager like [NixlTransferManager](https://github.com/ai-dynamo/modelexpress/blob/3376ee5ca73347e7094faca1c597881cce69ab7a/modelexpress_client/python/modelexpress/nixl_transfer.py#L42)

PR #160, opened 2026-03-09
fix: client wheel with a README and valid SPDX identifier

saturley-hall:
The current pyproject.toml misses a couple things:
* It should use a SPDX identifier for the license rather than a Classifier as that is EoL
* The referenced readme file does not exist. I used Claude to generate a candidate one that will show up on PyPi when the wheel is hosted

Since this is a published artifact it will need to be cherry-picked into the `release/0.3.0` branch.

## Summary by CodeRabbit

* **Documentation**
  * Added detailed Python ModelExpress client documentation covering installation methods, system prerequisites, quick-start examples for source and target vLLM instances, environment variable reference, programmatic usage examples, and performance tuning guidance.

* **Chores**
  * Updated project metadata to support Python 3.12.

### DIFF
--- a/modelexpress_client/python/README.md
+++ b/modelexpress_client/python/README.md
@@ -102,7 +102,7 @@ register_modelexpress_loaders()
 | Module | Description |
 |--------|-------------|
 | `modelexpress.client` | `MxClient` -- gRPC client for the ModelExpress server |
-| `modelexpress.vllm_loader` | `MxSourceModelLoader`, `MxTargetModelLoader` -- vLLM integration |
+| `modelexpress.vllm_loader` | `MxModelLoader` -- vLLM integration |
 | `modelexpress.nixl_transfer` | `NixlTransferManager` -- NIXL agent lifecycle and RDMA transfers |
 | `modelexpress.types` | `TensorDescriptor`, `WorkerMetadata` -- core data types |
 | `modelexpress.vllm_worker` | vLLM worker extensions |


modelexpress_client/python/README.md
```diff
@@ -0,0 +1,132 @@
...
+### Source Instance (loads weights from disk, serves via RDMA)
+
+```bash
+export MODEL_EXPRESS_URL="modelexpress-server:8001"
+export MODEL_NAME="deepseek-ai/DeepSeek-V3"
+export MX_EXPECTED_WORKERS=8
+
+vllm serve deepseek-ai/DeepSeek-V3 \
+    --load-format mx-source \
+    --tensor-parallel-size 8
```
AndyDai-nv: I think we need to specify `--worker-cls modelexpress.vllm_worker.ModelExpressWorker`? Correct me if I am wrong here. @nv-hwoo @KavinKrishnan

modelexpress_client/python/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+Python client for ModelExpress -- high-performance GPU-to-GPU model weight transfers using NVIDIA NIXL over RDMA/InfiniBand.
+
+Instead of each vLLM instance loading model weights from storage, one "source" instance loads the model and transfers weights directly to "target" instances via GPUDirect RDMA, bypassing the CPU entirely.
+
+## Installation
+
+```bash
+# From source
+pip install -e .
```
nv-hwoo, line 14: Should we also add pip install from PyPI?
```suggestion
# From PyPI
pip install modelexpress

# From source
pip install -e .
```

modelexpress_client/python/README.md
nv-hwoo: Ah good catch, yes, you are correct :)

PR #164, opened 2026-03-13
test: add multi-node k8s test script for cross-node gRPC transfers

nicolasnoble:
## Summary
- Adds Rust integration tests (`workspace-tests/tests/k8s_multinode_tests.rs`) using the `kube` crate for multi-node k8s testing, all `#[ignore]` by default
- 5 tests: cross-node gRPC transfer with default/64KB/4MB chunk sizes, transfer integrity via md5 content hash comparison, multi-replica independent cache verification
- Minimal shell wrapper (`test_multinode_k8s.sh`) that builds the Docker image, distributes via `microk8s images import`, and runs `cargo test`
- Adds `kube` (ws feature), `k8s-openapi`, and `anyhow` to workspace-tests dependencies

## Test plan
- [x] All 5 tests verified on 3-node microk8s cluster
- [x] `cargo clippy` passes with no warnings
- [x] `pre-commit run --all-files` passes
- [ ] CI passes (tests are `#[ignore]`, won't run in CI)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive multi-node Kubernetes integration test suite with automated test runner and orchestration capabilities.

* **Chores**
  * Added testing-related dependencies to workspace configuration.

### DIFF
--- a/workspace-tests/tests/k8s_multinode_tests.rs
+++ b/workspace-tests/tests/k8s_multinode_tests.rs
@@ -47,10 +47,31 @@ async fn select_nodes(client: &Client) -> Result<(String, String)> {
         .items
         .iter()
         .filter(|n| {
-            !n.spec
+            // Skip nodes marked unschedulable
+            if n.spec
                 .as_ref()
                 .and_then(|s| s.unschedulable)
                 .unwrap_or(false)
+            {
+                return false;
+            }
+            // Skip nodes with NoSchedule taints
+            if n.spec
+                .as_ref()
+                .and_then(|s| s.taints.as_ref())
+                .is_some_and(|taints| taints.iter().any(|t| t.effect == "NoSchedule"))
+            {
+                return false;
+            }
+            // Skip nodes that aren't Ready
+            n.status
+                .as_ref()
+                .and_then(|s| s.conditions.as_ref())
+                .is_some_and(|conds| {
+                    conds
+                        .iter()
+                        .any(|c| c.type_ == "Ready" && c.status == "True")
+                })
         })
         .filter_map(|n| n.metadata.name.clone())
         .collect();
@@ -74,9 +95,11 @@ async fn create_namespace(client: &Client, name: &str) -> Result<NamespaceGuard>
         "metadata": { "name": name }
     }))?;
 
-    // Delete if leftover from a previous failed run
+    // Delete if leftover from a previous failed run, wait until fully gone
     let _ = ns_api.delete(name, &DeleteParams::default()).await;
-    tokio::time::sleep(Duration::from_secs(2)).await;
+    while ns_api.get(name).await.is_ok() {
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    }
 
     ns_api.create(&PostParams::default(), &ns).await?;
 
@@ -383,7 +406,7 @@ async fn submit_checksum_job(
         if [ $? -ne 0 ]; then echo "RESULT=FAILED"; exit 1; fi
 
         echo "CHECKSUMS_START"
-        find /cache -type f -exec md5sum {{}} \; | awk '{{print $1}}' | sort -u
+        find /cache -type f ! -name '*.lock' ! -path '*/refs/*' -exec md5sum {{}} \; | awk '{{print $1}}' | sort -u
         echo "CHECKSUMS_END"
         echo "RESULT=SUCCESS"
         "#
@@ -490,7 +513,7 @@ async fn cross_node_transfer_default_chunks() -> Result<()> {
     let svc = deploy_server(&client, ns, "mx-server", &server_node).await?;
     submit_client_job(&client, ns, "dl-default", &client_node, &svc, "").await?;
 
-    let success = timeout(TIMEOUT, wait_for_job(&client, ns, "dl-default")).await??;
+    let success = wait_for_job(&client, ns, "dl-default").await?;
     let logs = get_job_logs(&client, ns, "dl-default").await?;
     let (files, bytes) = parse_job_results(&logs);
 
@@ -521,7 +544,7 @@ async fn cross_node_transfer_small_chunks() -> Result<()> {
     )
     .await?;
 
-    let success = timeout(TIMEOUT, wait_for_job(&client, ns, "dl-small")).await??;
+    let success = wait_for_job(&client, ns, "dl-small").await?;
     let logs = get_job_logs(&client, ns, "dl-small").await?;
     let (files, bytes) = parse_job_results(&logs);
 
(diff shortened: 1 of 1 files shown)


workspace-tests/Cargo.toml
```diff
@@ -22,6 +22,9 @@ serde_json = { workspace = true }
 tempfile = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
+kube = { workspace = true, features = ["ws"] }
+k8s-openapi = { workspace = true }
+anyhow = { workspace = true }
```
nv-hwoo, line 27: I think we should move these to under dev dependencies as they are only used for testing.

workspace-tests/tests/k8s_multinode_tests.rs
```diff
@@ -0,0 +1,642 @@
...
+        r#"
+        /app/modelexpress-cli \
+            --no-shared-storage \
+            --endpoint "http://{service}:8001" \
+            model download "{TEST_MODEL}"
+
+        if [ $? -ne 0 ]; then echo "RESULT=FAILED"; exit 1; fi
+
+        echo "CHECKSUMS_START"
+        find /cache -type f -exec md5sum {{}} \; | awk '{{print $1}}' | sort -u
```
nv-hwoo: Is there a reason why client does not filter `.lock` and `refs/` while the server checksum does? Should we add the same filter on the client side as well?

workspace-tests/tests/k8s_multinode_tests.rs
nv-hwoo, line 1: Thanks @nicolasnoble for adding these. These are great value adds since we need more k8s testings. It seems like this test is to be ran manually by devs locally (correct me if I'm wrong). Do you think we can also make these  to be triggered by our CI and make it run on the company CI clusters? (unless that requires some support from Ops team)

workspace-tests/tests/k8s_multinode_tests.rs
```diff
@@ -0,0 +1,642 @@
...
+async fn cross_node_transfer_default_chunks() -> Result<()> {
+    let client = Client::try_default().await?;
+    let (server_node, client_node) = select_nodes(&client).await?;
+    let ns = "mx-test-default";
+    let _guard = create_namespace(&client, ns).await?;
+
+    let svc = deploy_server(&client, ns, "mx-server", &server_node).await?;
+    submit_client_job(&client, ns, "dl-default", &client_node, &svc, "").await?;
+
+    let success = timeout(TIMEOUT, wait_for_job(&client, ns, "dl-default")).await??;
```
nv-hwoo: Do we need the outer timeouts when `wait_for_job()` already has a timeout within the function?

workspace-tests/tests/k8s_multinode_tests.rs
```diff
@@ -0,0 +1,642 @@
...
+        "kind": "Namespace",
+        "metadata": { "name": name }
+    }))?;
+
+    // Delete if leftover from a previous failed run
+    let _ = ns_api.delete(name, &DeleteParams::default()).await;
+    tokio::time::sleep(Duration::from_secs(2)).await;
+
+    ns_api.create(&PostParams::default(), &ns).await?;
+
```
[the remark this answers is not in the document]
nv-hwoo: @nicolasnoble could you address this comment? Claude also pointed the same issue and suggested polling until the namespace is gone (404 on GET).

workspace-tests/tests/k8s_multinode_tests.rs
```diff
@@ -0,0 +1,642 @@
...
+const IMAGE: &str = "modelexpress:multinode-test";
+const TEST_MODEL: &str = "hf-internal-testing/tiny-random-gpt2";
+const TIMEOUT: Duration = Duration::from_secs(300);
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/// Pick two different schedulable nodes from the cluster.
+async fn select_nodes(client: &Client) -> Result<(String, String)> {
```
nv-hwoo, line 42: What happens when there are no nodes available? Does it wait?

workspace-tests/Cargo.toml
nicolasnoble: `workspace-tests` is a `publish = false` crate that exists solely to hold tests and benchmarks - so the distinction between `[dependencies]` and `[dev-dependencies]` has no practical effect here. Both sections are available to integration tests regardless, and the crate will never be published. Keeping them where they are for simplicity.

workspace-tests/tests/k8s_multinode_tests.rs
nicolasnoble: Good catch - added the same `.lock` and `refs/` filters to the client-side find command so both sides match.

workspace-tests/tests/k8s_multinode_tests.rs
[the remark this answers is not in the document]
nicolasnoble: Replaced the `sleep(2)` with polling - now loops on `ns_api.get()` until it returns a 404 before creating.

workspace-tests/tests/k8s_multinode_tests.rs
nicolasnoble: Removed the outer timeout - `wait_for_job` already wraps itself in `timeout(TIMEOUT, ...)`.

workspace-tests/tests/k8s_multinode_tests.rs
nicolasnoble: It bails immediately with an error message. Also added taint and readiness filtering so it won't pick nodes that look schedulable but can't actually run pods.

workspace-tests/tests/k8s_multinode_tests.rs
nicolasnoble: Yeah, these are manual-only for now. Running them in CI would need a multi-node cluster with GPU access, which would require ops support to set up. Happy to look into that as a follow-up.

workspace-tests/tests/k8s_multinode_tests.rs
```diff
@@ -0,0 +1,665 @@
...
+    client: &Client,
+    namespace: &str,
+    release: &str,
+    node: &str,
+) -> Result<String> {
+    let svc_name = format!("{release}-svc");
+
+    // Deployment
+    let dep_api: Api<Deployment> = Api::namespaced(client.clone(), namespace);
+    let dep: Deployment = serde_json::from_value(serde_json::json!({
```
zhengluo-nv, line 140: Instead of using rust to deploy manifests, can we exact these manifests into yaml files, so that the rust code can be much cleaner? Then the yaml files for multi-node test can be reused as examples, etc.

nicolasnoble: The manifests here are heavily parameterized - node names, service names, chunk sizes, and test-specific shell scripts are all interpolated at runtime via `serde_json::json\!`. Extracting them to static YAML files would require adding a templating layer on top, which adds complexity without much benefit. These are test harnesses, not deployment examples - the actual K8s deployment examples live in `vllm-redis.yaml` / `vllm-kubernetes.yaml`.

PR #165, opened 2026-03-13
feat: Metadata resiliency phase 1: Simplify metadata schema and backends

zhengluo-nv:
## Summary

- Remove `StatusRecord`; merge source status into `WorkerRecord`; remove `GetStatus`/`PublishReady`/`GetReady` RPCs in favor of `UpdateStatus`/`GetMetadata`
- Remove `expectedWorkers` field from CRD spec, `k8s_types`, and server logic (was stored but never read)
- Fix Kubernetes backend SSA field manager conflicts: switch all patch operations from `PatchParams::apply` + `Patch::Apply` to `PatchParams::default` + `Patch::Merge`
- Fix `p2p_pb2` `NameError` in `vllm_loader.py` by moving import to module scope
- Remove `layered` and `memory` metadata backends; Redis and Kubernetes are the only supported backends
- Update example manifests and docs

## Testing Summary

End-to-end validation of the P2P weight transfer system after the unified mx loader refactor and metadata backend simplification.

**Environment**
- Cluster: Multi-node K8s with GPUDirect RDMA (NDR NICs, `rdma/ib`)
- Model: `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4` (75 GB, 2600 tensors across 2 GPU workers)
- Config: TP=2, `kv-cache-dtype=fp8`, vLLM plugin system (`VLLM_PLUGINS=modelexpress`)
- Images: `linux/amd64` Docker builds, pinned by digest

Did not test on DeepSeek TP=8 due to lack of GPU resources as of today. TP=2 should be the same.

### Kubernetes CRD Backend (`MX_METADATA_BACKEND=kubernetes`)

Deployment: `modelexpress-server-kubernetes.yaml` + `vllm-kubernetes.yaml` + RBAC

- **Cold start (disk load):** Pod 1 loaded weights from disk, registered with NIXL, published `ModelMetadata` CRD and tensor descriptor ConfigMaps. Both workers (rank 0 and rank 1) published with `status=READY` and 1300 tensors each.
- **P2P transfer:** Pod 2 queried the CRD, detected source with 1300 tensors per rank, performed RDMA transfer:
  - Worker 0: 1300 tensors, 37.62 GB in 6.21s → **48.5 Gbps**
  - Worker 1: 1300 tensors, 37.62 GB in 6.38s → **47.2 Gbps**
- **Scale resilience (1→2→3 replicas):** Verified the ConfigMap conflict fix — subsequent pods no longer fail with `409 Conflict` when applying tensor descriptor ConfigMaps. Pod 3 correctly detected an existing source and transferred via RDMA.
- **Inference validation:** Chat completion request sent directly to Pod 2's IP returned a valid response, confirming RDMA-transferred weights are correct.

### Redis Backend (`MX_METADATA_BACKEND=redis`)

Deployment: `modelexpress-server-redis.yaml` (Redis native sidecar) + `vllm-redis.yaml`

- **Redis sidecar:** Runs as a native sidecar `initContainer` (`restartPolicy: Always`), starts before the server container. Pod shows `2/2` ready.
- **Metadata publish:** Pod 1 published both workers to Redis. Verified via `redis-cli`:
  - `worker_rank=0 status=2 tensors=1300`
  - `worker_rank=1 status=2 tensors=1300`
- **P2P transfer:** Pod 2 queried Redis metadata via the MX server, detected the source, and transferred via RDMA:
  - Worker 0: 37.62 GB in 6.21s → **48.5 Gbps**
  - Worker 1: 37.62 GB in 6.38s → **47.2 Gbps**
- **Metadata overwrite:** After Pod 2 completed loading, it published its own tensor addresses to Redis (overwriting Pod 1's addresses). Server logs confirmed: `"Published metadata for model '...': 2 workers total (1300 new tensors)"`.

## Release Notes

* **New Features**
  * Added `UpdateStatus` RPC for simplified worker status coordination (Initializing/Ready/Stale states).

* **Refactor**
  * Removed in-memory metadata backend; now supports only Redis and Kubernetes CRD persistence.
  * Simplified deployment configuration with explicit backend selection.
  * Updated Kubernetes CRD schema with unified per-worker status field replacing previous readiness model.

* **Documentation**
  * Updated architecture documentation to reflect new status-based coordination flow.
  * Enhanced deployment guides for Redis and Kubernetes CRD backends.


modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -213,18 +212,10 @@ def _publish_metadata_and_ready(
...
-            mx_client.publish_ready(
+            mx_client.update_status(
                 model_name=model_name,
                 worker_id=device_id,
-                session_id=mx_client.session_id,
-                metadata_hash=metadata_hash,
-                nixl_ready=True,
-                stability_verified=True,
+                status=p2p_pb2.SOURCE_STATUS_READY,
             )
```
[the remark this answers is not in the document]
zhengluo-nv: I will fix this in the upcoming PRs because we will eventually separate the metadata into per rank.

examples/p2p_transfer_k8s/deploy/persistence/crd-modelmetadata.yaml
```diff
@@ -67,21 +54,19 @@ spec:
...
+                      status:
                         type: string
-                        description: Unique session ID for detecting source restarts
-                      publishedAt:
+                        description: Worker lifecycle status
+                        enum:
+                          - Unknown
+                          - Initializing
+                          - Ready
+                          - Stale
```
nv-hwoo: general question: what is Stale` state and when does worker fall into this state?

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -497,37 +469,26 @@ def _receive_from_peer(
...
+                                    dtype=t.dtype,
                                 )
-                                break
+                                for t in source_worker.tensors
+                            ]
+                            logger.info(
+                                f"[Worker {device_id}] Refreshed metadata: "
+                                f"{len(source_tensors)} tensors"
+                            )
+                            break
```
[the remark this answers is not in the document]
zhengluo-nv: This re-fetch logic will be revisited in the next PR. It's some legacy code needs to be fixed

modelexpress_common/proto/p2p.proto
```diff
@@ -95,55 +106,22 @@ message GetMetadataResponse {
...
-  string session_id = 3;
-
-  // Hash of metadata for validation
-  string metadata_hash = 4;
-
-  // Whether NIXL is registered and ready
-  bool nixl_ready = 5;
-
-  // Whether stability has been verified (warmup complete)
-  bool stability_verified = 6;
```
nv-hwoo, line 122: @KavinKrishnan wasn't this needed to make sure we don't break the warmups on the source side? was this specific to dsv3 for its deepgemm warmup phase?

modelexpress_server/src/metadata_backend/kubernetes.rs
```diff
@@ -481,4 +474,82 @@ impl MetadataBackend for KubernetesBackend {
...
+        let cr_name = sanitize_model_name(model_name);
+        let status_name = WorkerStatus::status_name_from_proto(status);
+        let updated_at_rfc3339 = chrono::DateTime::from_timestamp_millis(updated_at)
+            .map(|dt| dt.to_rfc3339())
+            .unwrap_or_else(|| chrono::Utc::now().to_rfc3339());
+
+        let max_retries: u32 = 5;
+        for attempt in 0..max_retries {
+            let current = api.get(&cr_name).await?;
+            let mut all_workers: Vec<WorkerStatus> =
```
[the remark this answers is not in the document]
zhengluo-nv: The CRD will be changed in the next PR. Not worth to fix it

examples/p2p_transfer_k8s/deploy/persistence/crd-modelmetadata.yaml
```diff
@@ -67,21 +54,19 @@ spec:
...
-                        description: Whether worker has passed stability verification
-                        default: false
-                      sessionId:
+                      status:
                         type: string
-                        description: Unique session ID for detecting source restarts
-                      publishedAt:
+                        description: Worker lifecycle status
+                        enum:
+                          - Unknown
```
nv-hwoo, line 71: Unrelated to this PR but I feel like `Unknown` is a very vague status. Do you think we can change to something more specific in the follow-up PRs?

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -497,37 +469,26 @@ def _receive_from_peer(
...
-                                    )
-                                    for t in source_worker.tensors
-                                ]
-                                logger.info(
-                                    f"[Worker {device_id}] Refreshed metadata: "
-                                    f"{len(source_tensors)} tensors from new session"
+                    # Re-fetch metadata in case source was restarted
+                    response = self._mx_client.get_metadata(model_name)
+                    for w in response.workers:
+                        if w.worker_rank == device_id and len(w.tensors) > 0:
```
nv-hwoo, line 475: should we also check if the source is ready or not, in case it restarted?

modelexpress_common/proto/p2p.proto
KavinKrishnan: Yes, we should probably keep this 

Maybe have it as a configurable workflow in cases like DeepGEMM warmup

zhengluo-nv: Sure, my idea is to merge this state into the status to keep only 1 status field instead of N. Otherwise there are so many combinations of status fields and it's confusing. Maybe we can define Ready as both metadata published and stability verified, and we can also add PENDING_VERIFICATION to represent metadata published but stability not verified

nv-hwoo: @KavinKrishnan btw how are we verifying if the warmup is finished? i think warmup happens after all the weight post processing. we can try to repro this and see if the current PR fails.

modelexpress_client/python/modelexpress/vllm_loader.py
zhengluo-nv: I think the final state will be every source publishing heartbeat to the MX server, and the MX client will only fetch a list of healthy sources. That approach is aligned with Itay and Nico. Do you think we need to add a temporary check here, or just implement heartbeat in the following PRs?

nv-hwoo: Ah I see, yea hearbeat makes sense 👍

modelexpress_common/proto/p2p.proto
zhengluo-nv: I think the stability_verified was never implemented so I removed it

AndyDai-nv (approved):
LGTM, thanks for the fix

examples/p2p_transfer_k8s/deploy/persistence/crd-modelmetadata.yaml
zhengluo-nv: Will Remove in the next PR

PR #156, opened 2026-03-05
feat: add mx gds loader

AndyDai-nv:
Add GDS loading support, example:
```bash
$ vllm serve Qwen/Qwen2.5-7B \
    --load-format mx-gds \
    --worker-cls modelexpress.vllm_worker.ModelExpressWorker

Initialized NIXL agent: mx-gds-0-943c39d1
GDS agent 'mx-gds-0-943c39d1' created on device 0 (max_chunk=1024KB)
GDS manager initialized for device 0
Found sharded model: 4 files, 339 tensors
Loaded model-00004-of-00004.safetensors
Loaded model-00001-of-00004.safetensors
Loaded model-00002-of-00004.safetensors
Loaded model-00003-of-00004.safetensors
GDS load complete: 15.23 GB in 11.51s
GDS weight loading complete
Model loading took 14.27 GiB memory and 13.070087 seconds
```

## Release Notes

* **New Features**
  * Introduced GPUDirect Storage (GDS) loader enabling direct GPU weight loading from safetensors files
  * Added vLLM and SGLang framework integrations for GDS-based model loading

* **Documentation**
  * Added comprehensive GDS loader documentation covering architecture, data flow, and configuration

* **Chores**
  * Expanded protobuf dependency compatibility
  * Enhanced Python environment ignore patterns

PR #166, opened 2026-03-13
feat: add GDS loading functionality in auto-detetction mx loader

AndyDai-nv:
## Summary

Integrate GPUDirect Storage (GDS) loading into the unified `mx` loader as a three-tier auto-detection strategy: RDMA -> GDS -> disk.

Users only need `vllm serve <model> --load-format mx`. When no RDMA source is available, the loader automatically detects GDS hardware (nvidia_fs kernel module + libcufile.so) and loads safetensors directly from NVMe to GPU via NIXL's GDS_MT backend, bypassing CPU memory entirely. Falls back to vLLM's default disk loader if GDS is unavailable.

- Add `gds_transfer.py`: NIXL GDS_MT transfer manager with FILE/VRAM batch registration, `initialize_xfer` API, and proper `deregister_memory` cleanup
- Add `gds_loader.py`: Framework-agnostic safetensors parser with sharded/single-file discovery, prefetch pipeline, and tqdm progress bar
- Modify `vllm_loader.py`: Insert GDS tier in `_load_as_source()`, feed GDS iterator through `model.load_weights()` for correct tensor name mapping (QKV merge, etc.)
- GDS availability detection via nvidia_fs kernel module + libcufile.so (NIXL silently falls back to POSIX, so agent creation alone cannot detect GDS)
- Graceful fallback: MX server unavailable no longer crashes the loader; publish errors are logged as warnings
- Add `huggingface_hub` to pyproject.toml dependencies
- Add unit tests for GDS detection, transfer manager, safetensors parsing, file discovery, and MxModelLoader integration

Addresses review feedback from #156: context manager support, useful_bytes infinite loop guard, device change handling, revision pass-through, huggingface_hub dependency.

```
 $ vllm serve Qwen/Qwen2.5-7B --load-format mx --worker-cls modelexpress.vllm_worker.ModelExpressWorker

...
(EngineCore_DP0 pid=699594) [2026-03-13 22:38:17] INFO gds_transfer.py:133: GDS_MT agent 'mx-gds-0-2cf966ba' created on device 0 (threads=8, max_chunk=128MB)
(EngineCore_DP0 pid=699594) [2026-03-13 22:38:17] INFO gds_loader.py:195: GDS manager initialized for device 0
(EngineCore_DP0 pid=699594) [2026-03-13 22:38:17] INFO gds_loader.py:222: Found sharded model: 4 files, 339 tensors
Loading safetensors via GDS:   0%|                                                                    | 0/4 [00:00<?, ?file/s][2026-03-13 22:38:17] INFO gds_loader.py:338: Loaded model-00004-of-00004.safetensors
Loading safetensors via GDS:  25%|███████████████                                             | 1/4 [00:00<00:00,  3.18file/s][2026-03-13 22:38:17] INFO gds_loader.py:338: Loaded model-00001-of-00004.safetensors
Loading safetensors via GDS:  50%|██████████████████████████████                              | 2/4 [00:00<00:00,  3.03file/s][2026-03-13 22:38:18] INFO gds_loader.py:338: Loaded model-00002-of-00004.safetensors
Loading safetensors via GDS:  75%|█████████████████████████████████████████████               | 3/4 [00:00<00:00,  3.01file/s][2026-03-13 22:38:18] INFO gds_loader.py:338: Loaded model-00003-of-00004.safetensors
Loading safetensors via GDS: 100%|████████████████████████████████████████████████████████████| 4/4 [00:01<00:00,  3.00file/s]
(EngineCore_DP0 pid=699594) [2026-03-13 22:38:18] INFO gds_loader.py:144: GDS load complete in 2.06s
```

## Test Plan

- [x] GDS loading: Qwen2.5-7B on GDS-capable node, 4 shards loaded in 2.06s, 15.23 GB
- [x] Correctness: tensor checksums match disk loading (e.g. `fb942744` for qkv_proj.weight)
- [x] Inference: curl chat completions returns valid response after GDS load
- [x] Fallback to disk: works when GDS hardware not present
- [x] Fallback on MX server unavailable: loader continues without P2P registration
- [ ] RDMA path: not tested in this PR (no multi-node setup available), existing behavior unchanged
- [ ] FP8 model (DeepSeek-V3): not tested due to GPU resource constraints

## Release Notes

* **New Features**
  * Added GPU Direct Storage (GDS) support for accelerated model weight loading with automatic fallback to alternative methods if unavailable.
  * Implemented intelligent three-tier loading strategy that prioritizes the fastest available transfer method.

* **Tests**
  * Added comprehensive test coverage for GDS functionality and model loading scenarios.


modelexpress_client/python/modelexpress/vllm_loader.py
zhengluo-nv: nit: Shall we rename this to `self._default_loader`?

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,322 @@
...
+from pathlib import Path
+from typing import Iterator
+
+import torch
+
+from .gds_transfer import GdsTransferManager, is_gds_available
+
+logger = logging.getLogger("modelexpress.gds_loader")
+
+SAFETENSORS_DTYPE_MAP: dict[str, torch.dtype] = {
```
zhengluo-nv, line 35: Is this the exhaust list of safetensors dtypes? Where can we find the list to make sure everything is supported?

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,322 @@
...
+            if file_tensors:
+                file_jobs.append((file_path, file_tensors))
+
+        if not file_jobs:
+            return
+
+        # Prefetch pipeline: load file[i+1] while yielding file[i]
+        from tqdm import tqdm
+        total_files = len(file_jobs)
+        pbar = tqdm(
```
zhengluo-nv: In vllm default loader, there is a self.load_config.use_tqdm_on_load flag which controls whether to show the progress bar. I think in GDS loader we should also use this config to control tqdm
https://github.com/vllm-project/vllm/blob/5ce2d10e4a6954802f482add02b04e23e737ad27/vllm/model_executor/model_loader/weight_utils.py#L703

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,322 @@
...
+                if name in header_info
+            }
+            if file_tensors:
+                file_jobs.append((file_path, file_tensors))
+
+        if not file_jobs:
+            return
+
+        # Prefetch pipeline: load file[i+1] while yielding file[i]
+        from tqdm import tqdm
```
nv-hwoo: are we installing `tqdm` into our package? (making sure we have all the dependencies)

modelexpress_client/python/modelexpress/gds_transfer.py
```diff
@@ -0,0 +1,236 @@
...
+            self._agent.deregister_memory(vram_descs)
+            raise RuntimeError("GDS batch transfer failed")
+
+        # Phase 4: Wait for completion
+        timeout = float(os.environ.get("MX_GDS_TIMEOUT", "120"))
+        t0 = time.perf_counter()
+        spins = 0
+        while True:
+            state = self._agent.check_xfer_state(handle)
+            if state == "DONE":
```
nv-hwoo, line 210: we should create str enums for these states

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -222,10 +223,7 @@ def _publish_metadata_and_ready(
                 f"[Worker {device_id}] Failed to publish metadata to MX server"
             )
 
-    except Exception as e:
-        import traceback
-        logger.error(f"[Worker {device_id}] EXCEPTION publishing metadata: {e}")
-        logger.error(f"[Worker {device_id}] Traceback: {traceback.format_exc()}")
```
nv-hwoo, line 331: why were these removed?

modelexpress_client/python/tests/test_gds_loader.py
```diff
@@ -0,0 +1,267 @@
...
+            return MxModelLoader(load_config)
+
+    @patch("modelexpress.vllm_loader.is_gds_available", return_value=True)
+    @patch("modelexpress.vllm_loader.MxGdsLoader")
+    def test_gds_success_skips_disk(self, mock_gds_cls, _mock_avail):
+        loader = self._make_loader()
+
+        mock_gds = MagicMock()
+        mock_gds.load_weights_into_model.return_value = 10
+        mock_gds_cls.return_value = mock_gds
```
nv-hwoo, line 222: What are we mocking here? Does GDS loader have a method called `load_weights_into_model`?

modelexpress_client/python/modelexpress/gds_transfer.py
```diff
@@ -0,0 +1,236 @@
...
+
+NIXL_AVAILABLE = False
+NixlAgent = None
+NixlAgentConfig = None
+try:
+    from nixl._api import nixl_agent as NixlAgent
+    from nixl._api import nixl_agent_config as NixlAgentConfig
+    NIXL_AVAILABLE = True
+except ImportError:
+    pass
```
nv-hwoo, line 35: Duplicate logic. This already exists in the nixl_transfer.py

modelexpress_client/python/modelexpress/gds_transfer.py
```diff
@@ -0,0 +1,236 @@
...
+        with GdsTransferManager(agent_name="mx-gds-0") as gds:
+            gds.batch_load_file(fd, file_size, tensor_list, device)
+    """
+
+    def __init__(self, agent_name: str):
+        self._agent_name = agent_name
+        self._device_id: int | None = None
+        self._agent: Any = None
+        override = os.environ.get("MX_GDS_MAX_CHUNK_KB")
+        self._max_chunk_size = int(override) * 1024 if override else _DEFAULT_MAX_CHUNK
```
nv-hwoo, line 104: nit:
```suggestion
        self._max_chunk_size = os.environ.get("MX_GDS_MAX_CHUNK_KB", _DEFAULT_MAX_CHUNK)
```

modelexpress_client/python/modelexpress/gds_transfer.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __enter__(self) -> GdsTransferManager:
+        self.initialize()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.shutdown()
+        return None
```
nv-hwoo, line 112: I don't think we're using this class as a context manager?
```suggestion
```

modelexpress_client/python/modelexpress/gds_transfer.py
```diff
@@ -0,0 +1,236 @@
...
+        t0 = time.perf_counter()
+        spins = 0
+        while True:
+            state = self._agent.check_xfer_state(handle)
+            if state == "DONE":
+                break
+            if state == "ERR":
+                self._agent.release_xfer_handle(handle)
+                self._agent.deregister_memory(file_descs)
+                self._agent.deregister_memory(vram_descs)
```
nv-hwoo: seems repetitive, we should pull it out as a new method (e.g. `free_nixl_memory`)

modelexpress_client/python/modelexpress/gds_loader.py
AndyDai-nv: Yes, it should cover all common dtype, special quantization data type may not be included but it is hard to cover them all atm, we can add support when there is a need.

modelexpress_client/python/modelexpress/gds_transfer.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # Phase 3: Submit all at once
+        handle = self._agent.initialize_xfer(
+            "READ", vram_descs.trim(), file_descs.trim(), self._agent.name
+        )
+
+        state = self._agent.transfer(handle)
+        if state == "ERR":
+            self._agent.deregister_memory(file_descs)
+            self._agent.deregister_memory(vram_descs)
```
nv-hwoo: missing release handle
```suggestion
            self._agent.release_xfer_handle(handle)
            self._agent.deregister_memory(file_descs)
            self._agent.deregister_memory(vram_descs)
```

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -365,9 +364,9 @@ def _detect_source(self, model_name: str, device_id: int):
             )
             return None
 
-        except Exception as e:
+        except Exception:
             logger.warning(
-                f"[Worker {device_id}] Error detecting source, falling back to disk: {e}"
+                f"[Worker {device_id}] MX server unavailable, skipping RDMA source detection"
```
nv-hwoo: I'm not sure if the new warning message describes the possible exception. It could be due to MX server not being available, but it could potentially be something else too. I think we should keep it as generic like before (e.g. "Error detecting source") and print out the error message.

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,322 @@
...
+                pbar.update(1)
+
+                if i + 1 < total_files:
+                    pending = pool.submit(
+                        self._load_file_tensors, *file_jobs[i + 1]
+                    )
+
+                for name, tensor in loaded.items():
+                    yield name, tensor
+
```
nv-hwoo, line 149: nit:
```suggestion
        try:
            idx = 0
            while idx < total_files:
                pending = pool.submit(self._load_file_tensors, *file_jobs[idx])
                loaded = pending.result()
                pbar.update(1)

                for name, tensor in loaded.items():
                    yield name, tensor
    
```

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,322 @@
...
+        model_path: str, revision: str | None = None
+    ) -> str:
+        """Resolve model_path to a local directory."""
+        p = Path(model_path)
+        if p.is_dir():
+            return str(p.resolve())
+
+        from huggingface_hub import snapshot_download
+        local_dir = snapshot_download(model_path, revision=revision)
+        logger.info("Resolved HF model '%s' -> %s", model_path, local_dir)
```
nv-hwoo, line 171: Should we rely on MX to resolve this instead of directly using hf? (maybe in the follow up PRs but wondering if that's how we should be doing or not).

modelexpress_client/python/tests/test_gds_loader.py
```diff
@@ -0,0 +1,267 @@
...
+        with patch.object(loader, "_register_and_publish"), \
+             patch("modelexpress.vllm_loader.process_weights_after_loading"):
+            loader._load_as_source(
+                model, model_config,
+                target_device=torch.device("cpu"),
+                device_id=0,
+                model_name="test-model",
+            )
+
+        loader._disk_loader.load_weights.assert_called_once_with(model, model_config)
```
nicolasnoble: Did you mean `_default_loader` here? `MxModelLoader` doesn't have a `_disk_loader` attribute - this will raise `AttributeError` when the test runs.

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,328 @@
...
+                pbar.close()
+            pool.shutdown(wait=True)
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def _resolve_model_path(
+        model_path: str, revision: str | None = None
```
nicolasnoble, line 162: `revision` is accepted here but `load_iter` never passes it through. If the user specifies a revision via vLLM config, this will silently download the latest `main` instead. Should `load_iter` accept and forward a `revision` parameter?

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -0,0 +1,328 @@
...
+    ) -> str:
+        """Resolve model_path to a local directory."""
+        p = Path(model_path)
+        if p.is_dir():
+            return str(p.resolve())
+
+        from huggingface_hub import snapshot_download
+        local_dir = snapshot_download(model_path, revision=revision)
```
nicolasnoble, line 170: Any reason not to pass `allow_patterns=["*.safetensors", "*.safetensors.index.json"]` to `snapshot_download`? Without it this pulls the entire snapshot including bin files, tokenizer data, etc. - unnecessary since the GDS loader only reads safetensors.

modelexpress_client/python/modelexpress/gds_loader.py
AndyDai-nv: tqdm is a transitive dependency of both vllm and huggingface_hub, which are already in our dependency list. No need to add it explicitly.

modelexpress_client/python/modelexpress/gds_transfer.py
AndyDai-nv: These state strings ("DONE", "ERR") are return values from the NIXL API, not our constants. We match what NIXL returns. Can wrap them if the API changes, but they're stable for now.

modelexpress_client/python/modelexpress/vllm_loader.py
AndyDai-nv: The duplicate error logging in _publish_metadata_and_ready was removed because the caller (_register_and_publish) now catches the exception and logs a single warning. Previously the same error was logged 3 times.

modelexpress_client/python/modelexpress/gds_transfer.py
AndyDai-nv: The current code converts KB to bytes (int(override) * 1024), which wouldn't work with the suggested simplification since _DEFAULT_MAX_CHUNK is already in bytes. Keeping as-is for clarity.

modelexpress_client/python/modelexpress/gds_transfer.py
AndyDai-nv: Available for standalone usage outside vLLM.

modelexpress_client/python/modelexpress/gds_loader.py
AndyDai-nv: The current pattern overlaps I/O with tensor processing — loads file[i+1] while vLLM consumes file[i]. A sequential loop would lose this overlap. Can simplify in a follow-up if profiling shows the overlap doesn't help.

modelexpress_client/python/modelexpress/gds_loader.py
AndyDai-nv: The model directory is typically pre-downloaded or mounted locally, in which case snapshot_download is never called. When it is called for a HuggingFace model ID, the full snapshot is needed since vLLM requires config.json, tokenizer files, etc. for model initialization — not just safetensors.

modelexpress_client/python/modelexpress/gds_loader.py
AndyDai-nv: Good idea. For now GDS loading is a local-only path (no MX server needed), but we can route through MX for model resolution in a follow-up PR.

PR #168, opened 2026-03-17
fix: wire provider delete_model into cache eviction service

nicolasnoble:
## Summary

- The LRU eviction service was only removing models from the SQLite database without deleting the actual files from disk, making cache eviction effectively a no-op for disk space
- Now that #154 landed with a correct `delete_model` implementation, this wires it into the eviction path
- Files are deleted first, then the DB record is removed on success. If filesystem deletion fails, the DB record is kept so the next eviction cycle can retry

## Test plan

- [x] Existing cache eviction tests pass
- [x] `cargo clippy` clean
- [x] Pre-commit hooks pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cache eviction process to properly remove cached models from disk rather than logging intended deletions. The eviction service now performs actual file cleanup when clearing the cache.

### DIFF
--- a/modelexpress_server/src/cache.rs
+++ b/modelexpress_server/src/cache.rs
@@ -464,9 +464,13 @@ mod tests {
         (db, temp_dir)
     }
 
-    fn create_test_service(db: ModelDatabase, config: CacheEvictionConfig) -> CacheEvictionService {
+    fn create_test_service(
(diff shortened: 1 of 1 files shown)


AndyDai-nv (approved):
LGTM

PR #169, opened 2026-03-17
feat: register post-processed tensors with NIXL using full module tree traversal

nv-hwoo:
## Problem

Currently we register the original weight tensors with NIXL _before_ the `process_weights_after_loading()` call in vLLM loader. This is a problem as NIXL registration holds references to the origin weight tensors (pinned), and any modifications to the model tensors by the kernel backends (e.g. FlashInfer) during `process_weights_after_loading()` cannot free the original weights after the new tensors are created/allocated. 

This ends up wasting a lot of extra HBM memories, causing CUDA OOM with large models like Kimi k2.5 or DSV3. For example, when loading DSV3 TEP8 on 8x H200 GPUs, the current code causes extra 50GB per worker, leaving very little or no room for kv caches. Also, when loading Kimi k2.5 TEP4 on 4x B200 GPUs (which should fit as each GPU has ~178GB HBM), we run into CUDA OOM during the `process_weights_after_loading()` as there are no more free GPU memory due to pinned original weight tensors.

## Solution

Move the NIXL registration to _after_ `process_weights_after_loading()` and register the final model tensors after all the processings are done. This eliminates the pinned memory issue and CUDA OOM issues described above.

## Code Changes

  - Move NIXL tensor registration to after `process_weights_after_loading()` so that all final tensors (parameters, buffers, and bare tensor attributes like FP8 scales) are captured
  - Replace `named_parameters()` with `_iter_module_tensors()` which walks the full PyTorch module tree to discover parameters, buffers, and tensor attributes (e.g. weight_scale, _k_scale)
  - Skip non-contiguous tensors (e.g. transposed views like W_UK_T) since they are views over contiguous tensors already registered
  - Target path now runs process_weights_after_loading() on dummy data to establish the final tensor layout, then receives fully-processed weights via RDMA — no post-processing needed after transfer

## Summary by CodeRabbit

* **New Features**
  * Expanded tensor discovery to identify parameters, buffers, and custom tensor attributes
  * Refined weight processing sequence for improved tensor registration and handling

* **Documentation**
  * Architecture guide updated to reflect enhanced tensor discovery and processing workflow


modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -111,13 +112,102 @@ def _get_worker_rank(device: torch.device) -> int:
...
 
-def _collect_cuda_tensors(model: nn.Module) -> dict[str, torch.Tensor]:
-    """Collect all CUDA parameter tensors from a model."""
-    return {
-        name: param.data
-        for name, param in model.named_parameters()
-        if param.is_cuda
-    }
+
+def _iter_module_tensors(
```
nicolasnoble, line 116: `_collect_cuda_tensors` no longer exists on this branch - `test_vllm_loader.py` still imports it in three places (`TestCollectCudaTensors`), so those tests will crash with `ImportError`. They need to be updated to exercise `_collect_module_tensors` instead, ideally covering the new discovery categories (buffers, bare tensor attributes, non-contiguous skipping).

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -111,13 +112,102 @@ def _get_worker_rank(device: torch.device) -> int:
...
+    skipped = 0
+    for name, tensor, _tensor_type in _iter_module_tensors(model):
+        t = tensor.data if hasattr(tensor, "data") else tensor
+
+        if not t.is_contiguous():
+            logger.debug(f"Skipping non-contiguous tensor '{name}' (view of another tensor)")
+            skipped += 1
+            continue
+
+        tensors[name] = t
```
nicolasnoble, line 217: Is there a risk of the same tensor being discovered under multiple names here? For example, if a module has a property that returns an existing parameter, `dir()` would find it under a different name while the `_parameters` scan already picked it up. Both would end up in the dict with the same `data_ptr()` but different keys - registered and transferred twice. A dedup pass by `data_ptr()` would be cheap insurance.

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -540,20 +622,20 @@ def _register_and_publish(
             logger.warning(f"[Worker {device_id}] NIXL not available, skipping registration")
             return
 
-        raw_tensors = _collect_cuda_tensors(model)
-        self._raw_tensors = raw_tensors
-        _log_tensor_summary(raw_tensors, device_id, "Registering tensors")
+        tensors = _collect_module_tensors(model)
```
nicolasnoble: This is the second `_collect_module_tensors` call on the target path - `_receive_from_peer` already walked the full module tree. On a 671B model with deeply nested submodules, the `dir()` scan isn't free. Would it be worth caching the result from the first call?

nicolasnoble:
**`_log_tensor_summary` (line 231):** Now that registration happens after `process_weights_after_loading()`, `weight_scale_inv` has been renamed to `weight_scale` - the `scale_inv` count will always be 0. Should this match `weight_scale` instead?

modelexpress_client/python/modelexpress/vllm_loader.py
nv-hwoo: Good catch. Updated the tests.

modelexpress_client/python/modelexpress/vllm_loader.py
nv-hwoo: Now checks and skips tensors with duplicate data_ptr.

modelexpress_client/python/modelexpress/vllm_loader.py
nv-hwoo: Yea good point. Refactoring the registeration stage part now avoids this issue.

PR #171, opened 2026-03-19
ci: enable Python tests in CI and fix stale test suite

nicolasnoble:
Enable Python tests in CI and fix stale test suite. The Python tests were never run in CI, and had silently broken - 3 out of 6 test classes were failing due to API drift and a protobuf version mismatch that made the package unimportable.

- Add python-test job to CI pipeline using uv and pytest
- Regenerate p2p_pb2.py with protobuf 5.x (was 6.x, incompatible with the <6.0.0 pin in pyproject.toml)
- Fix relative import in p2p_pb2_grpc.py (absolute import broke when used as a package)
- Add protobuf version compatibility test to catch gencode/runtime mismatches early
- Update _detect_source tests to match current API (get_metadata + status field, no more get_ready)
- Update _publish_metadata_and_ready tests (update_status replaces publish_ready, failure now raises)
- Add missing copyright headers to regenerated protobuf files

## Summary by CodeRabbit

* **Chores**
  * Added Python test automation to the CI/CD pipeline.

* **Tests**
  * Added protobuf compatibility validation.
  * Updated loader tests to align with current API behavior.

* **Style**
  * Fixed formatting inconsistencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,7 +309,7 @@ jobs:
   release:
     name: Create Release
     runs-on: ubuntu-latest
-    needs: [test, integration-test, security-audit, docker-build, k8s-grpc-test]
+    needs: [test, python-test, integration-test, security-audit, docker-build, k8s-grpc-test]
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
 
     steps:
@@ -352,7 +352,7 @@ jobs:
   notify:
     name: Notify on Failure
     runs-on: ubuntu-latest
-    needs: [test, integration-test, security-audit, docker-build, k8s-grpc-test]
+    needs: [test, python-test, integration-test, security-audit, docker-build, k8s-grpc-test]
     if: failure()
 
     steps:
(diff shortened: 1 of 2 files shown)


modelexpress_client/python/modelexpress/p2p_pb2.py
```diff
@@ -5,7 +5,7 @@
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # NO CHECKED-IN PROTOBUF GENCODE
 # source: p2p.proto
-# Protobuf Python Version: 6.31.1
```
zhengluo-nv, line 8: Is there a reason to pin to the old protobuf version?

nicolasnoble: Yes, so: protobuf and grpc are never compatible from one version to another, and upgrading it always needs to be done thoughtfully. It was upgraded erroneously in a previous PR without doing it fully, and this causes builds errors in some environments. A full upgrade PR will be coming up next, I'm only restoring the status quo from before.

zhengluo-nv (approved):

PR #172, opened 2026-03-19
refactor: make cache operations and streamed downloads provider-aware

andrewpaprotsky:
**Why**:
Client, server, and cache code were each rebuilding Hugging Face cache paths independently. That duplication made streamed downloads brittle when shared storage is disabled and made provider-specific cache operations harder to expose cleanly in the client and CLI.

**What**:
- extract shared provider cache plumbing for listing, clearing, and resolving model paths
- move Hugging Face snapshot layout handling behind that provider cache layer
- thread provider selection through client and CLI cache operations so clear/list actions target the correct cache entry and report the provider explicitly
- make server and client streaming use the same path resolver, and send the Hugging Face commit hash on the first chunk so the client reconstructs the exact snapshot directory

## Summary by CodeRabbit

* **New Features**
  * Added `--provider` / `-p` option to the `model clear` command (defaults to HuggingFace).
  * Model listings and details now display provider information.
  * Model information now includes provider data.

* **Improvements**
  * Enhanced path validation for streaming downloads.
  * Improved cache management with multi-provider support.
  * Better model listing and status displays with provider prefixes.

### DIFF
--- a/modelexpress_server/src/services.rs
+++ b/modelexpress_server/src/services.rs
@@ -348,10 +348,7 @@ impl ModelService for ModelServiceImpl {
 
                     let is_last_chunk = offset.saturating_add(bytes_read as u64) >= *total_size;
 
-                    let first_chunk = is_first_chunk;
-                    if first_chunk {
-                        is_first_chunk = false;
-                    }
+                    let first_chunk = std::mem::replace(&mut is_first_chunk, false);
 
                     let chunk = FileChunk {
                         relative_path: relative_path.to_string_lossy().to_string(),


modelexpress_client/src/lib.rs
```diff
@@ -119,29 +178,33 @@ impl Client {
...
-        let cache_dir = self.get_cache_dir();
-        let model_path = HuggingFaceProvider
+    pub async fn get_model_path(
+        &self,
+        model_name: &str,
+        provider: ModelProvider,
+    ) -> anyhow::Result<PathBuf> {
+        let cache_dir = self.get_cache_dir(provider);
+        let model_path = download::get_provider(provider)
             .get_model_path(model_name, cache_dir)
```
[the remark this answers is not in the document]
andrewpaprotsky: This is a valid inconsistency, but it predates this PR.

The direct-download fallback paths were already rediscovering the cache via CacheConfig::discover() / download_model_directly() instead of using the client-held cache config. This change only made get_model_path() use the new provider-aware resolver, which surfaced the mismatch more clearly.

So I see this as follow-up work, not a regression introduced here.

modelexpress_client/src/lib.rs
```diff
@@ -40,6 +38,117 @@ pub struct Client {
...
+                component,
+                Component::CurDir
+                    | Component::ParentDir
+                    | Component::RootDir
+                    | Component::Prefix(_)
+            )
+        })
+}
+
+fn prepare_stream_model_dir(
```
nicolasnoble, line 52: All the error paths in here wrap local filesystem failures (`create_dir_all`, `canonicalize`) as `Error::Server`. If the client runs out of disk space or hits a permission error, the user sees "Server error: Failed to create model directory" which would send them chasing a server-side problem that doesn't exist. Did you consider an `Error::Cache` or `Error::Io` variant instead?

modelexpress_client/src/lib.rs
```diff
@@ -40,6 +38,117 @@ pub struct Client {
     cache_config: Option<CacheConfig>,
 }
 
+fn is_safe_relative_path(path: &std::path::Path) -> bool {
+    !path.as_os_str().is_empty()
+        && !path.components().any(|component| {
+            matches!(
+                component,
+                Component::CurDir
```
nicolasnoble: The old check allowed `CurDir` (`.`) components - the new one rejects them. A path like `./config.json` is harmless but would now fail with "potentially unsafe file path." Is this intentional? Worth checking whether HF servers ever emit paths with a leading `./`.

modelexpress_server/src/services.rs
```diff
@@ -339,15 +348,19 @@ impl ModelService for ModelServiceImpl {
 
                     let is_last_chunk = offset.saturating_add(bytes_read as u64) >= *total_size;
 
+                    let first_chunk = is_first_chunk;
```
nicolasnoble: This reads a bit awkwardly with two names for the same boolean in the same scope. `std::mem::replace` would collapse it to one line:
```rust
let first_chunk = std::mem::replace(&mut is_first_chunk, false);
```

modelexpress_client/src/lib.rs
andrewpaprotsky: fixed by https://github.com/ai-dynamo/modelexpress/pull/172/commits/8b9fb50e2dd657b05037d6c80450cde757440a44

modelexpress_client/src/lib.rs
andrewpaprotsky: fixed by https://github.com/ai-dynamo/modelexpress/pull/172/commits/f4783bbcea70f269647eda538465d1f5ace2c835

modelexpress_server/src/services.rs
andrewpaprotsky: fixed by https://github.com/ai-dynamo/modelexpress/pull/172/commits/ec229afecdbb070f7735131ff8d3a816879de34b

PR #199, opened 2026-04-04
feat: add safety checks for P2P weight transfer

nicolasnoble:
P2P weight transfer currently has no validation that source and target
environments are compatible. Models using DeepSeek MLA attention produce
garbage inference after transfer, while standard GQA/MHA models transfer
correctly. The exact root cause in MLA post-processing is not yet
identified, but the failure is consistent and reproducible.

This PR adds three safety mechanisms:

- Strict model architecture allow-list gating P2P transfers. Only explicitly
  validated model_types, quantization methods, and dtypes are permitted.
  Unknown architectures fall back to disk loading. Bypass with
  MX_SKIP_FEATURE_CHECK=1.
- Manifest mismatch detection that raises ManifestMismatchError on tensor name
  or size mismatches between source and target. On mismatch, tries the next
  source candidate (important during rolling updates), then falls back to disk.
  Previously mismatches were silently skipped.
- TransferFingerprint logging capturing runtime environment for debugging.

Validated on nScale B200 cluster:

| Model | Architecture | Quantization | P2P Transfer |
|-------|-------------|--------------|--------------|
| Llama-3.1-8B-Instruct | llama (GQA) | none (BF16) | PASS |
| Llama-3.1-8B-Instruct-FP8 | llama (GQA) | fp8 | PASS |
| Llama-3-8B-Instruct-AWQ | llama (GQA) | awq | PASS |
| Llama-3-8B-Instruct-GPTQ | llama (GQA) | gptq | PASS |
| Mistral-7B-Instruct-v0.3 | mistral (GQA) | none (BF16) | PASS |
| Qwen2-7B-Instruct | qwen2 (GQA) | none (BF16) | PASS |
| Phi-3.5-mini-instruct | phi3 (GQA) | none (BF16) | PASS |
| Gemma-2-2b-it | gemma2 (GQA) | none (BF16) | PASS |
| DeepSeek-V2-Lite-Chat-FP8 | deepseek_v2 (MLA) | fp8 | FAIL |

## Release Notes

* **New Features**
  * Enhanced transfer safety with model feature validation and manifest compatibility checks
  * Added environment-variable controls for transfer behavior tuning
  * Improved transfer fingerprinting to detect incompatible environments

* **Bug Fixes**
  * Better error detection and handling for incompatible transfers between different environments
  * Stricter validation of tensor manifests during weight transfers

* **Documentation**
  * Updated architecture documentation with new 4-stage transfer pipeline and safety mechanisms
  * Added new environment variable reference for deployment configuration

* **Tests**
  * Added comprehensive test coverage for transfer safety validation

### DIFF
--- a/modelexpress_client/python/modelexpress/nixl_transfer.py
+++ b/modelexpress_client/python/modelexpress/nixl_transfer.py
@@ -19,7 +19,7 @@ from typing import Any
 
 import torch
 
-from .types import TensorDescriptor
+from .types import ManifestMismatchError, TensorDescriptor
 
 logger = logging.getLogger("modelexpress.nixl_transfer")
 
@@ -134,7 +134,7 @@ class NixlTransferManager:
 
         for name, tensor in tensors.items():
             if not tensor.is_contiguous():
-                raise RuntimeError(
+                raise ManifestMismatchError(
                     f"Tensor '{name}' is not contiguous. "
                     "Non-contiguous tensors cannot be used for RDMA transfers."
                 )
@@ -331,7 +331,7 @@ class NixlTransferManager:
             if self._registered_regions is None:
                 logger.error("Source sent region descriptors but we didn't register regions!")
                 logger.error("Set MX_CONTIGUOUS_REG=1 on target to enable region transfer")
-                raise RuntimeError("Region transfer mismatch: target must also use MX_CONTIGUOUS_REG=1")
+                raise ManifestMismatchError("Region transfer mismatch: target must also use MX_CONTIGUOUS_REG=1")
 
             logger.info(f"Region-based transfer: {len(source_tensors)} source regions -> {len(self._registered_regions)} local regions")
 
@@ -377,11 +377,53 @@ class NixlTransferManager:
             local_tensor_list = []
             total_bytes = 0
             matched_tensors = 0
+            skipped_source = []
+            skipped_target = []
+
+            source_names = {t.name for t in source_tensors}
+            local_names = set(self._tensors.keys())
+
+            # Detect manifest mismatches BEFORE transfer
+            source_only = source_names - local_names
+            target_only = local_names - source_names
+            if source_only:
+                skipped_source = sorted(source_only)
+                logger.warning(
+                    f"[Manifest Mismatch] {len(source_only)} source tensors have no "
+                    f"local match (will NOT be transferred): "
+                    f"{skipped_source[:5]}{'...' if len(skipped_source) > 5 else ''}"
+                )
+            if target_only:
+                skipped_target = sorted(target_only)
+                logger.warning(
+                    f"[Manifest Mismatch] {len(target_only)} local tensors have no "
+                    f"source match (will keep dummy values): "
+                    f"{skipped_target[:5]}{'...' if len(skipped_target) > 5 else ''}"
+                )
+            if source_only or target_only:
+                raise ManifestMismatchError(
+                    f"[Manifest Mismatch] Source has {len(source_names)} tensors, "
+                    f"target has {len(local_names)} tensors, "
+                    f"{len(source_names & local_names)} matched. "
+                    f"Source-only: {skipped_source[:5]}, "
+                    f"Target-only: {skipped_target[:5]}. "
+                    f"Aborting transfer to prevent incorrect inference."
+                )
 
             for src_tensor in source_tensors:
                 if src_tensor.name not in self._tensors:
                     continue
                 local_tensor = self._tensors[src_tensor.name]
+
+                # Validate size match for each tensor
+                local_size = local_tensor.numel() * local_tensor.element_size()
+                if src_tensor.size != local_size:
+                    raise ManifestMismatchError(
+                        f"[Manifest Mismatch] Tensor '{src_tensor.name}' size mismatch: "
+                        f"source={src_tensor.size}, target={local_size}. "
+                        f"Aborting transfer to prevent incorrect inference."
+                    )
+
                 remote_descs.append((src_tensor.addr, src_tensor.size, src_tensor.device_id))
                 local_tensor_list.append(local_tensor)
                 total_bytes += src_tensor.size
@@ -471,7 +513,7 @@ class NixlTransferManager:
                 break
             if status in ("ERR", "ERROR", "FAIL"):
                 self._agent.release_xfer_handle(handle)
-                raise RuntimeError(f"Transfer failed with status {status}")
+                raise ManifestMismatchError(f"Transfer failed with status {status}")
             time.sleep(0.001)
 
         # CRITICAL: Synchronize CUDA to ensure RDMA writes are visible
--- a/modelexpress_client/python/modelexpress/nixl_transfer.py
+++ b/modelexpress_client/python/modelexpress/nixl_transfer.py
@@ -513,7 +513,7 @@ class NixlTransferManager:
                 break
             if status in ("ERR", "ERROR", "FAIL"):
                 self._agent.release_xfer_handle(handle)
-                raise ManifestMismatchError(f"Transfer failed with status {status}")
+                raise RuntimeError(f"NIXL transfer failed with status {status}")
             time.sleep(0.001)
 
         # CRITICAL: Synchronize CUDA to ensure RDMA writes are visible
--- a/modelexpress_client/python/modelexpress/transfer_safety.py
+++ b/modelexpress_client/python/modelexpress/transfer_safety.py
@@ -239,6 +239,25 @@ def get_deep_gemm_version() -> str:
         return "unknown"
 
 
+def get_gpu_capability() -> str:
+    """Get the GPU compute capability as 'major.minor' (e.g. '9.0', '10.0').
+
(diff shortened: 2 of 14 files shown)


modelexpress_client/python/modelexpress/transfer_safety.py
```diff
@@ -0,0 +1,301 @@
...
+    Published by the source alongside tensor metadata. The target computes
+    its own fingerprint and compares. Mismatches cause transfer rejection.
+    """
+    vllm_version: str = ""
+    torch_version: str = ""
+    cuda_version: str = ""
+    deep_gemm_version: str = ""
+    attention_backend: str = ""
+    manifest_hash: str = ""
+    tensor_count: int = 0
```
zhengluo-nv, line 177: How many of these can be moved into SourceIdentity? Because I don't want the client to try and fail multiple times and increases latency

nicolasnoble: Yes, sorry, I said I'd split, and then didn't.

nicolasnoble: Done. Moved vllm_version, torch_version, cuda_version, and deep_gemm_version into extra_parameters so they're hashed into mx_source_id. Incompatible sources get filtered at ListSources. Attention backend stays post-load since it's not resolved until the model initializes.

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -471,7 +513,7 @@ def receive_from_source(
                 break
             if status in ("ERR", "ERROR", "FAIL"):
                 self._agent.release_xfer_handle(handle)
-                raise RuntimeError(f"Transfer failed with status {status}")
+                raise ManifestMismatchError(f"Transfer failed with status {status}")
             time.sleep(0.001)
```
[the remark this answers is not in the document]
nicolasnoble: Fixed. Changed to RuntimeError - the caller wraps it as SourceTransferError. Also added a ManifestMismatchError pass-through so real manifest mismatches don't get misclassified.

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -799,6 +825,16 @@ def _load_as_target(
...
 
+        # Log transfer fingerprint for debugging environment mismatches.
+        target_fp = TransferFingerprint.from_environment(self._tensors)
+        logger.info(
+            f"[Worker {global_rank}] Transfer fingerprint: "
+            f"vllm={target_fp.vllm_version}, cuda={target_fp.cuda_version}, "
+            f"attn={target_fp.attention_backend}, "
+            f"deepgemm={target_fp.deep_gemm_version}, "
+            f"manifest={target_fp.manifest_hash[:12]}... ({target_fp.tensor_count} tensors)"
+        )
```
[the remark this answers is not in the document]
nicolasnoble: Fixed. Moved the fingerprint log after _receive_from_peer which calls _register_tensors.

modelexpress_client/python/modelexpress/transfer_safety.py
```diff
@@ -0,0 +1,353 @@
...
+# Feature allow-list
+# ---------------------------------------------------------------------------
+
+# Model features that are known to produce correct results via P2P weight transfer.
+# Any feature not on this list causes automatic fallback to disk loading.
+# Add features here only after validating RDMA correctness end-to-end.
+# Model architectures validated for P2P weight transfer.
+# Only models whose model_type is in this set are allowed.
+# Any unknown architecture falls back to disk loading.
+ALLOWED_MODEL_TYPES: set[str] = {
```
AndyDai-nv: How about qwen3.5 and many other popular models?

modelexpress_client/python/modelexpress/transfer_safety.py
```diff
@@ -0,0 +1,353 @@
...
+            "tensor_count": self.tensor_count,
+        }, sort_keys=True)
+
+    @classmethod
+    def from_json(cls, s: str) -> TransferFingerprint:
+        """Deserialize from JSON string."""
+        d = json.loads(s)
+        return cls(**d)
+
+    def validate_against(self, source: TransferFingerprint) -> tuple[bool, list[str]]:
```
zhengluo-nv: Somehow I cannot find the usage of this method?

modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -947,6 +996,17 @@ def _load_as_source(
 
         # Register tensors + publish metadata
         self._register_tensors(model, global_rank, device_id)
+
+        # Log transfer fingerprint so targets can compare
+        source_fp = TransferFingerprint.from_environment(self._tensors)
```
zhengluo-nv, line 1001: Where is the source_fp used? Is there something missing?

modelexpress_client/python/modelexpress/transfer_safety.py
zhengluo-nv: Same concern: why do we need an allowlist even if we limited transfer fingerprint exact match? This looks like a even bigger limitation of the transfer strategy

AndyDai-nv (commented):
Quick questions:

1. What problem does this solve exactly?

The description says MLA models produce garbage inference after transfer, but the fix is an allow-list that blocks MLA transfers rather than fixing the root cause. This is a workaround — do we have a plan to actually fix MLA transfers, or is this intended to be permanent? Also is MLA really the root cause and the models we claim to support are verified? I have seen garbage output issue for normal qwen2 model(it is allowed in the PR) as well 


2. Do we really want to maintain a hardcoded model allow-list?

ALLOWED_MODEL_TYPES has some entries and is missing many popular models like qwen3.5, gpt-oss and others. Every new architecture requires someone to manually add it and validate end-to-end. Meanwhile users running unlisted models silently fall back to disk loading with no obvious indication why P2P isn't working. This seems like it'll cause more confusion than it prevents.  


3. Is TransferFingerprint over-engineered?

It collects vllm_version, torch_version, cuda_version, deep_gemm_version, gpu_capability — but validate_against only actually compares attention_backend and manifest_hash. The rest is gathered but unused. If the intent is "for logging", it doesn't need to be a fingerprint validation framework.

modelexpress_client/python/modelexpress/transfer_safety.py
nicolasnoble: Dropped the model_type allow-list entirely. The only feature we block now is MLA (detected via `kv_lora_rank` in the HF config). All other models - qwen3.5, whatever comes next - are allowed by default.

nicolasnoble: Agreed - removed the allow-list. The check now only blocks MLA models, which have confirmed weight corruption after RDMA transfer across all vLLM versions we've tested. Everything else passes through.

modelexpress_client/python/modelexpress/transfer_safety.py
nicolasnoble: It's called by `TransferFingerprint.from_environment()` - the manifest hash is computed from tensors at fingerprint creation time.

modelexpress_client/python/modelexpress/vllm_loader.py
nicolasnoble: It's logged for debugging - when a target reports corruption, operators can compare source and target fingerprints from the logs. Cross-validation (target rejects if fingerprints don't match) is deferred since it requires proto changes to publish the fingerprint.

modelexpress_client/python/modelexpress/transfer_safety.py
zhengluo-nv: Still cannot find this method being used anywhere. The TransferFingerprint looks like only for logging. Remove for now?

modelexpress_client/python/modelexpress/transfer_safety.py
```diff
@@ -0,0 +1,255 @@
...
+    if os.environ.get("MX_SKIP_FEATURE_CHECK", "0") == "1":
+        logger.warning("MX_SKIP_FEATURE_CHECK=1: bypassing feature checks")
+        return True, "bypassed"
+
+    features = detect_model_features(model_config)
+
+    if features["attention"] == "mla":
+        reason = (
+            "MLA attention models are blocked from P2P transfer due to "
+            "unresolved weight corruption (NVBug 6066010)"
```
zhengluo-nv: NVBug is internal. Can we expose this detail?

zhengluo-nv (commented):
2 minor comments

modelexpress_client/python/modelexpress/transfer_safety.py
nicolasnoble: Was always unused - it was built for future cross-validation that got deferred. Removed it along with the tests.

modelexpress_client/python/modelexpress/transfer_safety.py
nicolasnoble: Good catch, removed the internal reference.

zhengluo-nv (approved):

PR #208, opened 2026-04-07
feat: add GCS provider backend and cache model

andrewpaprotsky:
- Add `ModelProvider::Gcs` and wire GCS conversions through common, gRPC, and database layers.
- Introduce a manifest-backed GCS cache layout rooted under `gcs/<bucket>/...` with canonical `gs://` model names.
- Implement resumable GCS downloads with CRC32C verification, overlap protection, and atomic manifest replacement.

## Release Notes

* **New Features**
  * Added Google Cloud Storage (GCS) as a model provider option.

* **Bug Fixes**
  * Improved error message formatting for connection failures.
  * Enhanced validation of model file streaming to ensure data integrity and prevent incomplete downloads.

* **Chores**
  * Added dependencies for Google Cloud Storage support.

### DIFF
--- a/modelexpress_common/src/providers/gcs.rs
+++ b/modelexpress_common/src/providers/gcs.rs
@@ -1,0 +1,571 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::providers::ModelProviderTrait;
+use anyhow::{Context, Result};
+use google_cloud_storage::client::{Storage, StorageControl};
+use std::fs;
+use std::path::{Path, PathBuf};
+use tracing::info;
+
+mod cache_manifest;
+mod download_progress;
+mod download_task;
+mod downloader;
+mod lock_file;
+mod model_dir;
+mod model_name;
+mod path_ext;
+mod provider_cache;
+
+use downloader::Downloader;
+use model_dir::ModelDir;
+use model_name::ModelName;
+
+pub use provider_cache::GcsProviderCache;
+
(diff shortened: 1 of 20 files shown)


modelexpress_server/src/services.rs
```diff
@@ -291,6 +285,23 @@ impl ModelService for ModelServiceImpl {
...
+            error!(
+                "Resolved model path '{}' does not match expected cache layout '{}' for model '{}'",
+                model_path.display(),
+                expected_model_path.display(),
+                model_name
+            );
+            return Err(Status::internal(
+                "Resolved model path does not match expected cache layout",
+            ));
+        }
```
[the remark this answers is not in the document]
andrewpaprotsky: This would require prior local tampering of the server cache. Our current threat model assumes the cache directory is trusted and only populated by the configured providers; normal Hugging Face downloads do not create symlinks outside the cache root. Given that, I’m treating this as defense-in-depth rather than a blocking issue.

I don’t think that hardening belongs in this commit, though, because it is a separate defense-in-depth change with broader behavior impact; it should be handled in a focused follow-up.

modelexpress_common/src/providers/gcs.rs
```diff
@@ -0,0 +1,3037 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: What does this file do? Why is it so long? We need some modularization here to break it down properly

andrewpaprotsky: split `gcs.rs` into multiple submodules

zhengluo-nv:
Merged another PR. This one is duplicated

PR #217, opened 2026-04-09
fix: resolve HuggingFace cache path in model validate command

nicolasnoble:
`model validate <model-name>` always reported "not found in storage" for HuggingFace-downloaded models because it constructed the path as `<cache>/<org>/<model>` instead of the HF cache convention `<cache>/models--<org>--<model>/snapshots/<rev>`.

- Use the existing `resolve_model_path()` helper to find models in HF layout, falling back to a plain path join for other providers


zhengluo-nv (approved):

saturley-hall:
I am going to take this but please pick this back to `main` if it is relevant to be there

PR #219, opened 2026-04-09
fix: revert "fix:only start NIXL listener when P2P metadata is enabled (#211)"

nv-hwoo:
This reverts commit 37d85c3f6a59a16d4f9109b3dea8f41751031b0a.

The NIXL listen thread must always be enabled by default. MX_P2P_METADATA only controls whether the source starts the worker gRPC service for P2P metadata exchange vs using centralized storage for all blobs. It should not gate the NIXL listener.

Keeps the doc fix for MX_METADATA_PORT default (5555, not 0).


nicolasnoble (approved):
LGTM

PR #227, opened 2026-04-13
fix: propagate vLLM log handlers to modelexpress loggers

AndyDai-nv:
## Summary

- vLLM 0.19.0's EngineCore subprocess only attaches log handlers to the `"vllm"` namespace. All `"modelexpress.*"` loggers propagate to the root logger which has no handler, silently dropping all MxModelLoader output (startup, metadata publish, heartbeat, retry warnings, errors).
- Copies vLLM's handlers onto the `"modelexpress"` parent logger in `MxModelLoader.__init__()` so all child loggers inherit them via propagation. Idempotent guard prevents duplicate handlers.

Bug: https://nvbugspro.nvidia.com/bug/6071055

## Test plan

- [x] Unit tests added in `TestConfigureVllmLogging` (3 tests):
  - Handler propagation from `"vllm"` to `"modelexpress"` parent logger
  - Idempotency (no duplicate handlers on repeated calls)
  - Log output actually captured through vLLM's handler
- [x] All 50 existing tests in `test_vllm_loader.py` pass

### DIFF
--- a/modelexpress_client/python/modelexpress/vllm_loader.py
+++ b/modelexpress_client/python/modelexpress/vllm_loader.py
@@ -24,6 +24,7 @@ Usage:
 from __future__ import annotations
 
 import logging
+import os
 import time
 import uuid
 
@@ -64,7 +65,10 @@ def _configure_vllm_logging():
     vllm_logger = logging.getLogger("vllm")
     for handler in vllm_logger.handlers:
         mx_root.addHandler(handler)
-    if vllm_logger.level != logging.NOTSET:
+    mx_level = os.environ.get("MODEL_EXPRESS_LOG_LEVEL", "").upper()
+    if mx_level and hasattr(logging, mx_level):
+        mx_root.setLevel(getattr(logging, mx_level))
+    elif vllm_logger.level != logging.NOTSET:
         mx_root.setLevel(vllm_logger.level)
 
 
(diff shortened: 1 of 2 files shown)


modelexpress_client/python/modelexpress/vllm_loader.py
```diff
@@ -47,6 +47,27 @@
...
+    inherits them via propagation.
+    """
+    mx_root = logging.getLogger("modelexpress")
+    if mx_root.handlers:
+        return
+    vllm_logger = logging.getLogger("vllm")
+    for handler in vllm_logger.handlers:
+        mx_root.addHandler(handler)
+    if vllm_logger.level != logging.NOTSET:
+        mx_root.setLevel(vllm_logger.level)
```
nv-hwoo, line 72: We might not want to follow vllm log level in case we just want to set different log levels for MX (e.g. for debugging).

AndyDai-nv: Good catch. MODEL_EXPRESS_LOG_LEVEL env var now takes priority over vLLM's level. If unset, falls back to vLLM's level. This lets users set e.g. MODEL_EXPRESS_LOG_LEVEL=DEBUG independently for MX debugging.

nv-hwoo (approved):

PR #233, opened 2026-04-13
fix: re-initialize model after failed strategy mutates it (FP8 disk fallback crash)

nicolasnoble:
Same root cause as #231 (release/0.3.0), adapted for the strategy chain architecture on main.

When RdmaStrategy enters _load_as_target(), it runs process_weights_after_loading() before the RDMA receive to establish tensor layout for receive buffers. For FP8/modelopt models, this strips weight_loader from QuantizedParameters. If RDMA then fails, subsequent strategies (DefaultStrategy) crash on model.load_weights() because weight_loader is gone.

- LoadStrategyChain.run() detects stale NIXL state after a failed strategy, cleans up, and re-initializes the model before trying the next strategy
- run() returns the (possibly re-initialized) model so the caller uses the correct object
- Detection is precise: RDMA is the only strategy that runs process_weights_after_loading() before its failure point, and NIXL state is only set when _load_as_target() actually ran

Fixes: NVBugs 6071387

## Release Notes

* **Bug Fixes**
  * Improved model loading resilience with enhanced error recovery for failed initialization strategies
  * Fixed model state corruption by automatically reinitializing models when partial mutations are detected during strategy failures

### DIFF
--- a/modelexpress_client/python/modelexpress/load_strategy/__init__.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/__init__.py
@@ -79,14 +79,7 @@ class LoadStrategyChain:
                     f"raised unexpected error, trying next: {e}"
                 )
 
-            # A failed strategy may have mutated the model (e.g. RDMA runs
-            # process_weights_after_loading() before the transfer attempt,
-            # stripping weight_loader from FP8 QuantizedParameters).  Detect
-            # this via stale NIXL state and re-initialize so the next
-            # strategy gets a clean model.
-            if ctx.tensors or ctx.nixl_manager is not None:
-                ctx.tensors = {}
-                ctx.nixl_manager = None
+            if strategy.rollback(ctx):
                 del model
                 torch.cuda.empty_cache()
                 logger.info(
--- a/modelexpress_client/python/modelexpress/load_strategy/__init__.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/__init__.py
@@ -39,6 +39,30 @@ __all__ = [
 logger = logging.getLogger("modelexpress.load_strategy")
 
 
+def _reset_vllm_compilation_state(compilation_config) -> None:
+    """Reset per-model mutable state on vLLM's CompilationConfig.
+
+    vLLM registers attention / MLA / Mamba / FusedMoE layers and accumulates
+    compile stats into dicts / sets / counters on ``compilation_config``
+    during ``initialize_model()``. These live on the config object, not the
+    model, so they survive ``del model`` and either crash the next
+    ``initialize_model()`` (``static_forward_context`` has an explicit
+    duplicate-name guard) or silently corrupt subsequent state (MoE layer
+    list, custom op counters, traced files, compilation time).
+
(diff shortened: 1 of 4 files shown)


modelexpress_client/python/modelexpress/load_strategy/__init__.py
```diff
@@ -65,13 +68,33 @@ def run(model: nn.Module, ctx: LoadContext) -> None:
...
                 if strategy.load(model, ctx):
-                    return
+                    return model
             except Exception as e:
                 logger.warning(
                     f"[Worker {ctx.global_rank}] Strategy {strategy.name} "
                     f"raised unexpected error, trying next: {e}"
                 )
 
+            # A failed strategy may have mutated the model (e.g. RDMA runs
```
zhengluo-nv: I think a cleaner way is to add a rollback method in the LoadStrategy interface, and let each strategy implements its own cleanup logic

zhengluo-nv: Then here we will need to call strategy.rollback() when an exception is raised

nicolasnoble: Good call. Added a rollback() to LoadStrategy with a default no-op, RdmaStrategy overrides it to clean up NIXL state. Chain calls it after each failure. See 56ee520.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -35,22 +34,27 @@ class RdmaStrategy(LoadStrategy):
 
     name = "rdma"
 
+    def rollback(self, ctx: LoadContext) -> bool:
```
zhengluo-nv, line 38: Can the rollback method actually execute the rollback, instead of returning true/false? I think for different strategy, the rollback steps are different so the base class should not apply the same rollback

nicolasnoble: The strategy can't own the full rollback - it can't replace the model object. Python reference semantics mean reassigning `model` inside rollback() doesn't affect the chain's variable. The chain has to handle del + re-init since it's the one holding the reference. The bool return lets each strategy define its own cleanup (NIXL state here, whatever else for future strategies) while the chain handles the one thing only it can do: swapping the model.

nicolasnoble: Another data point for this split, from the 6084603 investigation: the re-init also needs six fields on `vllm_config.compilation_config` cleared before `initialize_model()` can safely run again. `static_forward_context` crashes on duplicate; the other five (`static_all_moe_layers`, `enabled_custom_ops`, `disabled_custom_ops`, `traced_files`, `compilation_time`) silently corrupt state. Those entries weren't created by any strategy - they came from the chain's own `initialize_model()` call at the top of `MxModelLoader.load_model()`. Same argument as the model itself: chain owns the state it creates, so the cleanup lives in the chain. Pushed as `_reset_vllm_compilation_state()` (audited against vLLM 0.17.1).

zhengluo-nv (approved):

PR #236, opened 2026-04-14
fix: update example manifests and Helm docs to public 0.3.0 image URI

zhengluo-nv:
## Summary

- Update server example manifests from internal `nvcr.io/nvidian/dynamo-dev/modelexpress-server:latest` to public `nvcr.io/nvidia/ai-dynamo/modelexpress-server:0.3.0`
- Replace private `nvcr.io/nvidian/dynamo-dev/modelexpress-client` image in vLLM client manifests with a `<your-registry>` placeholder, pointing users to `examples/p2p_transfer_k8s/Dockerfile.client` to build their own image
- Update Helm README docker pull example from `0.2.0` to `0.3.0`

### Files changed

| File | Change |
|------|--------|
| `examples/p2p_transfer_k8s/server/kubernetes_backend/modelexpress-server-kubernetes.yaml` | server image → public `nvidia/ai-dynamo` registry, `0.3.0` |
| `examples/p2p_transfer_k8s/server/redis_backend/modelexpress-server-redis.yaml` | server image → public `nvidia/ai-dynamo` registry, `0.3.0` |
| `examples/p2p_transfer_k8s/client/vllm/vllm-multi-node.yaml` | client image → `<your-registry>` placeholder + Dockerfile.client comment |
| `examples/p2p_transfer_k8s/client/vllm/vllm-single-node-p2p.yaml` | client image → `<your-registry>` placeholder + Dockerfile.client comment |
| `examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml` | client image → `<your-registry>` placeholder + Dockerfile.client comment |
| `helm/README.md` | docker pull version bump `0.2.0` → `0.3.0` |

## Test plan

- [ ] Verify no remaining internal registry references in example manifests
- [ ] YAML manifests pass validation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #241, opened 2026-04-15
fix: adopt hidden CUDA tensors for correct P2P RDMA inference

zhengluo-nv:
## Summary

`process_weights_after_loading` creates CUDA tensors on plain Python objects attached to `nn.Module` attributes (e.g. `FusedMoEQuantConfig.a1_gscale = 1.0 / activation_scale`) that are invisible to `named_parameters()`/`named_buffers()`. On the RDMA target, these tensors retain values computed from dummy weights, causing **garbled inference despite all registered tensor bytes matching the source byte-for-byte**.

### Root Cause

For NVFP4 MoE layers, `make_nvfp4_moe_quant_config` computes reciprocal activation scales as new tensors:
```python
a1_gscale = 1.0 / a13_scale   # stored on FusedMoEQuantConfig, NOT on the layer
a2_gscale = 1.0 / a2_scale    # stored on FusedMoEQuantConfig, NOT on the layer
```
These are used during the MoE forward pass for activation quantization but are not `nn.Parameter` or `nn.Buffer`, so they are missing from the NIXL RDMA manifest. On the target, they hold `1.0 / dummy_scale` values.

### Fix

`adopt_hidden_tensors()` recursively scans each module's non-Module attributes for orphaned CUDA tensors and registers them as non-persistent buffers before NIXL collection. Called inside `register_tensors()` so all loading strategies (RDMA, default, GDS, ModelStreamer) get it automatically.

### Commits

1. **`fix: adopt hidden CUDA tensors for correct P2P RDMA inference`** — core fix (`adopt_hidden_tensors`, `_find_hidden_cuda_tensors`, debug logging, GPU-side `safe_checksum`)
2. **`fix: address review nitpicks`** — MX_SKIP_FEATURE_CHECK comment, vacuous test fix, zero-numel tensor filter
3. **`feat: add Dynamo DynamoGraphDeployment example`** — `examples/dynamo_p2p_transfer_k8s/` with Dockerfile, README, DGD manifest, CRD, RBAC; also replaces private nvcr.io/nvidian image refs with public images (from PR #236)

### Files Changed

| File | Change |
|------|--------|
| `tensor_utils.py` | `adopt_hidden_tensors()`, `_find_hidden_cuda_tensors()`, debug logging in `capture_tensor_attrs()`, GPU-side `safe_checksum()` |
| `load_strategy/base.py` | Call `adopt_hidden_tensors()` inside `register_tensors()` |
| `tests/test_tensor_utils.py` | 25 unit tests (16 CPU, 9 CUDA-gated) |
| `ARCHITECTURE.md` | Updated Transfer Safety, Tensor Discovery, FP8 Handling sections; added dynamo example to project structure |
| `DEPLOYMENT.md` | Documented `MODEL_EXPRESS_LOG_LEVEL` env var |
| `examples/dynamo_p2p_transfer_k8s/` | New: Dockerfile, README, DGD manifest (MX server + Frontend + VllmWorker), CRD, RBAC |
| `examples/p2p_transfer_k8s/**/*.yaml` | Replaced private `nvcr.io/nvidian/` image refs with public `nvcr.io/nvidia/ai-dynamo/` or `<your-registry>/` placeholders |
| `helm/README.md` | Updated server image tag to 0.3.0 |

## Adopted Hidden Tensors Analysis

Hidden CUDA tensors discovered per model/version:

### By category (Kimi-K2.5-NVFP4 on v0.17.1, PP stage 1)

| Category | Count | Shape | Source Object | Total Memory |
|----------|-------|-------|---------------|-------------|
| MoE activation scale reciprocal (a1) | 30 | [384] fp32 | `FusedMoEQuantConfig._a1.alpha_or_gscale` | 45 KiB |
| MoE activation scale reciprocal (a2) | 30 | [384] fp32 | `FusedMoEQuantConfig._a2.alpha_or_gscale` | 45 KiB |
| MoE weight global scale (w1) | 30 | [384] fp32 | `FusedMoEQuantConfig._w1.alpha_or_gscale` | 45 KiB |
| MoE weight global scale (w2) | 30 | [384] fp32 | `FusedMoEQuantConfig._w2.alpha_or_gscale` | 45 KiB |
| MoE fused output scale (g1) | 30 | [384] fp32 | `TrtllmNvFp4MoeExperts.g1_scale_c` | 45 KiB |
| FlashInfer MLA workspace | 2 | [128M] uint8 | `FlashInferMLAImpl._workspace_buffer` | 256 MiB |

### Which tensors caused the garbled inference?

The **a1_gscale** and **a2_gscale** tensors (60 total) — `1.0 / activation_scale` reciprocals computed from dummy weights and used directly in the TRT-LLM MoE kernel forward pass. The w1/w2 global scales and g1_scale_c also had wrong values. FlashInfer workspaces are zero-initialized buffers that don't cause correctness issues but are captured for completeness.

### Cross-version comparison

| Model | Quant | vLLM | Platform | Hidden adopted | Total manifest | Inference |
|-------|-------|------|----------|---------------|----------------|-----------|
| Kimi-K2.5-NVFP4 | NVFP4 | v0.19.0 | StatefulSet (native MP) | 61 | 1473 | exact match |
| Kimi-K2.5-NVFP4 | NVFP4 | v0.17.1 | StatefulSet (Ray) | 152 | 1654 | exact match |
| Kimi-K2.5-NVFP4 | NVFP4 | v0.16.0 | DynamoGraphDeployment | 1 | 1503 | exact match |
| DeepSeek-V3 | FP8 | v0.17.1 | Deployment (TP=8) | 2 | 1635 | exact match |

Note: DeepSeek-V3 FP8 P2P works correctly on `main` without `adopt_hidden_tensors` — the FP8 case was already fixed by `capture_tensor_attrs` + `storage_view` (PR #200). The `adopt_hidden_tensors` fix is specifically needed for **NVFP4 MoE** models where quant scales live on non-Module objects.

### Correctness (Kimi-K2.5-NVFP4, TP=4, PP=2, 8 GPUs)

| Test | vLLM v0.19.0 | vLLM v0.17.1 | vLLM v0.16.0 (Dynamo) |
|------|-------------|-------------|----------------------|
| Source inference (disk load) | correct | correct | correct |
| Target inference (P2P) | **correct** (was garbled) | **correct** | **correct** |
| Source vs Target (temp=0) | exact match | exact match | exact match |
| Chained P2P (target -> replica) | exact match | - | - |
| Multiple diverse prompts | all exact match | - | round-robin both replicas serving |

### Performance

| Metric | Value |
|--------|-------|
| Extra VRAM from adopted tensors | +84 MiB (~0.05%) |
| Transfer size increase | +0.13-0.17 GB (~0.2%) |
| Transfer bandwidth | 20-215 Gbps (varies by node pair) |
| `adopt_hidden_tensors` latency | 0.073-0.086s |

### Debugging

Set `MODEL_EXPRESS_LOG_LEVEL=DEBUG` to see:
- Each captured bare CUDA tensor (from `capture_tensor_attrs`)
- Each adopted hidden tensor with source object type and attribute path
- Per-tensor checksums during registration (GPU-side sum, no bulk CPU copy)

## Test plan

- [x] Unit tests: 25 tests (16 CPU-only pass, 9 CUDA-gated)
- [x] E2E: Source inference correct (disk load)
- [x] E2E: Target P2P inference matches source (temp=0, temp=1.0)
- [x] E2E: Chained P2P (disk -> source -> target -> replica) all match
- [x] E2E: Multiple diverse prompts produce identical output
- [x] E2E: Verified on vLLM v0.16.0 (Dynamo DGD), v0.17.1 (Ray), v0.19.0 (native MP)
- [x] E2E: DeepSeek-V3 FP8 confirmed working on main without fix (regression-safe)
- [x] VRAM: No meaningful increase (+84 MiB / 0.05%)
- [x] Latency: `adopt_hidden_tensors` adds <0.1s to startup
- [x] Dynamo DGD: ModelExpress server + Frontend + multi-node VllmWorker deployed and tested


docs/ARCHITECTURE.md
```diff
@@ -510,7 +517,7 @@ Each strategy is fully self-contained: it handles weight loading, post-processin
...
 
-MLA models produce correct inference when loaded from disk but corrupted inference after RDMA weight transfer. The bytes transfer correctly (checksums match), but inference diverges. This has been reproduced across all tested vLLM versions (0.12.0 through 0.17.1). Root cause is under investigation. Set `MX_SKIP_FEATURE_CHECK=1` to bypass the block for debugging.
+MLA models previously produced corrupted inference after RDMA transfer despite all registered tensor bytes matching. The root cause was hidden CUDA tensors stored on quant method objects (e.g. `FusedMoEQuantConfig.a1_gscale = 1/activation_scale`) that were invisible to `named_parameters()`/`named_buffers()` and computed from dummy weights before RDMA. The fix (`adopt_hidden_tensors()`) recursively scans module attributes for orphaned CUDA tensors and registers them as non-persistent buffers so they are included in the RDMA manifest. Verified correct on vLLM v0.17.1 and v0.19.0 with Kimi-K2.5-NVFP4. Set `MX_SKIP_FEATURE_CHECK=1` to bypass the MLA feature gate.
```
nicolasnoble: This reads as "MLA was the failure class, now fixed," but this is specifically an NVFP4 MoE hidden-tensor issue - DSV3 (MLA + FP8) works on main without this fix, Kimi (MLA + NVFP4 MoE) needs it. Could we tighten to something like "NVFP4 MoE models (known case: Kimi-K2.5) previously produced corrupted inference..."? Otherwise the next person hitting an MLA regression on a different model might assume it's a new bug rather than the same class of "post-processing stashes state on non-Module objects."

modelexpress_client/python/modelexpress/tensor_utils.py
```diff
@@ -16,12 +15,13 @@
 
 
 def safe_checksum(tensor: torch.Tensor) -> str:
```
nicolasnoble, line 17: Raw byte-sum is a pretty weak "checksum" - `[1, 255]` and `[255, 1]` collide, and flipping one byte up by 1 and another down by 1 also collides. For per-tensor debug logging that's a lot of potential false "matches."

Classic djb2 won't parallelize cleanly on GPU, but we can get djb2-flavor mixing at roughly the same cost as current via position-weighted multiply with Knuth's constant:

```python
idx = torch.arange(n, device=flat.device, dtype=torch.int64)
weights = (idx * 2654435761 + 1) & 0xFFFFFFFF
h = ((flat.to(torch.int64) * weights) & 0xFFFFFFFF).sum().item() & 0xFFFFFFFF
```

One extra multiply over what's there now, permutation-sensitive, much better bit diffusion. Alternatively, if we want to keep the current impl, could we rename to `safe_rolling_sum` so the name matches what it actually does?

modelexpress_client/python/modelexpress/tensor_utils.py
```diff
@@ -64,6 +68,97 @@ def capturing_setattr(self, name, value):
         nn.Module.__setattr__ = original_setattr
 
 
+def _find_hidden_cuda_tensors(
```
nicolasnoble, line 79: Could we add a known-limitation note to the docstring? The `hasattr(obj, "__dict__")` check correctly skips `type` and `nn.Module`, but it also silently skips any class using `__slots__`. No vLLM quant class uses slots today, but if upstream ever adopts them, hidden tensors on slotted objects would be missed with no error - which is exactly the bug this PR is fixing. Something like:

```python
"""Recursively find CUDA tensors in a non-Module Python object graph.

Known limitation: objects using ``__slots__`` are skipped because they
lack ``__dict__``. No current vLLM quant class uses slots, but any
upstream adoption would silently cause hidden tensors to be missed.
"""
```

nicolasnoble (commented):
This PR bundles core fix + checksum refactor + Dynamo DGD example + public image URI scrub. Fair given the inference day timing, no ask to split here. Just noting the DGD example would have been easier to review on its own since it's independent of the fix.

A few inline notes below.

modelexpress_client/python/modelexpress/tensor_utils.py
```diff
@@ -64,6 +68,97 @@ def capturing_setattr(self, name, value):
...
+                continue
+
+            tensors = _find_hidden_cuda_tensors(attr_val, visited=set())
+            for tensor_path, tensor in tensors:
+                if tensor.data_ptr() in existing_ptrs:
+                    continue
+                safe_path = (
+                    tensor_path.replace(".", "_").replace("[", "").replace("]", "")
+                )
+                buf_name = f"_mx_{attr_name}_{safe_path}"
```
KavinKrishnan, line 155: just to be safe:

if two different attributes on the same module produce hidden tensors with overlapping sanitized paths, the second register_buffer would silently overwrite the first.

i.e

```
layer.quant.a1.scale = torch.tensor([1.0], device="cuda")
layer.quant.a1_scale = torch.tensor([2.0], device="cuda")
```

probably want to add:

if hasattr(module, buf_name):
    buf_name = f"{buf_name}_{adopted}" 

or

safe_path = tensor_path.replace(".", "__dot__")

docs/ARCHITECTURE.md
KavinKrishnan: Should we make MX_SKIP_FEATURE_CHECK=1 the default  now that adopt_hidden_tensors fixes the MLA corruption

modelexpress_client/python/modelexpress/tensor_utils.py
zhengluo-nv: Updated using the proposed algorithm. This is only used in debug log level, so has no effect to the actual workload. The previous md5 hash is really really slow

modelexpress_client/python/modelexpress/tensor_utils.py
zhengluo-nv: Added comment

modelexpress_client/python/modelexpress/tensor_utils.py
zhengluo-nv: Updated

docs/ARCHITECTURE.md
zhengluo-nv: According to Nico we should keep the gate until we test enough number of MLA models

zhengluo-nv: This should motivate the team to add automated CI testing on different models

zhengluo-nv: Updated doc

zhengluo-nv:
Addressed comments

AndyDai-nv (approved):

AndyDai-nv:
LGTM, better wait for others to take a look since this is a big change

nicolasnoble (approved):

PR #242, opened 2026-04-15
fix: only re-initialize model when RDMA target load was actually attempted

nicolasnoble:
The re-initialization added in PR #231 triggers on transfer_allowed,
which is true whenever P2P is configured. On the very first source pod
startup (no existing sources), no RDMA attempt is made and the model is
never mutated, but the re-init path runs anyway and calls
initialize_model() a second time, crashing vLLM with
"Duplicate layer name".

- Return (loaded, attempted) tuple from _try_load_from_candidates()
- Track whether _load_as_target() was actually called
- Gate re-initialization on rdma_attempted instead of transfer_allowed


zhengluo-nv:
CI tests failed

PR #244, opened 2026-04-16
fix: clear vLLM compilation_config registries before model re-init

nicolasnoble:
PR #231's re-init path (further narrowed by PR #242 to rdma_attempted) does
del model + initialize_model() after a failed RDMA target attempt, but
leaves vllm_config.compilation_config.static_forward_context populated with
the layer registrations from the first init. The second initialize_model()
trips vLLM's duplicate-layer-name guard at layer 0, so the target pod crashes
on its first recovery attempt after the source pod is scaled down.

- Clear static_forward_context and static_all_moe_layers on the vllm_config
  before the re-init
- Audited against vLLM 0.17.1; other vLLM versions may add init=False fields
  on CompilationConfig that need similar treatment

### DIFF
--- a/modelexpress_client/python/modelexpress/vllm_loader.py
+++ b/modelexpress_client/python/modelexpress/vllm_loader.py
@@ -694,6 +694,17 @@ class MxModelLoader(BaseModelLoader):
                     _nixl_managers.pop(device_id, None)
                     del model
                     torch.cuda.empty_cache()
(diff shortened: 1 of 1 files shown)


nv-nmailhot (approved):

zhengluo-nv (approved):

zhengluo-nv:
Let's better test this end to end on a cluster or local

PR #251, opened 2026-04-22
feat: add k8s-service metadata backend

nicolasnoble:
Adds `MX_METADATA_BACKEND=k8s-service`: source pools behind a K8s
Service per TP rank, clients resolve via the Service DNS.

- `SourceIdentity.revision` feeds into mx_source_id; Rust/Python hash
  parity pinned in cross-check tests.
- `GetTensorManifestResponse` carries NIXL endpoint info; one Service
  round-trip opens a transfer.
- `K8sServiceMetadataClient` retries `FAILED_PRECONDITION` on fresh
  channels so kube-proxy re-picks backends on revision skew.
- `REQUIRES_P2P_METADATA` class attribute forces the P2P path;
  `MX_P2P_METADATA` stays opt-in for central-coord backends.
- Docs + example manifests for 1-GPU-per-pod and multi-GPU-per-pod
  shapes.


modelexpress_client/python/modelexpress/client_factory.py
```diff
@@ -0,0 +1,66 @@
...
+from .k8s_service_client import K8sServiceMetadataClient
+
+logger = logging.getLogger("modelexpress.client_factory")
+
+MetadataClient = Union[MxClient, K8sServiceMetadataClient]
+
+_CENTRAL_BACKEND_ALIASES = frozenset({
+    "", "server", "redis", "kubernetes", "k8s", "crd",
+})
+_K8S_SERVICE_ALIASES = frozenset({"k8s-service", "service"})
```
zhengluo-nv, line 34: nit: Can we use use a generic name like "decentralized" or "distributed".

modelexpress_client/python/modelexpress/metadata.py
```diff
@@ -238,8 +265,31 @@ def _publish_metadata_to_server(
...
+    For backends that DON'T force it (``MxClient`` backed by the
+    central server), the ``MX_P2P_METADATA`` env var controls
+    whether the source publishes lightweight pointers (and serves
+    the full metadata itself) or full metadata to the server.
+    """
+    # Strict identity check against True so MagicMock's auto-attribute
+    # (and any other non-literal truthy value) doesn't accidentally
+    # force the P2P path in tests or misconfigured clients.
+    if getattr(mx_client, "REQUIRES_P2P_METADATA", False) is True:
+        env_value = os.environ.get("MX_P2P_METADATA", "")
```
zhengluo-nv, line 281: I would suggest to set MX_P2P_METADATA default to True. Later we can enable p2p metadata and clean up the old path.

modelexpress_client/python/modelexpress/k8s_service_client.py
```diff
@@ -0,0 +1,238 @@
...
+        self._worker_rank = worker.worker_rank
+        source_id = compute_mx_source_id(identity)
+        logger.info(
+            "K8sServiceMetadataClient.publish_metadata: "
+            "computed mx_source_id=%s for worker_rank=%d",
+            source_id, worker.worker_rank,
+        )
+        return source_id
+
+    def list_sources(
```
zhengluo-nv, line 130: List method is supposed to return a list of sources that match the source identity. I think we should query the k8s service and filter endpoints that match source identity and status filter. The source identity can be populated by the vllm loader via a label.

The current implementation is a shortcut. It relies on k8s service round robin to select an endpoint for you. We have 0 control over which endpoint/source to initialize transfer, also no way to filter out these unhealthy/unmatched sources if the k8s service is not configured correctly.

modelexpress_client/python/modelexpress/k8s_service_client.py
```diff
@@ -0,0 +1,238 @@
...
+from .source_id import compute_mx_source_id
+
+logger = logging.getLogger("modelexpress.k8s_service_client")
+
+_DEFAULT_SERVICE_PATTERN = "mx-sources-rank-{rank}:6555"
+_DEFAULT_MAX_RETRIES = 5
+_DEFAULT_BACKOFF_SECONDS = 0.5
+
+
+class K8sServiceMetadataClient:
```
zhengluo-nv: Since this class and the MxClient share the same traits. We can introduce an ABC, and rename these 2 classes MxClient and MxK8sServiceClient

modelexpress_client/python/modelexpress/k8s_service_client.py
```diff
@@ -0,0 +1,238 @@
...
+            raise RuntimeError(
+                "K8sServiceMetadataClient needs a worker_rank before "
+                "list_sources can resolve the Service endpoint; pass "
+                "worker_rank to the constructor or call publish_metadata "
+                "first"
+            )
+        source_id = compute_mx_source_id(identity)
+        ref = p2p_pb2.SourceInstanceRef(
+            mx_source_id=source_id,
+            worker_id=f"svc-rank-{self._worker_rank}",
```
zhengluo-nv: This hardcoded worker id seems wrong and not work with multiple replicas scenario

modelexpress_client/python/modelexpress/metadata.py
nicolasnoble: Agreed on the direction, and I'm fine flipping this for the next release. But defaulting to 0 was Ganesh's call when we first added this flag - he wanted the old path to stay default until P2P was battle-tested. I'd want his sign-off before we flip. @ganeshku1 wdyt?

modelexpress_client/python/modelexpress/k8s_service_client.py
nicolasnoble: As discussed on Slack, the user specifically asked for this shape: expose a Service, let K8s load-balance, period. No client-side endpoint enumeration, no RBAC for the endpoints API - that's the coupling they wanted to avoid with this backend.

Per-backend addressability via K8s-native endpoint discovery is a different thing - worth considering as a separate backend variant (`k8s-endpoints` or similar) if a user asks for it, but it'd re-introduce the infrastructure requirements the k8s-service design was built to skip. Keeping this one pool-level.

modelexpress_client/python/modelexpress/k8s_service_client.py
nicolasnoble: Done. Added `MxClientBase(ABC)` in `client.py` with the 5 shared abstract methods and a `REQUIRES_P2P_METADATA = False` class-level default. `MxClient` inherits from it; `K8sServiceMetadataClient` renamed to `MxK8sServiceClient` and also inherits. `LoadContext.mx_client` narrowed from the Union to `MxClientBase`. All tests pass.

modelexpress_client/python/modelexpress/metadata.py
ganeshku1: > Agreed on the direction, and I'm fine flipping this for the next release. But defaulting to 0 was Ganesh's call when we first added this flag - he wanted the old path to stay default until P2P was battle-tested. I'd want his sign-off before we flip. @ganeshku1 wdyt?

Yes, i it is fine flipping this for next release and we should have good enough CI coverage by then.

modelexpress_client/python/modelexpress/k8s_service_client.py
nicolasnoble: Changed the synthetic `worker_id` to `""` with a comment explaining why: this backend has no per-pod addressability (kube-proxy picks the backend at connection time), `get_metadata` doesn't read the field, and stamping a constructed ID there suggested semantics that don't exist here. Tied to the other thread where we agreed to keep this backend pool-level.

modelexpress_client/python/modelexpress/k8s_service_client.py
```diff
@@ -0,0 +1,238 @@
...
+                    "%d tensors, attempt=%d)",
+                    endpoint, resp.mx_source_id, resp.worker_rank,
+                    len(resp.tensors), attempt,
+                )
+                return p2p_pb2.GetMetadataResponse(
+                    found=True,
+                    worker=worker,
+                    mx_source_id=resp.mx_source_id,
+                    worker_id=worker_id,
+                )
```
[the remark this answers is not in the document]
nicolasnoble: Fixed. `get_metadata` validates `resp.mx_source_id == mx_source_id` and `resp.worker_rank == self._worker_rank` after the response; mismatches retry on a fresh channel up to `MX_K8S_SOURCE_RETRIES`, then raise. Tests cover both paths.

modelexpress_client/python/modelexpress/load_strategy/base.py
```diff
@@ -96,7 +98,7 @@ def build_load_context(
         global_rank=global_rank,
         device_id=device_id,
         identity=build_source_identity(vllm_config, model_config),
-        mx_client=MxClient(),
+        mx_client=create_metadata_client(worker_rank=global_rank),
```
nicolasnoble, line 100: Fixed `publish_metadata` and `RdmaStrategy.is_available`: bypass the server-address gate when `ctx.mx_client.REQUIRES_P2P_METADATA is True`. Test added.

examples/k8s_service_sources/sources-tp2-single-pod.yaml
```diff
@@ -0,0 +1,148 @@
...
+            - name: VLLM_PLUGINS
+              value: "modelexpress"
+            - name: MX_METADATA_BACKEND
+              value: "k8s-service"
+            - name: MX_K8S_SERVICE_PATTERN
+              value: "mx-sources-rank-{rank}:6555"
+            - name: MX_METADATA_PORT
+              value: "5555"
+            - name: MX_WORKER_GRPC_PORT
+              value: "6555"
```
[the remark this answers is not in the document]
nicolasnoble: Fixed. All three YAMLs now pass `--tensor-parallel-size` explicitly via `args:`, matching the pattern in `examples/p2p_transfer_k8s/`. Multi-GPU-per-pod uses TP=2; 1-GPU-per-pod uses TP=1 with a comment that cross-pod TP requires external orchestration.

docs/ARCHITECTURE.md
```diff
@@ -352,15 +352,92 @@ Runs in a background tokio task on a configurable interval (default 1 hour). LRU
...
+
+Each worker publishes independently. The `mx_source_id` is a 16-char hex key computed from `SHA256(canonical_json(SourceIdentity))` where `SourceIdentity` includes a `revision` field for content-addressed identity (HuggingFace commit SHA, S3 object version, or a deployer-provided string). Two sources with identical `mx_source_id` are guaranteed to have bit-identical weight bytes, which is what lets peer-direct deployments safely validate ranges without a central authority. Large u64 values (GPU addresses) are serialized as strings to avoid JSON precision loss.
+
+The Rust and Python implementations of `compute_mx_source_id` are locked together via cross-checked pinned-hash unit tests (`source_identity.rs::test_python_cross_check_*` and `test_source_id.py::test_pinned_hash_*`). Either side drifting on canonical JSON encoding or hashing breaks both test sets together.
```
nicolasnoble: Updated. Wording now says the ID validates declared identity, not tensor bytes, and bit-identity holds when `revision` names immutable content.

examples/k8s_service_sources/sources-tp2.yaml
```diff
@@ -0,0 +1,213 @@
...
+    spec:
+      containers:
+        - name: vllm
+          image: nvcr.io/nvidian/dynamo-dev/modelexpress-client:latest
+          imagePullPolicy: IfNotPresent
+          securityContext:
+            capabilities:
+              add:
+                - IPC_LOCK
+          env:
```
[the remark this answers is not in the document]
nicolasnoble: Declined - these are examples demonstrating the k8s-service routing pattern, not production manifests. Security hardening is a deployer concern; matches the minimal style of the existing `p2p_transfer_k8s` examples.

examples/k8s_service_sources/target.yaml
```diff
@@ -0,0 +1,161 @@
...
+          ports:
+            - name: worker-grpc
+              containerPort: 6555
+          resources:
+            limits:
+              nvidia.com/gpu: 1
+              rdma/shared_ib: 1
+            requests:
+              nvidia.com/gpu: 1
+              rdma/shared_ib: 1
```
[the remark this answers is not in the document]
nicolasnoble: Declined - example scope. The bot's own note that `containerPort` is informational for most CNIs confirms it's not a reachability issue; NetworkPolicy / service-mesh setup is a deployer concern.

docs/ARCHITECTURE.md
```diff
@@ -352,15 +352,92 @@ Runs in a background tokio task on a configurable interval (default 1 hour). LRU
...
 
-Each worker publishes independently (no Lua merge scripts needed). The `mx_source_id` is a 16-char hex key computed from `SHA256(canonical_json(SourceIdentity))`. Large u64 values (GPU addresses) are serialized as strings to avoid JSON precision loss.
+The peer-direct backend lives in the Python client and is selected via `MX_METADATA_BACKEND`:
+
+- **K8s-service** (`k8s-service`/`service`): each source pool sits behind a Kubernetes Service (one per tensor-parallel rank, label selector pinned to `mx.rank=R`). Clients open a direct gRPC channel to the Service DNS and call `GetTensorManifest`; kube-proxy load-balances across ready backends. No central server is involved. `mx_source_id` is computed client-side via the same canonical JSON + SHA256 scheme and validated on the response. See [`../examples/k8s_service_sources/`](../examples/k8s_service_sources/) for the deployment shape.
```
nicolasnoble: Fixed. Two `MX_P2P_METADATA=1` mentions now note the decentralized-backend auto-force. ARCHITECTURE.md's env var table was behind DEPLOYMENT.md's - added `MX_MODEL_REVISION`, `MX_K8S_SERVICE_PATTERN`, `MX_K8S_SOURCE_RETRIES`, `MX_K8S_SOURCE_BACKOFF_SECONDS`, and broadened the `MX_METADATA_BACKEND` entry to cover client-side values.

modelexpress_client/python/modelexpress/source_id.py
```diff
@@ -0,0 +1,58 @@
...
+    canonical = _canonical_json(identity)
+    digest = hashlib.sha256(canonical.encode()).hexdigest()
+    return digest[:16]
+
+
+def _canonical_json(identity: p2p_pb2.SourceIdentity) -> str:
+    sorted_extra = {
+        k.lower(): v.lower()
+        for k, v in sorted(identity.extra_parameters.items())
+    }
```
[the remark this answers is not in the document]
nicolasnoble: Fixed on both sides with option 2 (deterministic normalize). Canonical JSON generation now sorts `extra_parameters` by original key first, lowercases, keeps the first value on collision. Rust's `HashMap` iteration non-determinism no longer leaks into the hash. Added cross-checked pinned-hash test for the case-colliding pair (`{"Foo": "a", "foo": "b"}` -> `bd9ea6c70d83fef1` on both sides).

modelexpress_client/python/modelexpress/metadata.py
nicolasnoble: Will open a separate PR for this.

zhengluo-nv (approved):
LGTM. Pending another review

PR #254, opened 2026-04-27
feat: add file selector to model file requests

andrewpaprotsky:
## Why

- Model file clients currently have to list or stream every file in a model directory, even when they only need a small known subset.
- Partial file requests are useful for callers that already know the model-root-relative paths they need and want to avoid unnecessary metadata and transfer work.

## What

- Adds `ModelFileSelector` to `ModelFilesRequest` with exact model-root-relative `paths`.
- Wires the optional selector through `ListModelFiles` and `StreamModelFiles`.
- Keeps existing behavior unchanged when `file_selector` is unset: all safe model files are listed or streamed.
- Updates generated Go protobuf bindings for the new request field.

## How

- Extends the model proto with `file_selector = 4` and a `ModelFileSelector` message.
- Filters collected files by exact relative path match after the existing path-safety checks.
- Passes `None` from the Rust client streaming path so existing callers remain compatible.
- Refreshes tests and generated code for the new request shape.

## Release Notes

* **New Features**
  * Added file filtering capabilities for model downloads. You can now specify particular model files to retrieve during streaming and listing operations, enabling more efficient and targeted downloads without transferring unnecessary files from the server.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/modelexpress/pull/254?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


modelexpress_server/src/services.rs
```diff
@@ -134,19 +138,24 @@ fn collect_model_files(base_path: &Path, current_path: &Path) -> Vec<(PathBuf, u
...
-                        if is_safe {
-                            files.push((relative.to_path_buf(), metadata.len()));
-                        } else {
+                        if !is_safe {
                             tracing::warn!(
                                 "Skipping potentially unsafe file path: {:?} (relative: {:?})",
                                 path,
                                 relative
                             );
+                        } else if file_selector.is_none_or(|selector| {
```
nv-hwoo, line 147: Full truth table
```
┌─────────┬───────────────┬─────────────────────────────┬─────────────────────────────────┐
│ is_safe │ file_selector │ selector contains relative? │             Action              │
├─────────┼───────────────┼─────────────────────────────┼─────────────────────────────────┤
│ false   │ (anything)    │ (anything)                  │ warn!, do not push              │
├─────────┼───────────────┼─────────────────────────────┼─────────────────────────────────┤
│ true    │ None          │ n/a                         │ push                            │
├─────────┼───────────────┼─────────────────────────────┼─────────────────────────────────┤
│ true    │ Some(_)       │ yes                         │ push                            │
├─────────┼───────────────┼─────────────────────────────┼─────────────────────────────────┤
│ true    │ Some(_)       │ no                          │ silently drop (no log, no push) │
└─────────┴───────────────┴─────────────────────────────┴─────────────────────────────────┘
```
The 4th row is the new branch: A client can request ["a.txt", "b.txt"], only a.txt exists, and the server silently returns just a.txt. There's no tracing::warn! or tracing::debug! on the dropped path, so it's also invisible in server logs.

Can we handle the last case mentioned here?

nv-hwoo (commented):
Overall looks good, just one comment.

modelexpress_server/src/services.rs
andrewpaprotsky: addressed in https://github.com/ai-dynamo/modelexpress/pull/254/commits/b314dc54d395f924ee88a4351811f2445242906c

missing selector paths now return NotFound instead of silently returning a partial file set

nv-hwoo (approved):

PR #255, opened 2026-04-29
feat: per-rank UCX_NET_DEVICES pinning via MX_RDMA_NIC_PIN

nicolasnoble:
Optional per-worker NIC pinning for NixlTransferManager. UCX's default NIC selection can land a worker's RDMA stream on a NIC sitting on a different NUMA node from its GPU, forcing traffic across CPU UPI / Infinity Fabric and capping single-flow bandwidth well below the local PIX path.

- MX_RDMA_NIC_PIN: "auto" / "1" / "true" runs the topology probe; a comma-separated NIC list is a legacy explicit map indexed by device_id; "off" / unset (default) leaves UCX selection alone.
- Probe scores each (gpu, nic) pair by PCIe sysfs realpath common- prefix depth (the affinity metric nvidia-smi topo -m uses).
- Greedy assignment in visible-index order; ties broken by fewest prior assignments then lex-smallest NIC name. Every rank computes the same global map from the same /sys snapshot.
- Filters bonded interfaces and rate < 200 Gb/s by default (configurable via MX_RDMA_NIC_PIN_MIN_RATE_GBPS).
- Cross-socket picks log at WARNING; INFO line includes common- depth, NUMA, NIC rate, and full GPU->NIC map.

## Summary by CodeRabbit

* **New Features**
  * Added automatic NIC pinning optimization for RDMA/UCX network performance. GPUs are intelligently matched to network interface cards based on PCIe affinity and topology.
  * Introduced `MX_RDMA_NIC_PIN` configuration option supporting three modes: disabled, explicit NIC assignment list, or automatic detection.
  * Network device assignments now persist for the worker lifetime, improving performance consistency.
  * Added enhanced cross-socket detection and detailed logging for troubleshooting.


modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -39,6 +39,337 @@ def is_nixl_available() -> bool:
...
+    the actual selection signal in probe_nic_pin_for_device(). NIC name
+    ordering only affects the final lex tiebreak.
+
+    NUMA is read from /sys/class/infiniband/<nic>/device/numa_node and
+    is kept on the tuple for diagnostic logging only; -1 if the kernel
+    reports unknown.
+    """
+    import os
+
+    base = "/sys/class/infiniband"
```
zhengluo-nv: Is this auto detection too specific to our dev cluster environment? What if no infiniband?

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -39,6 +39,337 @@ def is_nixl_available() -> bool:
...
+        logger.warning(
+            f"MX_RDMA_NIC_PIN explicit list: device_id {device_id} out of "
+            f"range for list of length {len(nic_list)}; skipping pin"
+        )
+        return None
+
+    try:
+        min_rate = float(os.environ.get("MX_RDMA_NIC_PIN_MIN_RATE_GBPS", "200"))
+    except ValueError:
+        min_rate = 200.0
```
zhengluo-nv: That feels also very specific to our dev environment. How do you know if customer's environment has >200 gbps NIC?

modelexpress_client/python/modelexpress/nixl_transfer.py
nicolasnoble: Probe surface is `/sys/class/infiniband`, the kernel verbs API. That covers IB, RoCE, OPA, and AWS EFA, not just our cluster. Fabrics outside the verbs API like Slingshot via CXI aren't visible; users on those clusters should leave MX_RDMA_NIC_PIN unset or set an explicit NIC list. No-IB case is silent: `_list_compute_ib_nics` returns [], probe returns None, UCX_NET_DEVICES untouched. Scope documented explicitly in 515038d.

modelexpress_client/python/modelexpress/nixl_transfer.py
nicolasnoble: Fair, the 200 Gb/s default was carrying assumptions from our dev environment. Replaced with a max-rate auto-filter in 515038d: keep only NICs at max(rate) over the discovered set. On a 100 Gb/s cluster all NICs are kept; on a mixed cluster the slower side-fabric NICs are stripped automatically. `MX_RDMA_NIC_PIN_MIN_RATE_GBPS` overrides with an explicit lower bound when a cluster intentionally has multiple rate tiers.

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -39,6 +39,388 @@ def is_nixl_available() -> bool:
...
+      - any other truthy value (e.g. "auto", "1", "true", "yes", "on"):
+        runs probe_nic_pin_for_device(). Rate filtering defaults to
+        max-rate auto-detect (keep only NICs at the fastest rate
+        present, strips slower side-fabric NICs without hardcoding a
+        number). MX_RDMA_NIC_PIN_MIN_RATE_GBPS overrides with an
+        explicit absolute lower bound when needed.
+    """
+    import os
+
+    raw = os.environ.get("MX_RDMA_NIC_PIN", "").strip()
```
zhengluo-nv: We should add this to all examples and maybe enable by default? Or put a commented out line

nicolasnoble: Done in af3fb44. Added MX_RDMA_NIC_PIN and MX_RDMA_NIC_PIN_MIN_RATE_GBPS to the env table in docs/DEPLOYMENT.md, plus a new "NIC Pinning (UCX Workaround)" subsection that names openucx/ucx#11259 as the upstream issue and notes the env var can be removed once a patched UCX is deployed. Short pointer in ARCHITECTURE.md too. Injected MX_RDMA_NIC_PIN=auto into 6 multi-rank P2P client manifests; left ModelStreamer-only single-node examples and pure-infra manifests untouched.

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -39,6 +39,388 @@ def is_nixl_available() -> bool:
     return NIXL_AVAILABLE
 
 
+def _read_int_file(path: str) -> int | None:
```
zhengluo-nv: Can we move all these workaround changes to a separate file like ucx_utils, becuase they only apply to UCX not other NIXL backend.

nicolasnoble: Done in 3f3a538. Moved the helpers to a new modelexpress/ucx_utils.py module. Public surface is two entry points: apply_nic_pin_for_device for the side-effecting integration in NixlTransferManager.initialize, and probe_nic_pin_for_device for diagnostics. nixl_transfer.py is now a single-line call into ucx_utils with no UCX-specific logic. No semantic change.

zhengluo-nv (approved):

PR #260, opened 2026-05-04
feat: Add opt-in OTEL tracing in vllm loader

emricksini-h:
## Description

Adds opt-in OpenTelemetry tracing to the Python client. When `opentelemetry-api` is installed (via the new [otel] extra), `LoadStrategyChain.run()` emits a "Load model" span with `model_name`, `global_rank`, `eligible_strategies`, and the winning `weight_loading_strategy`. When the package is not installed, a no-op tracer is used so there is zero runtime cost and no extra dependency for users who don't want it.

This allow to still have the "Load model" span when enabling tracing in vLLM (version > 0.16.0) when `--load-format mx` is enabled.

Also includes changes from #257.

## Release Notes

* **New Features**
  * Added OpenTelemetry tracing support for model-loading operations, capturing model name, rank, strategies, and weight-loading decisions.
  * Includes graceful fallback when OpenTelemetry is not installed.

* **Tests**
  * Added test coverage for tracing functionality to verify span emission and attributes.

* **Chores**
  * Added optional OpenTelemetry API and SDK dependencies.


KavinKrishnan (approved):

emricksini-h:
Hi @KavinKrishnan ! Thanks for the approval.
Unforntunaly I cannot merge it myself because of the failing checks.
For the Lint PR one, I think it could be resolved with the suggestion in #263 .

PR #261, opened 2026-05-04
feat(model-streamer): distributed streaming and S3 URI support for --load-format mx

noa-neria:
# feat(model-streamer): distributed streaming and S3 URI support for `--load-format mx`

## Summary

- **`MX_MS_DISTRIBUTED`**: New environment variable that enables distributed safetensors
  streaming via the model streamer load strategy when using tensor parallelism. When set
  to `1`, each worker streams and broadcasts a unique part of the model to the other processes via GPU-to-GPU communication, rather than each worker downloading the full checkpoint. Requires `tensor_parallel_size > 1`
  and a CUDA-capable platform; defaults to `0` (disabled) similar to the upstream `distributed` flag.

- **S3/GCS URI support with `--load-format mx`**: vLLM rejects `--load-format mx` when
  the `--model` argument is an object storage URI (`s3://`, `gs://`), because it only
  permits `runai_streamer` or `runai_streamer_sharded` in that path. A lightweight
  monkey-patch on `VllmConfig.try_verify_and_update_config` temporarily swaps the format
  to `runai_streamer` during validation, then restores `mx` so `MxModelLoader` handles
  the actual loading. Note: `MX_MODEL_URI` must still be set to trigger the model
  streamer load strategy in the current design.

## Tests

Unit tests added in `test_model_streamer_strategy.py` covering all branches of the
distributed streaming logic: default-off behavior, activation via `MX_MS_DISTRIBUTED=1`,
no-op when `tensor_parallel_size=1`, no-op on non-CUDA platforms, and correct
`cuda:{device_id}` assignment per worker.

## Testing

Validated end-to-end on an AWS g5.12xlarge (4× A10G) with `vllm==0.20.1`:

```bash
# Distributed streaming with TP=2 from S3
MX_MODEL_URI=s3://core-llm/Llama-3-8b/ MX_MS_DISTRIBUTED=1 \
  vllm serve s3://core-llm/Llama-3-8b/ \
  --tensor-parallel-size 2 --enforce-eager --load-format mx
```

Log confirms 2-way distributed streaming:
```
[Worker 0] Streaming: 291/291 tensors (100%) in 8s   # 7.5 GiB / worker
[Worker 0] MxModelLoader.load_model() COMPLETE in 8.90s
[Worker 1] MxModelLoader.load_model() COMPLETE in 8.90s
Application startup complete.


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Added `MX_MS_DISTRIBUTED` configuration option to enable distributed weight streaming directly to each worker's GPU on multi-GPU systems.

* **Documentation**
  * Documented `MX_MS_DISTRIBUTED` option requirements: tensor parallelism > 1 and CUDA-capable platform.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -398,6 +398,7 @@ All storage backends (S3, GCS, Azure) are included as core dependencies — no e
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `MX_MODEL_URI` | (none) | Model location. Must be set to enable ModelStreamer. Accepts: remote URI (`s3://bucket/model`, `gs://...`, `az://...`), absolute local path (`/models/deepseek-ai/DeepSeek-V3`), or HuggingFace model ID (`deepseek-ai/DeepSeek-V3` — resolved via `HF_HUB_CACHE`). |
+| `MX_MS_DISTRIBUTED` | `0` | Enable distributed streaming (streams directly to each worker's GPU). Requires tensor parallelism > 1 and a CUDA-capable platform. Set to `1` to activate. |
 | `RUNAI_STREAMER_CONCURRENCY` | `8` | Number of concurrent read threads |
 | `RUNAI_STREAMER_MEMORY_LIMIT` | (none) | CPU staging buffer size in bytes. `0` reuses a single-tensor buffer (most memory efficient). See [runai-model-streamer docs](https://github.com/run-ai/model-streamer). |
 
--- a/examples/p2p_transfer_k8s/client/vllm/vllm-single-node-streamer-local.yaml
+++ b/examples/p2p_transfer_k8s/client/vllm/vllm-single-node-streamer-local.yaml
@@ -4,7 +4,8 @@
 # Single-node vLLM deployment with ModelExpress ModelStreamer (local disk).
 # Streams weights from PVC via runai-model-streamer with concurrent I/O.
 #
-# MX_MODEL_URI accepts:
+# MX_MODEL_URI activates the streamer; its value is passed to vllm via --model.
+# Accepts:
 #   - HF model ID: "deepseek-ai/DeepSeek-V3" (resolved via HF_HUB_CACHE)
 #   - Absolute path: "/models/my-model"
 #   - Remote URI: "s3://bucket/model" (see vllm-single-node-streamer-s3.yaml)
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: modelexpress_client/python/modelexpress/vllm_loader.py [hunk 1]
```diff
@@ -48,6 +48,52 @@
 _nixl_managers: dict[int, NixlTransferManager] = {}
 
 
+def _patch_vllm_s3_format_check() -> None:
```
zhengluo-nv: nit: Can we move this into the model_streamer_streategy to keep vllm_loader simple?

zhengluo-nv (approved):

noa-neria:
@zhengluo-nv I have minimized the vllm patch and moved the code to the streamer strategy file, as discussed. Note that the patch must be applied in the vllm loader, since the streamer load strategy is imported later after vllm reached the method we need to patch.

Also, as we discussed the URI source is now the model passed to vllm with `--model`. This ensures that the model config and weights are both loaded from the URI. 

To summarize - three tightly coupled changes on the `mx` load path:                                                                                                                                                                                                                                                                                                                              
   
  - **Distributed streaming.** `MX_MS_DISTRIBUTED=1` enables direct safetensors streaming to each worker's GPU (`cuda:N`) via `runai-model-streamer`. Active when tensor-parallelism > 1 on CUDA-alike platforms; off by default.                                                                                                                                                   
  - **`mx` works with object-storage URIs end-to-end.** A small shim on `VllmConfig.try_verify_and_update_config` temporarily detaches `model_config.model_weights` for the duration of the call, so vllm's S3 guard no longer rejects `load_format=mx`. The patch lives next to `ModelStreamerStrategy` and is installed eagerly from `engines/vllm/loader.py`.
  - **URI sourced from vllm config, not `MX_MODEL_URI`.** The streamer reads its URI from `model_config.model_weights` (falling back to `model_config.model`), matching vllm's `runai_streamer_loader`. `MX_MODEL_URI` remains as the activation gate; the streamer YAMLs now pass the URI to vllm via `--model $(MX_MODEL_URI)` and `MODEL_NAME` is dropped from the s3 example.   
                                                                                                                                                                                                                                                                                                                                                                                    
  After these changes, both the model config and the safetensors flow from the same object-storage URI, set once on `--model`.

zhengluo-nv (approved):

PR #262, opened 2026-05-05
refactor: organize engine and metadata loading paths

zhengluo-nv:
Refactors the Python client loading path so engine-specific code sits behind a small adapter boundary, while preserving the existing vLLM entry point used by current manifests.

Design doc: https://docs.google.com/document/d/1MlvAx_evyI75a1TkOIlKR3UV7-ESjw2R1WJNBYrWum8/edit?tab=t.5fveasr8cwdc#heading=h.7rw11mbt6qjl

## Why

ModelExpress now has more than one loading path and more than one engine integration target:

- RDMA P2P from another live worker
- ModelStreamer from object storage / local disk
- GDS / native fallback
- vLLM today, with SGLang and TRT-LLM integration work following the same pattern

Before this PR, vLLM-specific logic, metadata publication, rank/device identity, and strategy selection were coupled together. That made it hard to reason about which behavior was engine-neutral and which behavior belonged to vLLM.

This PR separates those concerns without changing the public vLLM loader import path.

Configurable chain, @register_strategy + entry-points, MxConfig are new features and they are for follow up PRs. 

## Before / after

| Aspect | Before | After |
|---|---|---|
| Strategy result contract | Strategies returned engine-specific objects or booleans directly | Strategies pass a `LoadResult` through the chain |
| Engine coupling | Strategies called vLLM-shaped behavior directly | Strategies call `EngineAdapter` hooks |
| vLLM integration code | Mixed into top-level `modelexpress` modules | Implemented under `modelexpress.engines.vllm` |
| Existing vLLM entry point | `modelexpress.vllm_loader` | Still available as the compatibility entry point |
| Metadata helpers | Top-level modules mixed with loading code | Moved under `modelexpress.metadata` |
| File renames | Broad package rename was considered | `load_strategy/` is intentionally kept as-is |
| P2P rank identity | Some paths used global rank where worker rank is required | Adapter exposes engine-specific worker/global rank mapping |

## What changed

**Engine adapter boundary:** added `modelexpress.adapter.EngineAdapter`, `LoadResult`, and adapter hooks for weight iterators, native load, tensor discovery, rank/device identity, lifecycle hooks, and retry reinitialization.

**vLLM adapter:** moved vLLM-specific implementation under `modelexpress.engines.vllm`, including the adapter and loader implementation. The existing `modelexpress.vllm_loader` module remains as the compatibility import path.

**Load strategy chain:** updated RDMA, ModelStreamer, GDS, and default strategies to operate through `LoadResult` and adapter hooks instead of directly assuming vLLM internals.

**Metadata package:** moved metadata publication, heartbeat, client factory, K8s service client, worker server, and source identity code into `modelexpress.metadata` without keeping top-level shim modules.

**Docs and examples:** updated architecture/deployment docs and examples to reflect the current package boundaries and required RBAC/example layout.

## Compatibility notes

- Existing vLLM users should continue using `--load-format mx` and the existing `modelexpress.vllm_loader` entry point.
- `load_strategy/` is not renamed in this PR.
- Top-level metadata helper modules were moved; internal imports now use `modelexpress.metadata.*`.

### Unit tests

Focused Python tests were run for the strategy chain and ModelStreamer behavior:

```bash
UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest \
  tests/test_model_streamer_strategy.py \
  tests/test_vllm_loader.py::TestLoadStrategyChainRunErrorHandling
```

Result: **26 passed**.

### nscale Dynamo DGD end-to-end

Built branch images and deployed the Dynamo aggregated vLLM example with TP=4 / PP=1. Scaled `VllmWorker` from 1 to 2 replicas.

| Check | Result |
|---|---|
| First replica startup | Ready |
| Second replica startup | Ready |
| Second replica weight path | RDMA P2P |
| Transfer size | 149.59 GB per rank |
| Transfer time | ~5.6s |
| Throughput | ~211-212 Gbps |
| Frontend inference | HTTP 200 responses; traffic observed on both replicas |

### AWS ModelStreamer S3 validation

Built and pushed the linux/arm64 client image for AWS GB200:

```bash
nvcr.io/nvidian/dynamo-dev/model-express-dev-containers:vllm-streamer-29ae4a6
```

Deployed `examples/p2p_transfer_k8s/client/vllm/vllm-single-node-streamer-s3.yaml` as a temporary manifest with:

```bash
MX_MODEL_URI=s3://model-express-models/deepseek-ai/DeepSeek-V3/
AWS_DEFAULT_REGION=us-east-1
```

Result: ModelStreamer successfully streamed all 163 safetensors files.

| Check | Result |
|---|---|
| S3 object discovery | 163 safetensors files |
| Streamed bytes | ~641.3 GiB per rank |
| Stream time | ~356-376s |
| ModelStreamer completion | `Model streamer weight loading complete` on all ranks |

The AWS test also found a deployment pitfall now documented in the local test skill: the `model-express-models` bucket is in `us-east-1`. With `AWS_DEFAULT_REGION=us-west-2`, boto3 `HeadObject` and range reads can still work, but RunAI libstreamer fails with `File access error`.

The same AWS TP=4 smoke test then hit a vLLM serving configuration limit after successful weight load: DeepSeek-V3 default max length `163840` required more KV cache than available (`10.72 GiB` needed vs `4.26 GiB` available). For TP=4 smoke inference, use a lower `--max-model-len`, for example `32768`.

## Follow-ups

- Add engine adapters for SGLang and TRT-LLM on top of this boundary.
- Add automatic UCX device selection instead of requiring manual per-rank device selection.
- Add a dedicated ModelStreamer S3 smoke recipe for AWS that sets the bucket region and TP=4 serving parameters explicitly.

## Release Notes

* **New Features**
  * Enhanced engine adapter system enabling flexible, engine-specific loading strategies with improved retry and rollback capabilities.
  * KV transfer connector configuration for vLLM decode workers in Kubernetes deployments.

* **Documentation**
  * Updated architecture documentation reflecting reorganized module structure.

* **Chores**
  * Expanded Kubernetes RBAC permissions for model cache entry management.
  * Restructured internal modules for better API organization and maintainability.

### DIFF
--- a/examples/dynamo_p2p_transfer_k8s/vllm/vllm-single-node-disaggregated.yaml
+++ b/examples/dynamo_p2p_transfer_k8s/vllm/vllm-single-node-disaggregated.yaml
@@ -119,13 +119,6 @@ spec:
         - name: UCX_RNDV_THRESH
           value: "0"
       extraPodSpec:
-        tolerations:
-          - key: nvidia.com/gpu
-            operator: Exists
-            effect: NoSchedule
-          - key: dra
-            operator: Exists
-            effect: NoSchedule
         # Keep decode replicas apart, but allow one decode and one prefill
         # worker to share an 8-GPU node so all GPUs can be utilized.
         affinity:
@@ -157,6 +150,8 @@ spec:
             - --trust-remote-code
             - --disaggregation-mode
             - decode
+            - --kv-transfer-config
+            - '{"kv_connector":"NixlConnector","kv_role":"kv_both"}'
       volumeMounts:
         - name: shared-model-cache
           mountPoint: /models
@@ -199,13 +194,6 @@ spec:
         - name: UCX_RNDV_THRESH
           value: "0"
       extraPodSpec:
-        tolerations:
-          - key: nvidia.com/gpu
-            operator: Exists
-            effect: NoSchedule
-          - key: dra
-            operator: Exists
-            effect: NoSchedule
         # Keep prefill replicas apart, but allow one decode and one prefill
         # worker to share an 8-GPU node so all GPUs can be utilized.
         affinity:
--- a/examples/dynamo_p2p_transfer_k8s/vllm/vllm-single-node-disaggregated.yaml
+++ b/examples/dynamo_p2p_transfer_k8s/vllm/vllm-single-node-disaggregated.yaml
@@ -150,8 +150,6 @@ spec:
             - --trust-remote-code
             - --disaggregation-mode
             - decode
-            - --kv-transfer-config
-            - '{"kv_connector":"NixlConnector","kv_role":"kv_both"}'
       volumeMounts:
         - name: shared-model-cache
           mountPoint: /models
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -145,6 +145,13 @@ class VllmAdapter(EngineAdapter):
 
     def _reset_compilation_state(self) -> None:
         compilation_config = self.vllm_config.compilation_config
+        # vLLM registers each attention / MLA / Mamba / FusedMoE layer into
+        # fields on vllm_config.compilation_config during initialize_model().
+        # Those fields live on the config object, not the model, so they survive
+        # del model and trip duplicate registration on the next initialize_model().
+        # Clear them so re-init starts from a clean slate. Audited against vLLM
+        # 0.17.1; other versions may add init=False fields that need similar
+        # treatment.
         compilation_config.static_forward_context.clear()
         compilation_config.static_all_moe_layers.clear()
         compilation_config.enabled_custom_ops.clear()
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -1,0 +1,171 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""vLLM implementation of the ModelExpress engine adapter contract."""
+
+from __future__ import annotations
+
+import copy
+import logging
+import uuid
+from typing import Iterator
+
+import torch
+
+from ...adapter import EngineAdapter
+from ...load_strategy.context import LoadContext, LoadResult
+from ...metadata.client_factory import create_metadata_client
+from ...metadata.publish import build_source_identity
+from ...rank_utils import get_global_rank, get_worker_rank
+from ...tensor_utils import adopt_hidden_tensors, capture_tensor_attrs, collect_module_tensors
+
+logger = logging.getLogger("modelexpress.engines.vllm.adapter")
+
+
+class VllmAdapter(EngineAdapter):
+    """Adapter that maps strategy hooks onto vLLM's native loader APIs."""
+
+    def __init__(self, vllm_config, model_config):
+        self.vllm_config = vllm_config
+        self.model_config = model_config
+        self.load_config = vllm_config.load_config
+        self.target_device = self._resolve_target_device()
+
+    def build_identity(self):
+        return build_source_identity(self.vllm_config, self.model_config)
+
+    def get_worker_rank(self) -> int:
+        return get_worker_rank(self.target_device)
+
+    def get_global_rank(self) -> int:
+        return get_global_rank(self.target_device)
+
+    def get_device_id(self) -> int:
+        return get_worker_rank(self.target_device)
+
+    def get_target_device(self) -> torch.device:
+        return self.target_device
+
+    def discover_tensors(self, result: LoadResult) -> dict[str, torch.Tensor]:
+        if result.model is None:
+            raise RuntimeError("vLLM tensor discovery requires result.model")
+        adopt_hidden_tensors(result.model)
+        return collect_module_tensors(result.model)
+
+    def prepare_rdma_target(self, result: LoadResult) -> LoadResult:
+        if result.model is None:
+            raise RuntimeError("vLLM RDMA target preparation requires result.model")
+
+        from vllm.model_executor.model_loader.dummy_loader import DummyModelLoader
+
+        dummy_config = copy.copy(self.load_config)
+        try:
+            dummy_config.load_format = "dummy"
+        except AttributeError:
+            object.__setattr__(dummy_config, "load_format", "dummy")
+        DummyModelLoader(dummy_config).load_weights(result.model, self.model_config)
+        return result
+
+    def before_rdma_receive(self, result: LoadResult) -> LoadResult:
+        return self._process_weights_after_loading(result)
+
+    def apply_weight_iter(
+        self,
+        result: LoadResult,
+        weights_iter: Iterator[tuple[str, torch.Tensor]],
+    ) -> LoadResult:
+        if result.model is None:
+            raise RuntimeError("vLLM weight iterator loading requires result.model")
+        result.model.load_weights(weights_iter)
+        return result
+
+    def load_via_native(self, result: LoadResult) -> LoadResult:
+        if result.model is None:
+            raise RuntimeError("vLLM native loading requires result.model")
+
+        from vllm.model_executor.model_loader.default_loader import DefaultModelLoader
+
+        disk_config = copy.copy(self.load_config)
+        try:
+            disk_config.load_format = "auto"
+        except AttributeError:
+            object.__setattr__(disk_config, "load_format", "auto")
+
+        DefaultModelLoader(disk_config).load_weights(result.model, self.model_config)
+        return result
+
+    def after_weight_iter_load(self, result: LoadResult) -> LoadResult:
+        return self._process_weights_after_loading(result)
+
+    def after_native_load(self, result: LoadResult) -> LoadResult:
+        return self._process_weights_after_loading(result)
+
+    def reinit_for_retry(self, result: LoadResult) -> LoadResult:
+        from vllm.model_executor.model_loader.utils import initialize_model
+
+        old_value = result.value
+        result.value = None
+        result.model = None
+        del old_value
+        torch.cuda.empty_cache()
+        self._reset_compilation_state()
+        logger.info(
+            "[Worker %s] Re-initializing vLLM model after failed strategy",
+            self.get_global_rank(),
+        )
+        with self.target_device:
+            model = initialize_model(
+                vllm_config=self.vllm_config,
+                model_config=self.model_config,
+            )
+        return LoadResult(value=model, model=model, publishable=result.publishable)
+
+    def _process_weights_after_loading(self, result: LoadResult) -> LoadResult:
+        if result.model is None:
+            raise RuntimeError("vLLM post-load processing requires result.model")
+
+        from vllm.model_executor.model_loader.utils import process_weights_after_loading
+
+        with capture_tensor_attrs():
+            process_weights_after_loading(
+                result.model, self.model_config, self.target_device,
+            )
+        return result
+
+    def _resolve_target_device(self) -> torch.device:
+        load_device = (
+            self.vllm_config.device_config.device
+            if self.load_config.device is None
+            else self.load_config.device
+        )
+        return torch.device(load_device)
+
+    def _reset_compilation_state(self) -> None:
+        compilation_config = self.vllm_config.compilation_config
+        compilation_config.static_forward_context.clear()
+        compilation_config.static_all_moe_layers.clear()
+        compilation_config.enabled_custom_ops.clear()
+        compilation_config.disabled_custom_ops.clear()
+        compilation_config.traced_files.clear()
+        compilation_config.compilation_time = 0.0
+
+
+def build_vllm_load_context(vllm_config, model_config) -> LoadContext:
+    """Build a LoadContext from vLLM config objects."""
+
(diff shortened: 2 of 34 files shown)


examples/dynamo_p2p_transfer_k8s/vllm/vllm-single-node-disaggregated.yaml
```diff
@@ -161,6 +154,8 @@ spec:
             - --trust-remote-code
             - --disaggregation-mode
             - decode
+            - --kv-transfer-config
+            - '{"kv_connector":"NixlConnector","kv_role":"kv_both"}'
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in `bcea35b`.

I removed the decode-side `--kv-transfer-config` and value so the decode worker now relies on Dynamo automatic `NixlConnector` injection from `--disaggregation-mode decode`. I also verified the prefill worker still keeps the explicit `--kv-transfer-config` for `kv_both`, since prefill still needs to declare the connector config explicitly in this manifest.

Validation covered by YAML validation in `pre-commit run`.

modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -0,0 +1,131 @@
...
+                model = initialize_model(
+                    vllm_config=vllm_config, model_config=model_config
+                )
+
+            model = LoadStrategyChain.run(model, ctx)
+
+            # Update global registries
+            _tensor_registry[ctx.device_id] = ctx.tensors
+            if ctx.nixl_manager is not None:
+                _nixl_managers[ctx.device_id] = ctx.nixl_manager
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in `bcea35b`.

The vLLM loader now updates `_tensor_registry[ctx.device_id]` and then either stores the active `ctx.nixl_manager` or clears the old entry with `_nixl_managers.pop(ctx.device_id, None)` when the current load does not create a manager. That prevents a later non-P2P load on the same device from inheriting a stale compatibility manager.

I also added `test_load_model_clears_stale_nixl_manager_when_none_registered` to cover this path.

modelexpress_client/python/modelexpress/load_strategy/gds_strategy.py
```diff
@@ -7,10 +7,9 @@
 
 import logging
 
-import torch.nn as nn
-
-from .base import LoadContext, LoadStrategy, register_tensors, publish_metadata
-from ..tensor_utils import capture_tensor_attrs
+from ..adapter import EngineAdapter
+from .base import LoadContext, LoadStrategy, _as_load_result, register_tensors
+from .context import LoadResult
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in `bcea35b`.

I changed this through the new strategy failure contract rather than returning `False` or a rollback-capable result. `GdsStrategy.load()` now raises `StrategyFailed(mutated=False)` when `MxGdsLoader.load_iter()` fails before model weights are applied, and raises `StrategyFailed(mutated=True)` when `ctx.adapter.apply_weight_iter()` fails after weight application may have started.

The chain now treats `mutated=True` as the signal to rollback strategy-owned state and reinitialize the model before trying the next strategy, so the fallback strategy is not run against partially applied weights.

Added coverage in `test_gds_apply_weight_iter_failure_is_mutated`.

modelexpress_client/python/modelexpress/load_strategy/model_streamer_strategy.py
```diff
@@ -17,10 +17,10 @@
 from typing import Iterator
 
 import torch
-import torch.nn as nn
 
-from .base import LoadContext, LoadStrategy, register_tensors, publish_metadata
-from ..tensor_utils import capture_tensor_attrs
+from ..adapter import EngineAdapter
+from .base import LoadContext, LoadStrategy, _as_load_result, register_tensors
+from .context import LoadResult
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in `bcea35b`.

ModelStreamer now follows the same clean-vs-mutated failure split as GDS. `_stream_weights()` failures are reported as `StrategyFailed(mutated=False)` because no model mutation has started yet. `ctx.adapter.apply_weight_iter()` failures are reported as `StrategyFailed(mutated=True)` because the iterator may already have partially updated the model.

With `mutated=True`, `LoadStrategyChain` rolls back strategy-owned state and asks the adapter to reinitialize the model before trying the next strategy.

Added coverage in `test_apply_weight_iter_failure_is_mutated`.

zhengluo-nv:
Addressed the remaining CodeRabbit outside/nitpick notes in `bcea35b`.

Details:

- RDMA rollback now shuts down the active NIXL manager before clearing strategy context. `RdmaStrategy.rollback()` calls `ctx.nixl_manager.shutdown()` when present, logs shutdown failures without masking fallback handling, then clears `ctx.tensors` and `ctx.nixl_manager`. Added `test_rollback_shuts_down_nixl_manager`.

- Added `EngineAdapter.get_global_rank()` to the adapter contract and kept the vLLM implementation returning the engine global rank. This makes the calls in vLLM retry/context setup type-safe through the shared adapter interface. Added decorator/default capability coverage for `get_global_rank`.

- Kept the unexpected-exception fallback behavior intentionally. Expected strategy misses now use `StrategyFailed`, including the `mutated` flag. Truly unexpected exceptions should be rare, but the chain still rolls back strategy-owned state and tries the next strategy to avoid crashing the engine during load fallback. I added a comment in `LoadStrategyChain.run()` to make that policy explicit.

- The rollback boolean path was replaced rather than moved. `LoadStrategy.rollback(ctx)` is now cleanup-only, and `StrategyFailed(mutated=True)` is the explicit signal to reinitialize the model before trying the next strategy. This avoids mixing cleanup state with retry-safety state.

Validation:

- `uv run --project modelexpress_client/python --extra dev pytest modelexpress_client/python/tests/` passed: 202 passed, 18 skipped.
- `pre-commit run` passed.

nicolasnoble (commented):
Inline comments cover the specific spots. The after_weight_iter_load mutation gap is the riskiest of the bunch and would justify another rev before merge.

Also in general: comparing this to the latest design doc, am I right that this is the Phase 1 slice and that the configurable chain, @register_strategy + entry-points, MxConfig, and engine-import lint are deferred to a follow-up? If so, a one-liner in the PR description noting that would help the SGLang and TRT-LLM follow-ups know what's available.

modelexpress_client/python/modelexpress/load_strategy/gds_strategy.py
```diff
@@ -19,39 +18,47 @@ class GdsStrategy(LoadStrategy):
...
         finally:
             gds_loader.shutdown()
 
-        with capture_tensor_attrs():
-            process_weights_after_loading(model, ctx.model_config, ctx.target_device)
-
-        register_tensors(model, ctx)
-        publish_metadata(ctx)
-        return True
+        result = ctx.adapter.after_weight_iter_load(result)
```
nicolasnoble: Should this be inside the same StrategyFailed(mutated=True) try as apply_weight_iter? after_weight_iter_load runs process_weights_after_loading on the vLLM side, which can mutate model state and fail. Right now an exception here escapes to the chain's generic except, which doesn't reinit, so the next strategy gets a partially post-processed model. Same shape applies in model_streamer_strategy.py:114 and default_strategy.py:32. Default's case is academic since nothing falls through it, but the inconsistency is worth fixing for symmetry.

modelexpress_client/python/modelexpress/load_strategy/base.py
```diff
@@ -214,7 +171,7 @@ def publish_metadata(ctx: LoadContext) -> None:
     try:
         publish_metadata_and_ready(
             ctx.mx_client, ctx.nixl_manager, ctx.tensors,
-            ctx.global_rank, ctx.device_id, ctx.identity, ctx.worker_id,
+            ctx.worker_rank, ctx.device_id, ctx.identity, ctx.worker_id,
```
nicolasnoble, line 171: This passes ctx.worker_rank, but the parameter on the receiving end of publish_metadata_and_ready is named global_rank, and inside that function it's used to key _heartbeat_threads[global_rank] = heartbeat. unpublish_metadata a few lines below then pops by ctx.global_rank, so when worker_rank != global_rank with DP > 1, the pop misses and the heartbeat keeps running, marking invalidated memory READY after teardown. Want to rename the parameter to worker_rank and key the lifecycle dict consistently? The wire field already gets the right value, only the local key is wrong.

modelexpress_client/python/modelexpress/load_strategy/context.py
```diff
@@ -0,0 +1,61 @@
...
+    mx_client: MxClientBase
+    worker_id: str
+    adapter: EngineAdapter | None = None
+    worker_rank: int | None = None
+    nixl_manager: NixlTransferManager | None = None
+    tensors: dict[str, torch.Tensor] = field(default_factory=dict)
+
+    def __post_init__(self) -> None:
+        if self.worker_rank is None:
+            self.worker_rank = self.global_rank
```
nicolasnoble: Any reason to keep this fallback? Production vLLM goes through build_vllm_load_context and sets worker_rank explicitly, so this branch is only reachable via direct LoadContext construction. The risk is that the next adapter or a legacy test path forgets to set worker_rank and silently re-conflates the two ranks, which is exactly what 05205ad was fixing. Maybe assert worker_rank is not None when adapter is set, or default to adapter.get_worker_rank() when adapter is present?

modelexpress_client/python/modelexpress/load_strategy/base.py
```diff
@@ -168,10 +116,19 @@ def register_tensors(model: nn.Module, ctx: LoadContext) -> None:
...
+        if result.model is None:
+            logger.info(
+                f"[Worker {ctx.global_rank}] No model available, skipping NIXL registration"
+            )
+            return
+
+        if ctx.adapter is not None:
+            ctx.tensors = ctx.adapter.discover_tensors(result)
+        else:
+            # Compatibility path for tests and legacy callers.
```
nicolasnoble: Is there an actual legacy caller for this path, or is it tests-only? If tests are the only user, dropping the branch and having them provide a fake adapter would simplify register_tensors and lock in the adapter-required invariant.

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def _resolve_target_device(self) -> torch.device:
+        load_device = (
+            self.vllm_config.device_config.device
+            if self.load_config.device is None
+            else self.load_config.device
+        )
+        return torch.device(load_device)
+
+    def _reset_compilation_state(self) -> None:
```
nicolasnoble, line 146: The original docstring on _reset_vllm_compilation_state called out that newer vLLM versions may add init=False fields on CompilationConfig that need similar treatment. The concern carries over here. Worth snapshotting the relevant fields at first init and diff-restoring on reset, or filing a vLLM upstream issue for a clean reset API? Not blocking, but the explicit field list is foreseeable upgrade breakage.

modelexpress_client/python/modelexpress/load_strategy/gds_strategy.py
zhengluo-nv: Good catch. Will wrapped after_weight_iter_load() / after_native_load() in mutated failure handling for GDS, ModelStreamer, and Default strategy.

modelexpress_client/python/modelexpress/load_strategy/base.py
zhengluo-nv: In this PR I fixed the worker rank so it should work with DP > 1. This is the legacy code and I did not fix it. Will fix in the next revision

modelexpress_client/python/modelexpress/load_strategy/context.py
zhengluo-nv: Agent likes to introduce bad fallback logic. Good catch

modelexpress_client/python/modelexpress/load_strategy/base.py
zhengluo-nv: Dropped

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
zhengluo-nv: To fully snapshot and restore this CompilationConfig seems not trivial, because the old config object may be referred by other components in vLLM. So I will add back the original warning docstring and keep this a separate PR

zhengluo-nv:
Configurable chain, @register_strategy + entry-points, MxConfig are new features and they are for follow up PRs. This PR is refactor only and is already large.

Order of execution in my mind: 
1. vLLM integration refactor (This PR)
2. SGLang integration unification (next PR)
3. TRTLLM integration unification (depending on upstream TRTLLM integration PR)
4. Configurable chain
5. MxConfig refactor

nicolasnoble (approved):
All findings addressed cleanly. Approving.

One minor thing for a follow-up if you want: the log-line prefixes inside publish_metadata_and_ready switched from [Worker {global_rank}] to [Worker {worker_rank}] alongside the heartbeat keying fix. The keying change is correct, but with DP > 1, worker_rank is shared across replicas, so those specific log lines no longer uniquely identify a replica while the rest of the codebase still prefixes with global_rank. Could keep the prefix on global_rank and just key the dict by worker_rank, or accept the minor logging-attribution loss. Not a blocker.

zhengluo-nv:
DP>1 running on the same node is a problem today, I have been testing it and triggered some other bugs like  https://github.com/ai-dynamo/modelexpress/issues/266

Noted on the logging behavior change when DP > 1. Need some more refactor to fix this in a clean way. Will do as follow up

PR #269, opened 2026-05-07
fix: resolve vLLM worker device id

zhengluo-nv:
## Overview

PR #262 changed the vLLM device-id lookup to read `device.index`. In vLLM TP workers, `vllm_config.device_config.device` is shared and usually has no index, so every worker falls back to `device_id=0`. That breaks multi-GPU CUDA/NIXL paths because all workers register as if they are on the same local device.

This PR resolves the local CUDA ordinal through vLLM's platform API, `current_platform.current_device()`, so each TP worker reports the device vLLM actually assigned. It also removes the stale generic `get_device_id()` helper because that fallback was only safe for indexed `torch.device` values and is not correct for the vLLM adapter.

## Validation

- `uv run --project modelexpress_client/python --extra dev pytest modelexpress_client/python/tests/test_vllm_adapter.py`
- `pre-commit run`
- Built and pushed client image `nvcr.io/nvidian/dynamo-dev/model-express-dev-containers:client-788a3e5`
- Kubernetes e2e with vLLM TP=2 and `--load-format mx`: worker logs showed device IDs `0` and `1` from `current_platform.current_device()`, model metadata reached Ready for both worker ranks, and `/v1/chat/completions` succeeded
- Full Python suite was attempted; remaining local failures were macOS `cuda.bindings` import gaps in pool-registration tests, not this vLLM adapter path

## Summary by CodeRabbit

* **Refactor**
  * Enhanced device detection in the vLLM engine adapter for more reliable CUDA device identification during model inference operations.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -16,7 +16,7 @@ from ...adapter import EngineAdapter
 from ...load_strategy.context import LoadContext, LoadResult
 from ...metadata.client_factory import create_metadata_client
 from ...metadata.publish import build_source_identity
-from ...rank_utils import get_device_id, get_global_rank
+from ...rank_utils import get_global_rank
 from ...tensor_utils import adopt_hidden_tensors, capture_tensor_attrs, collect_module_tensors
 
 logger = logging.getLogger("modelexpress.engines.vllm.adapter")
@@ -44,7 +44,7 @@ class VllmAdapter(EngineAdapter):
         return get_global_rank(self.target_device)
 
     def get_device_id(self) -> int:
-        return get_device_id(self.target_device)
+        return _get_vllm_device_id()
 
     def get_target_device(self) -> torch.device:
         return self.target_device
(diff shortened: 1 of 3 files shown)


KavinKrishnan (approved):

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
```diff
@@ -181,6 +181,15 @@ def _get_vllm_worker_rank(vllm_config: VllmConfig) -> int:
     return worker_rank
 
 
+def _get_vllm_device_id() -> int:
+    """Return the local CUDA ordinal vLLM assigned to this worker."""
+    from vllm.platforms import current_platform
+
+    device_id = int(current_platform.current_device())
```
nicolasnoble, line 188: Worth keeping device_id and target_device in sync here. Two cases:

- load_config.device is None: target_device ends up as bare `torch.device("cuda")` and device_id is correct from current_platform. Harmless today because the `with target_device:` blocks and `process_weights_after_loading` follow `torch.cuda.current_device()` for bare cuda, and `get_global_rank`'s `device.index` fallback only fires when torch.distributed isn't initialized.
- load_config.device is set to cuda:N: target_device carries the explicit index, but device_id is whatever `current_platform.current_device()` returns. Pre-#262, the old `get_device_id(target_device)` returned the explicit index and propagated it to NIXL agent name and gRPC port, so this is a behavior shift on the explicit-device path.

Could `_get_vllm_device_id` take target_device as a hint, returning `target_device.index` when present and falling back to `current_platform.current_device()` otherwise? Plus a `build_vllm_load_context` test asserting device_id and target_device.index agree on both paths.

zhengluo-nv: Follow up the fix in https://github.com/ai-dynamo/modelexpress/pull/270

PR #279, opened 2026-05-14
feat(ci): add K8s P2P CI tests for vLLM and TRT-LLM

tanushriya910:
## Summary

Adds a K8s-based CI for ModelExpress P2P weight transfer. Spins up a per-framework
namespace with an mx-server + source/target worker Jobs, runs a pytest that asserts
the RDMA transfer happened and both pods serve `/v1/completions` afterward.

**What's shipping**

- `.github/workflows/modelexpress-ci-tests.yml` — build + test matrix per inference
  engine. Composite actions under `.github/actions/` handle vCluster bootstrap, per-namespace
  apply/wait/test/cleanup, and credential plumbing.
- `ci/k8s/server/` — mx-server Dockerfile, ModelMetadata CRD, RBAC, and deployment manifest.
- `ci/k8s/client/test_p2p_k8s.py` + `conftest.py` — one shared test file parameterised
  by `--p2p-marker` so each framework reuses it with its own log-marker assertion.
- `ci/k8s/client/vllm/` — Dockerfile + Azure manifest. vLLM uses the modelexpress
  `vllm.general_plugins` entry point; same command on both source and target.
- `ci/k8s/client/trt-llm/` — Dockerfile + Azure manifest + `trtllm_p2p_launcher.py`. TRT-LLM
  needs an explicit source/target launcher because there's no plugin mechanism (target needs
  `LLM(checkpoint_loader=MxLiveCheckpointLoader(), load_format=LoadFormat.PRESHARDED)`,
  which can't be expressed via `trtllm-serve` flags). PRESHARDED + `publish_from_worker`
  patches live under `trtllm_patches/v1.3.0rc5/`.
- `ci/k8s/client/sgl/` — Dockerfile + Azure manifest scaffolded for SGLang. Matrix entries
  are commented out until upstream sgl-project/sglang#24723 lands (or we could implement the workaround in the meantime)
- `ci/TEST_PLAN.md` — coverage matrix with priority, engine support, status, and blockers.
  Captures what's in CI today and what's scoped for follow-up.

**What's verified**

- vLLM and TRT-LLM matrix entries are wired into the workflow and pass end-to-end
  when run by manual `kubectl` + `pytest` deploy against the Azure A100 cluster.
- SGLang is scaffolded but matrix entries remain commented out pending the upstream PR.

**What's intentionally gated**

- The workflow's `push:` and `pull_request:` triggers are commented out;
  `workflow_dispatch:` is the only active trigger, so the workflow won't auto-run.
  This is deliberate until we have one clean manual workflow run from the GitHub Actions UI
  to confirm the actions/composites work in the hosted environment, not just from a laptop.
  Once that's green, a follow-up uncomments the auto-triggers.
- Same gating logic for SGLang: matrix entry stays commented until the upstream PR makes
  the `--modelexpress-config` flag available in a tagged `lmsysorg/sglang` image.

**Base image decisions worth flagging for reviewers**

- vLLM uses `vllm/vllm-openai:v0.17.1` (public upstream vLLM image from Docker Hub).
- TRT-LLM uses `nvcr.io/nvidia/ai-dynamo/tensorrtllm-runtime:1.1.1`, NOT the public
  `nvcr.io/nvidia/tensorrt-llm/release` image. Reason: every PyPI `nixl[cu12]` wheel (incl. 0.10.1) requires
  UCX >= 1.21, while the public TRT-LLM image ships UCX 1.20.0 — pip-installing modelexpress
  on top of it segfaults NIXL in `ucp_init_version`. The Dynamo runtime builds NIXL from
  source against a known-good UCX with `-Dstatic_plugins=UCX`, sidestepping the issue.

### How tests are triggered and queued

GitHub Actions drives everything. Two runner types:
- **`ubuntu-latest`** (GH-hosted) — builds and pushes the worker images to `nvcr.io`.
- **`prod-modelexpress-default-small-v1`** (self-hosted, NVIDIA-managed) — all cluster
  interaction: vCluster bootstrap, namespace lifecycle, pytest execution.

Triggers: `workflow_dispatch` only for now (manual fire from the Actions UI or
`gh workflow run`). `push:` and `pull_request:` triggers are present in the YAML
but commented out until we have one clean manual workflow run to validate
end-to-end. 

### What runs on the clusters

Two layers:

**Persistent** (bootstrapped once, reused across every CI run):
- vCluster instance `mx-ci-cluster` inside host namespace `mx-ci` on the Azure AKS cluster.
- `ModelMetadata` CRD installed once inside the vCluster.

**Per-test** (created at job start, deleted at job end — namespace named
`mx-ci-<framework>-<run_id>` so concurrent runs can't collide):
- `mx-server` Deployment using the Kubernetes CRD metadata backend.
- **Source** worker Job — pulls model from HuggingFace, publishes `ModelMetadata`,
  serves `/v1/completions`.
- **Target** worker Job — discovers the source via `mx-server`, pulls weights
  over NIXL/UCX/InfiniBand RDMA, serves `/v1/completions`.
- `pytest` runs from the self-hosted runner (port-forwarded into the pods),
  asserts (a) the RDMA transfer log line appears in the target pod, (b) both
  endpoints return valid completions.

The host cluster is Azure AKS with GPU nodes and InfiniBand connectivity.
vCluster is used because the `ModelMetadata` CRD is cluster-scoped — vCluster
gives us a full nested Kubernetes where we can apply CRDs freely without needing
cluster-admin on the host. This follows the same pattern Dynamo's CI uses.

### GitLab mirror

This CI runs directly off `ai-dynamo/modelexpress` on GitHub using the self-hosted 
runner pool and repo secrets. 

## Next Steps

In Progress implementation to cover the following test cases
- TP > 1 cross-host (row 5)
- Model Streamer S3 (row 6, vLLM only for now)

## Test plan

- [ ] Manually fire the workflow via `workflow_dispatch` in the GitHub Actions UI on this branch and confirm `build`, `build-server`, `ensure-vcluster`, and `test (vllm)` + `test (trt-llm)` matrix entries all go green
- [ ] If green, follow up with a tiny PR that uncomments the `push:` and `pull_request:` triggers
- [ ] When sgl-project/sglang#24723 lands in a tagged `lmsysorg/sglang` image, uncomment the `sglang` matrix entries (build + test) and re-fire the workflow

## Summary by CodeRabbit

# Release Notes

* **Tests**
  * Added comprehensive P2P RDMA weight-transfer testing infrastructure for multiple inference frameworks (vLLM, TRT-LLM, SGLang) across Kubernetes environments (Azure and GCP).

* **Chores**
  * Introduced automated CI/CD pipeline with coordination server deployment and configuration management for distributed testing scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/modelexpress/pull/279)


.github/actions/run-mx-p2p-test/action.yml
```diff
@@ -0,0 +1,308 @@
...
+        kubectl rollout status deployment/mx-server -n ${NAMESPACE} --timeout=120s
+        echo "::endgroup::"
+
+    - name: Clear stale ModelMetadata CRs
+      if: inputs.source_extra_env != ''
+      shell: bash
+      env:
+        NAMESPACE: ${{ inputs.namespace }}
+      run: |
+        kubectl delete modelmetadata -n ${NAMESPACE} --all --ignore-not-found
```
zhengluo-nv: We don't want to manually delete them. The server should handle the stale data and we want to test it

ci/k8s/server/crd-modelmetadata.yaml
```diff
@@ -0,0 +1,139 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: Better not to duplicate this file to avoid drift. Can we copy it from the examples folder at runtime?

ci/k8s/server/rbac-modelmetadata.yaml
```diff
@@ -0,0 +1,75 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: Better not to duplicate this file

.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -0,0 +1,263 @@
...
+  #   branches:
+  #     - main
+  #     - 'release/*.*.*'
+
+concurrency:
+  group: ${{ github.event_name == 'pull_request' && format('{0}-{1}', github.workflow, github.event.pull_request.number) || format('{0}-{1}', github.workflow, github.run_id) }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  MX_CI_MODEL: Qwen/Qwen2.5-0.5B
```
zhengluo-nv, line 58: Use qwen3.5 0.8B? Newer model is more complex in terms of architecture, so it's "harder" and can potentially detect more bugs

.github/actions/run-mx-p2p-test/action.yml
```diff
@@ -0,0 +1,308 @@
...
+    - name: Wait for source to publish metadata
+      if: inputs.source_extra_env != ''
+      shell: bash
+      env:
+        NAMESPACE: ${{ inputs.namespace }}
+        TIMEOUT: ${{ inputs.pod_ready_timeout_seconds }}
+      run: |
+        echo "::group::Wait for source to publish ModelMetadata CR"
+        deadline=$((SECONDS + TIMEOUT))
+        while [ $SECONDS -lt $deadline ]; do
```
nicolasnoble, line 185: If the deadline expires before metadata appears, this loop falls out without an exit code change. The next step applies the target manifest against a namespace that has no `ModelMetadata` CR, and the test fails downstream with a confusing error instead of "source never published." Want to add a final check?

```bash
if ! kubectl get modelmetadata -n ${NAMESPACE} -o name 2>/dev/null | grep -q .; then
  echo "ERROR: source did not publish metadata within ${TIMEOUT}s"
  exit 1
fi
```

ci/k8s/client/vllm/manifest-azure.yaml
```diff
@@ -0,0 +1,110 @@
...
+            - name: UCX_TLS
+              value: "rc_x,rc,dc_x,dc,cuda_copy"
+            - name: UCX_RNDV_SCHEME
+              value: "get_zcopy"
+            - name: UCX_RNDV_THRESH
+              value: "0"
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: hf-token-secret
```
nicolasnoble, line 97: The workflow comment marks `HF_TOKEN` as optional, but `hf-token-secret` is referenced unconditionally here (and in `trt-llm/manifest-azure.yaml`) without `secretKeyRef.optional: true`. If the workflow is dispatched without setting `HF_TOKEN`, the secret never gets created and the pod fails with `CreateContainerConfigError` before the test runs. Either drop "optional" in the workflow comment, mark the `secretKeyRef` as optional, or always create the secret with an empty value if needed.

.github/actions/run-mx-p2p-test/action.yml
```diff
@@ -0,0 +1,308 @@
...
+        done
+
+    - name: Cleanup test namespace
+      if: always()
+      shell: bash
+      env:
+        NAMESPACE: ${{ inputs.namespace }}
+      run: |
+        echo "::group::Delete namespace ${NAMESPACE}"
+        kubectl delete namespace ${NAMESPACE} --ignore-not-found --timeout=120s || true
```
nicolasnoble: With fixed namespace names (`mx-ci-vllm`, `mx-ci-trt-llm`) and `|| true` swallowing the failure, a stuck namespace from run N can collide with run N+1. GPU pods retained, secrets retained, CRs retained, all green on the workflow side. Drop the `|| true` so cleanup failures surface, or move to per-run-scoped namespace names like `mx-ci-vllm-${{ github.run_id }}`.

ci/k8s/server/mx-server.yaml
```diff
@@ -0,0 +1,67 @@
...
+      app: mx-server
+  template:
+    metadata:
+      labels:
+        app: mx-server
+    spec:
+      serviceAccountName: modelexpress
+      imagePullSecrets:
+        - name: worker-registry
+      containers:
```
nicolasnoble, line 30: Any reason to skip `resources.requests` and `.limits` on mx-server? Same on the worker manifests `ci/k8s/client/{vllm,trt-llm,sgl}/manifest-azure.yaml`, which set `nvidia.com/gpu` and `rdma/ib` but leave memory/cpu unset. Probably fine on the vCluster since it's effectively dedicated, but bites if any of these manifests get copy-pasted into a shared-cluster context. Worth setting at least conservative memory requests so the scheduler does the right thing.

.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -0,0 +1,263 @@
...
+name: ModelExpress CI Tests
+
+on:
+  # Automatic triggers disabled while the workflow is being validated against
+  # the cluster. Re-enable by uncommenting the push and pull_request blocks
+  # once the workflow has been confirmed to run green end-to-end on a real PR.
+  # Keep workflow_dispatch so it can still be fired manually from the GitHub
+  # Actions UI in the meantime — GH Actions requires at least one trigger.
+  workflow_dispatch:
+  # push:
```
nicolasnoble, line 44: The workflow ships only fireable via `workflow_dispatch`. The validate-end-to-end-first rationale makes sense. Is there a tracking issue or TODO for flipping these back on? Without one, the workflow can rot silently against secrets, action interfaces, and runner pool changes.

ci/k8s/client/trt-llm/trtllm_p2p_launcher.py
```diff
@@ -0,0 +1,91 @@
...
+        level=logging.INFO,
+        format="%(asctime)s %(name)s %(levelname)s %(message)s",
+        stream=sys.stdout,
+        force=True,
+    )
+    logger = logging.getLogger(__name__)
+
+    model = os.environ.get("MODEL_NAME", "Qwen/Qwen2.5-0.5B")
+    port = int(os.environ.get("WORKER_PORT", "8000"))
+    mx_server = os.environ.get("MX_SERVER_ADDRESS", "mx-server:8000")
```
nicolasnoble: The launcher reads `MX_SERVER_ADDRESS`, but `publish_from_worker` on the source path reads `MODEL_EXPRESS_URL` per the comment in `manifest-azure.yaml`. Manifests have to set both to the same value, which is a footgun. Want to alias here? Something like:

```python
mx_server = os.environ.get("MX_SERVER_ADDRESS") or os.environ.get("MODEL_EXPRESS_URL", "mx-server:8000")
```

and the same on the publish side. Then the manifest only needs one.

.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -0,0 +1,263 @@
...
+          vcluster_namespace: ${{ env.VCLUSTER_NAMESPACE }}
+
+  # ---------------------------------------------------------------------------
+  # Run P2P weight-transfer test for each framework.
+  # Each matrix job creates its own namespace for full isolation, allowing
+  # parallel execution in the future (remove max-parallel: 1 to enable).
+  # ---------------------------------------------------------------------------
+  test:
+    name: P2P test (${{ matrix.framework }})
+    runs-on: prod-default-small-v2
```
nicolasnoble: `prod-default-small-v2` is the dynamo-repo runner pool. It's registered against `github.com/ai-dynamo/dynamo` in velonix and won't pick up jobs from this repo. PR #415 provisions `prod-modelexpress-default-small-v1` against `github.com/ai-dynamo/modelexpress`; that's the name this workflow needs to target. The GPU pool `prod-modelexpress-tester-amd-gpu-v1` from #415 isn't referenced here either. Worth confirming whether it's meant to be used or is scaffolding for a future test path; the current AKS+vCluster flow runs GPU work inside the cluster, so runner-side GPU may not be needed.

.github/actions/run-mx-p2p-test/action.yml
tanushriya910: good point, removed this in the new rev

.github/actions/run-mx-p2p-test/action.yml
tanushriya910: true, resolved in new rev

.github/actions/run-mx-p2p-test/action.yml
tanushriya910: thanks, yeah I implemented both in the new rev

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: yep, I will track this as a TODO in the linear task

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: true, we can always change the model. For the first run I just want to make sure the end to end workflow works and then we can update the model to something else

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: thanks for the pointer, yeah I just had that as a placeholder before PR#415 updated in new commit. We don't need prod-modelexpress-tester-amd-gpu-v1 for the P2P tests at-least but might pull it in later for tests that don't need necessarily need RDMA

ci/k8s/client/vllm/manifest-azure.yaml
tanushriya910: made it not optional to be consistent

ci/k8s/server/crd-modelmetadata.yaml
tanushriya910: oh right, thats what I had don't before and then someone moved it somewhere else and things started failing. so I just moved it to the dedicated ci/k8s directory for now so folks don't move it without realizing it will impact CI setup.

ci/k8s/server/mx-server.yaml
tanushriya910: added in the new commit

ci/k8s/server/rbac-modelmetadata.yaml
tanushriya910: same issue as with crd-metadata. will follow up later to see a better way to avoid duplication without breaking ci setup

.github/actions/connect-vcluster/action.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Connect to vCluster
+      id: connect
+      shell: bash
+      env:
+        VCLUSTER_NAME: ${{ inputs.vcluster_name }}
+        NAMESPACE: ${{ inputs.vcluster_namespace }}
+      run: |
+        echo "::group::Port-forward and generate vCluster kubeconfig"
+        kubectl port-forward -n ${NAMESPACE} svc/${VCLUSTER_NAME} 8443:443 &
+        sleep 5
```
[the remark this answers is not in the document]
tanushriya910: following the example for dynamo for now for vcluster setup, will adjust as a follow up in case we see issues

.github/actions/setup-mx-vcluster/action.yml
```diff
@@ -0,0 +1,127 @@
...
+    required: false
+    default: 'mx-ci-cluster'
+  vcluster_namespace:
+    description: 'Host namespace where the vCluster will be created'
+    required: false
+    default: 'mx-ci'
+  vcluster_k8s_version:
+    description: 'Kubernetes version for the vCluster control plane'
+    required: false
+    default: 'v1.32.13'
```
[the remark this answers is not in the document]
tanushriya910: will need to check with ops it is supported, this is the same setting as dynamo for now

ci/k8s/client/sgl/manifest-azure.yaml
```diff
@@ -0,0 +1,126 @@
...
+          resources:
+            limits:
+              nvidia.com/gpu: "1"
+              rdma/ib: "1"
+          volumeMounts:
+            - name: shm
+              mountPath: /dev/shm
+          securityContext:
+            capabilities:
+              add: [IPC_LOCK]
```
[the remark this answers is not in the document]
tanushriya910: followed the examples in modelexpress folder but will follow up in the next PR on tightening this as any changes in manifest will need to be tested again to make sure it works fine

ci/k8s/server/Dockerfile.server
```diff
@@ -0,0 +1,31 @@
...
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates libssl-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /app/target/release/modelexpress-server .
+
+EXPOSE 8000
+
+CMD ["./modelexpress-server", "--port", "8000"]
```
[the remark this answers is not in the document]
tanushriya910: followed the examples in modelexpress folder but will follow up in the next PR on tightening this as any changes in manifest/dockerfile will need to be tested again to make sure it works fine

ci/k8s/client/sgl/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+# Workaround while waiting for upstream merge: build SGLang from the PR
+# branch as a base image and `FROM` that here instead (see docs/SGLANG.md
+# "build from local PR branch" recipe). The CI workflow's sglang matrix
+# entry is intentionally commented out until this dependency clears.
+#
+# Build (from repo root):
+#   docker buildx build --platform linux/amd64 \
+#       -f ci/k8s/client/sgl/Dockerfile \
+#       -t <IMAGE> --push --load .
+FROM lmsysorg/sglang:latest
```
zhengluo-nv: The lastest image points to the latest release version, which does not have our MX integration PR. You can use this nightly image for now

https://github.com/ai-dynamo/modelexpress/commit/e047ee746e1c1db56c501001b819f0bdd200934e#diff-e550205c3551a735ec629e95d1ff80d85b3db3e7c2388b700be2f2263f7b7a6a

ci/k8s/client/sgl/manifest-azure.yaml
```diff
@@ -0,0 +1,131 @@
...
+            - remote_instance
+            - --remote-instance-weight-loader-backend
+            - modelexpress
+            - --modelexpress-config
+            - '{"url": "mx-server:8000", "transport": "nixl"}'
+          env:
+            - name: MODEL_EXPRESS_URL
+              value: "mx-server:8000"
+            - name: MX_SERVER_ADDRESS
+              value: "mx-server:8000"
```
zhengluo-nv, line 101: nit: duplicate

ci/k8s/client/sgl/manifest-azure.yaml
```diff
@@ -0,0 +1,131 @@
...
+            - name: UCX_TLS
+              value: "rc_x,rc,dc_x,dc,cuda_copy"
+            - name: UCX_RNDV_SCHEME
+              value: "get_zcopy"
+            - name: UCX_RNDV_THRESH
+              value: "0"
+            - name: HF_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: hf-token-secret
```
zhengluo-nv, line 113: Wondering what hf token will we use the CI tests? Are you going to create a dedicated HF account or use dynamo ops team's?

ci/k8s/client/trt-llm/Dockerfile
```diff
@@ -0,0 +1,74 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: Shall we wait until TRTLLM PRs are merged and integration is ready? Is this test case skipped?

.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -0,0 +1,263 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - framework: vllm
+            dockerfile: ci/k8s/client/vllm/Dockerfile
+          - framework: trt-llm
```
zhengluo-nv: This may should be gated. Tracking in https://linear.app/nvidia/issue/MX-299/trt-llm-adapter

.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -0,0 +1,263 @@
...
+            dockerfile: ci/k8s/client/trt-llm/Dockerfile
+          # SGLang — gated on upstream sgl-project/sglang#24723 (adds the
```
zhengluo-nv: This is no longer gated

ci/k8s/client/sgl/Dockerfile
tanushriya910: got it, updated

ci/k8s/client/sgl/manifest-azure.yaml
tanushriya910: dynamo ops team is configuring this as a github secret so it will be the same we are using in dynamo CI tests

ci/k8s/client/trt-llm/Dockerfile
tanushriya910: I commented it out from the test matrix for now while we are waiting for open PRs to merge

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: updated

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: updated

zhengluo-nv (approved):

PR #280, opened 2026-05-14
feat: add TRT-LLM engine integration

zhengluo-nv:
## Overview

Adds the ModelExpress-owned side of the TRT-LLM integration. This pairs with [NVIDIA/TensorRT-LLM#14151](https://github.com/NVIDIA/TensorRT-LLM/pull/14151), where TRT-LLM keeps only the minimal checkpoint-loader delegation surface.

TRT-LLM passes its initialized model and mapping into `modelexpress.engines.trtllm.loader.MXCheckpointLoader`. ModelExpress owns source discovery, identity construction, metadata lookup/publication, NIXL transfer, heartbeat, fallback behavior, and TRT-LLM-specific tensor discovery.

TRT-LLM does not have separate source and target launch modes in this path. Every server starts the same way; the first compatible server loads natively and publishes metadata, and later servers can receive weights from a ready source through the shared ModelExpress strategy chain.

## What changed

**TRT-LLM adapter and loader:** adds `modelexpress.engines.trtllm` with `TrtllmAdapter` and `MXCheckpointLoader`. The loader runs through the shared `LoadStrategyChain`, matching the vLLM and SGLang integration shape.

**Load-format naming:** the canonical TRT-LLM load format is `modelexpress`, avoiding `mx` because that conflicts with MXFP4/MXFP8 terminology. vLLM native support also uses `modelexpress`; vLLM plugin registration keeps `mx` as a backward-compatible alias for older deployments.

**Shared strategy behavior:** adds small shared hooks needed by TRT-LLM, including selected-strategy tracking and a source-publish path that can run after engine post-load processing.

**Post-load publication:** TRT-LLM publishes only after its weight loading, post-load processing, and CUDA stream synchronization have completed. This avoids advertising tensor addresses/content before TRT-LLM has finished mutating the model, while still allowing a successful P2P receiver to become a source for later replicas.

**Engine-neutral config cleanup:** loading config is carried through `LoadContext` instead of re-reading environment variables after context construction. `MX_MODEL_URI` remains a gate for ModelStreamer, and the resolved model URI is frozen into `LoadContext` by each engine adapter.

**Legacy cleanup:** removes the old `modelexpress.trtllm_live_transfer` implementation and deletes the historical `trtllm_patches/v1.3.0rc5` patch scripts.

**TRT-LLM example image:** keeps only the minimum image-build helpers under `examples/p2p_transfer_k8s/client/trtllm`: install the local ModelExpress client and fix NIXL runpath. Runtime TRT-LLM patching is intentionally not committed here; local test-only patches can live outside the repo.

## Compatibility notes

- Requires the TRT-LLM-side delegation hooks from [NVIDIA/TensorRT-LLM#14151](https://github.com/NVIDIA/TensorRT-LLM/pull/14151), or an equivalent TRT-LLM image with those hooks.
- The Python package owns the TRT-LLM ModelExpress behavior; TRT-LLM should import the high-level ModelExpress loader instead of low-level MX internals.
- The example Dockerfile now assumes `TRTLLM_IMAGE` already contains the TRT-LLM ModelExpress hooks; override `TRTLLM_IMAGE` with a local validation image until those hooks are available in a released TRT-LLM image.
- The successful e2e run used TP=4 source and target co-located on one B200 node. It validates the loader, metadata, rank matching, RDMA transfer, receiver publication, and serving readiness; cross-node fabric scheduling remains a separate cluster-capacity validation.

### Unit and local validation

Focused validation after rebasing onto current `origin/main`:

```bash
UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest \
  tests/test_model_streamer_strategy.py \
  tests/test_vllm_adapter.py \
  tests/test_trtllm_loader.py
```

Result: **53 passed**.

Additional local checks:

```bash
git diff --check
UV_CACHE_DIR=/private/tmp/uv-cache uv run python -m py_compile \
  modelexpress/load_strategy/context.py \
  modelexpress/load_strategy/model_streamer_strategy.py \
  modelexpress/engines/vllm/adapter.py \
  modelexpress/engines/sglang/adapter.py \
  modelexpress/engines/trtllm/adapter.py \
  modelexpress/engines/trtllm/loader.py \
  tests/test_model_streamer_strategy.py \
  tests/test_vllm_adapter.py \
  tests/test_trtllm_loader.py \
  ../../examples/p2p_transfer_k8s/client/trtllm/fix_nixl_runpath.py
bash -n examples/p2p_transfer_k8s/client/trtllm/install_modelexpress_client.sh
pre-commit run --files <touched files>
```

Result: passed.

### TRT-LLM e2e

Validated direct `trtllm-serve` with `nvidia/Kimi-K2.5-NVFP4` on the nscale B200 cluster.

| Check | Result |
|---|---|
| Image | `nvcr.io/nvidian/dynamo-dev/model-express-dev-containers:trtllm-mx-1.3.0rc11-narrow-2992272-publishfix2-amd64` |
| Source | TP=4 source cold-loaded from disk and published 4 Ready ModelExpress metadata records |
| Target source discovery | TP=4 target found one rank-matched ready source worker per rank |
| Target weight path | RDMA P2P; no disk fallback lines observed |
| Transfer | 2188 tensors / 151.11 GB per rank in about 5.55s |
| Throughput | about 217-218 Gbps per rank |
| Target metadata | target published 4 additional Ready metadata records after post-load publish |
| Serving | `/health` and `/v1/models` returned HTTP 200 |

## Follow-ups

- Build the example image from a TRT-LLM base that already includes the ModelExpress hooks once such an image is available.
- Run a separate cross-node TRT-LLM e2e once clean B200 nodes with enough GPUs/RDMA are available.


modelexpress_client/python/modelexpress/load_strategy/base.py
```diff
@@ -189,11 +185,44 @@ def publish_metadata(ctx: LoadContext) -> None:
...
+    """
+    result = _as_load_result(result_or_model)
+    register_tensors(result, ctx)
+    publish_metadata(ctx)
+
+    model = result.model
+    if model is None:
+        return
+
+    _retain_source_runtime(model, ctx)
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in `ebc1971`: `publish_loaded_model()` now takes a `LoadResult` only and returns before `register_tensors()`, `publish_metadata()`, or `_retain_source_runtime()` when `result.publishable` is false. The TRT-LLM post-load path now wraps the model in `LoadResult(value=model, model=model)` so there is a single caller shape instead of accepting raw models here. Added regression coverage for the non-publishable path. Validation: `UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest tests/test_trtllm_loader.py` passed with 23 tests, `git diff --check` passed, and touched-file pre-commit passed.

zhengluo-nv:
Addressed the remaining CodeRabbit nitpicks in `ebc1971`:

- Pinned the temporary TRT-LLM validation image default from `:latest` to `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc11@sha256:d91c80ba8baf763782b1078267ed6b1e06363bebff4961094bf6e5679d371d04`, while keeping `TRTLLM_IMAGE` override support. README now documents the pinned default and override.
- Replaced hardcoded `/usr/local/lib/python3.12/dist-packages` NIXL binding discovery with dynamic discovery from `site.getsitepackages()`, `sysconfig.get_paths()["purelib"]`, and `sys.path`.
- Pinned `grpcio==1.66.2` and `grpcio-tools==1.66.2` in the temporary TRT-LLM image install helper, leaving the protobuf constraint unchanged.

Validation: `tests/test_trtllm_loader.py` passed with 23 tests, helper-script `py_compile` passed, `git diff --check` passed, and touched-file pre-commit passed.

modelexpress_client/python/modelexpress/load_strategy/model_streamer_strategy.py
```diff
@@ -49,17 +45,18 @@ def is_available(self, ctx: LoadContext) -> bool:
             )
             return False
 
-        model_uri = os.environ.get("MX_MODEL_URI", "")
-        if not model_uri:
+        if not ctx.model_streamer_uri:
```
nv-hwoo, line 49: `MX_MODEL_URI` semantics changed silently. On main this was a gate, setting anything truthy to enable streaming, and the actual URI came from `model_config.model_weights` or `model_config.model`. After this PR the env var IS the URI. Anyone running with MX_MODEL_URI=1 to enable streaming will now try to stream from `"1"`. Is that intentional? If yes I think it deserves a CHANGELOG / release-notes line so deployments know to update. If no, the adapters could keep the old gate-vs-URI separation.

modelexpress_client/python/modelexpress/engines/trtllm/loader.py
```diff
@@ -0,0 +1,384 @@
...
+            """Load TRT-LLM weights through MX, falling back to native disk load.
+
+            `mapping` matches TRT-LLM's BaseCheckpointLoader contract. `model`
+            is an MX-only extension passed by TRT-LLM's model loader so P2P can
+            write directly into live parameter buffers. Keep `model` explicit
+            instead of inside `kwargs` so native fallback loaders never receive
+            an argument they do not understand.
+            """
+            self._p2p_succeeded = False
+            self._last_load_ctx = None
```
nv-hwoo: `_last_load_ctx` gets wiped if `load_weights()` is ever called a second time without model=.

Every call to `MXCheckpointLoader.load_weights()` resets `self._last_load_ctx = None` and `self._p2p_succeeded = False` at the very top (lines 210-211), before checking whether model was actually passed in. If a second call comes in without `model=`, it falls through to disk-fallback at line 223, but the resets have already erased the state from the first (successful) call. Later, `post_load_publish(primary_model)` checks `ctx.adapter.model` is not model (line 342), sees no matching context, and silently skips publication. The primary load went fine, but it never gets advertised as a source, so other replicas can't pull from it.

Today, TRT-LLM rc11 (the image this PR validated against) only calls load_weights() once, so this never triggers. But on TRT-LLM main, when speculative decoding is enabled, model_loader.py does:
1. `load_weights(primary_dir, mapping=..., model=primary_model)`: good, has `model=`
2. `load_weights(spec_config.speculative_model, mapping=...)`: no `model=`, this is the one that wipes state
3. `post_load_publish(primary_model, ...)`: now silently skips

The patch script (`patch_trtllm_mx_runtime.py`) only injects `model=` into call site 1, not 2. So the moment we bump the base image to one that has the draft-weight load path, this becomes a live bug: P2P transfer succeeds, but no receivers can chain off the source, and you'd only notice by looking at why a second-replica startup fell back to disk.

Two fixes, ideally both:
- In `loader.py`, move the two reset lines to right before `_build_load_context` (line 226), so the model is None branch returns early without touching prior state. (1/2)
- In `patch_trtllm_mx_runtime.py`, also inject `model=model` into the draft-load call site so MX can see and route the draft load too. (Or explicitly opt out of MX for draft weights, whichever you prefer.)

Could you also add a regression test along the lines of load_weights(model=A) → load_weights(model=None) → post_load_publish(A) still publishes? That would have caught this.

modelexpress_client/python/modelexpress/engines/trtllm/loader.py
```diff
@@ -0,0 +1,384 @@
...
+            handler, "baseFilename", None
+        ) == rank_log:
+            mx_logger.removeHandler(handler)
+            handler.close()
+
+    formatter = _TrtllmStyleFormatter(
+        "[%(asctime)s] [ModelExpress] [%(mx_level_tag)s] %(message)s",
+        datefmt="%m/%d/%Y-%H:%M:%S",
+    )
+    stream = open(rank_log, "w", buffering=1)
```
nv-hwoo: We're opening the rank log with `"w"`. It's entered both during `load_weights()` and again during `post_load_publish()` for the same rank, so the post-load scope wipes the load-attempt log.

zhengluo-nv: Addressed in cd9635b: rank logs now open in append mode while still cleaning duplicate handlers and flushing/fsyncing on scope exit. The TRT-LLM loader test now enters the rank-log scope twice and verifies both load and publish lines remain in the file and stderr.

nv-hwoo (approved):

zhengluo-nv (commented):
Do not merge until we came to agreement with TRTLLM on the integration interface

PR #282, opened 2026-05-14
feat(vmm-arena): per-allocation cuMemCreate + 16 TiB VA + 1-MR dmabuf

nicolasnoble:
One cuMemCreate + cuMemMap per mx_malloc into a 16 TiB cuMemAddressReserve,
selective free via cuMemUnmap + cuMemRelease in mx_free, single-MR registration
at end-of-load through cuMemGetHandleForAddressRange + ibv_reg_dmabuf_mr over
the full bump range.

PyTorch's caching allocator amortizes tensor allocations into pool segments
before reaching us, so one plugin call equals one physical handle. The bump
pointer never reuses VA, so cumulative allocation determines high-water rather
than peak live. 16 TiB of VA reserved up front costs nothing while no cuMemMap
has happened, and CUDA virtual addresses are 49 bits on Hopper/Ada/Blackwell.

cuMemGetHandleForAddressRange with CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD exports
a dmabuf for the whole bump range. ibv_reg_dmabuf_mr pins the currently-mapped
physical pages and produces one lkey/rkey covering every live tensor. The
registration succeeds over a VA range with mid-range holes from prior
cuMemUnmaps, and the HCA translation table survives subsequent CUDA-side
unmaps because the dmabuf attach holds the pin.

The hot path is a per-arena C extension capsule: std::atomic<size_t>
next_offset for the bump pointer, std::atomic<int> state, mutex-protected
unordered_map<va, handle>. PyTorch's mx_malloc / mx_free dispatch through a
module-global active-capsule pointer, mirroring the Python use_arena context.
Py_mod_gil = Py_MOD_GIL_NOT_USED declared behind a Python 3.13+ build guard.
close_and_drain spins on an in_flight counter after the state CAS, with the
GIL released during the spin so concurrent malloc/free Python callbacks can
complete. PyList_New and Py_BuildValue failures roll the state back to OPEN
so the caller can retry close and recover the records. The CAS bump loop,
map insert + mapped_bytes accounting, and free-side state-check + lookup +
fetch_sub all run inside Py_BEGIN_ALLOW_THREADS / Py_END_ALLOW_THREADS so
concurrent threads attach independently under nogil.

Deployment requirement: UCX_CUDA_COPY_REG_WHOLE_ALLOC=off until the upstream
UCX cuda_copy_md length-truncation fix at https://github.com/openucx/ucx/pull/11461
ships. Without the workaround, UCX internally rewrites the requested
registration length via cuMemGetAddressRange, which returns per-handle bounds
on multi-handle VMM, and the resulting MR covers only one chunk.

Empirical validation on Blackwell B200 + ConnectX over InfiniBand:

| Test                                              | Load time | MR | register_arena | Bump    | Wire        |
|---------------------------------------------------|----------:|---:|---------------:|--------:|------------:|
| Qwen3-30B-A3B-FP8 single-pod disk-load            |    15.5s  | 1  |          0.1s  | 32.0 GB |         n/a |
| Qwen3-30B-A3B-FP8 P2P target receive              |     8.8s  | 1  |          0.2s  | 32.5 GB | 210.8 Gbps  |
| Hermes 405B FP8 TP=8 single-pod disk-load (rank)  |     250s  | 1  |          0.3s  | 51.4 GB |         n/a |

Hermes is the structural acceptance: 405B FP8 TP=8 cumulative allocation peaks
at 51.4 GB per rank with mx_free reclaiming process_weights_after_loading's
discarded tensors during load. Qwen3-30B P2P inference is byte-identical
between source and target.

## Release Notes

* **New Features**
  * Added VMM Arena support with optional C++ extension for CUDA allocations, enabling single-registration path via `MX_VMM_ARENA=1` environment variable
  * Extended environment variable configuration options for arena registration and CUDA optimization

* **Documentation**
  * Updated architecture, deployment, and contribution guides with VMM Arena details, configuration, and deployment recommendations

* **Tests**
  * Added comprehensive test coverage for VMM Arena functionality across allocation, lifecycle, and CUDA integration scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/modelexpress/pull/282)


modelexpress_client/python/modelexpress/vmm/_alloc_ext.cpp
```diff
@@ -0,0 +1,706 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: Can you move all these vmm files into a vmm folder and add a README to explain this feature? Not only for customers but also for agent hand over

modelexpress_client/python/setup.py
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Build script for ModelExpress C++ extensions.
+
+Project metadata lives in pyproject.toml. This setup.py only adds the
+ext_modules entry that pyproject.toml can't express. Setuptools merges
+the two when both are present.
+
+Pattern follows dynamo's GMS setup.py - pure setuptools build, no CUDA
```
zhengluo-nv: nit: Don't need to mention GMS

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -267,6 +267,88 @@ def register_tensors(self, tensors: dict[str, torch.Tensor]) -> bytes:
 
         return self._metadata
 
+    def register_arena(self, arena: Any, tensors: dict[str, torch.Tensor]) -> bytes:
```
zhengluo-nv: nit: why `arena: Any`?

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -267,6 +267,88 @@ def register_tensors(self, tensors: dict[str, torch.Tensor]) -> bytes:
...
+        internally truncates the requested length via
+        cuMemGetAddressRange (which on multi-handle VMM returns
+        per-handle bounds, not the full reserve).
+        """
+        if self._agent is None:
+            raise RuntimeError("NIXL agent not initialized")
+
+        self._tensors = tensors
+        tensor_descriptors = []
+        for name, tensor in tensors.items():
```
zhengluo-nv: Duplicate code alert: this method is similar to `def register_tensors(self, tensors: dict[str, torch.Tensor]) -> bytes`. Consider abstract some parts?

nicolasnoble: Done in 3a5969a. Extracted _build_tensor_descriptors so both call sites are one line each and the contiguity-check contract lives in one place.

modelexpress_client/python/setup.py
nicolasnoble: Done in fa8e884. Also swept the two other GMS call-outs in vmm/backend.py and load_strategy/base.py while in the area.

modelexpress_client/python/modelexpress/nixl_transfer.py
nicolasnoble: Done in e478aef. Used a TYPE_CHECKING forward ref so the runtime cuda-python import stays optional.

modelexpress_client/python/modelexpress/vmm/_alloc_ext.cpp
nicolasnoble: Done in 83f1e1b. Files moved under modelexpress/vmm/ as arena.py / backend.py / hook.py / _alloc_ext.cpp. Added an __init__.py for the public surface and a README covering customer-facing setup plus an agent-handoff reading order.

modelexpress_client/python/modelexpress/vmm/arena.py
```diff
@@ -0,0 +1,355 @@
...
+            raise VmmArenaError(
+                "_vmm_alloc_ext C extension is not available; "
+                f"import failed: {_EXT_IMPORT_ERROR}"
+            )
+        if total_bytes <= 0:
+            raise ValueError("total_bytes must be positive")
+
+        self._device = device
+        self._backend: VmmBackend = backend if backend is not None else _StubBackend()
+        self._granularity = self._backend.allocation_granularity()
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in 465e7bc, went stricter than the proposed default. VmmArena now raises ValueError on backend=None instead of defaulting to CudaVmmBackend or _StubBackend. Production callers must pass it explicitly (loader.py already does); 28 test sites updated to pass _StubBackend() explicitly. Eliminates the silent-fallthrough class entirely.

modelexpress_client/python/modelexpress/vmm/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Set one environment variable on the worker process:
+
+```
+MX_VMM_ARENA=1
+```
+
+That is the only knob. The arena enables itself on supported devices and
+falls back silently when the underlying C extension is missing.
+
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in 4df8902. Aligned line 25 to "falls back with a warning" since that matches actual runtime behavior; the loader does log on each fallback path.

modelexpress_client/python/setup.py
```diff
@@ -0,0 +1,94 @@
...
+    and emits a warning instead of aborting the install. The resulting
+    wheel is pure-Python; the runtime detects the missing .so and falls
+    back gracefully.
+    """
+
+    def build_extensions(self):
+        cxx = os.environ.get("CXX", "g++")
+        self.compiler.set_executable("compiler_so", cxx)
+        self.compiler.set_executable("compiler_cxx", cxx)
+        self.compiler.set_executable("linker_so", f"{cxx} -shared")
```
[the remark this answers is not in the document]
nicolasnoble: Keeping the default. $CXX is already the override mechanism for alternative toolchains; setting CXX=clang++ picks up the alternate compiler. Removing the default lets setuptools auto-detect, which on Linux uses `gcc` for compiler_so and linker_so and is broken for C++ extensions (no -lstdc++ link, no -x c++ flag). `g++` is the right default for Linux C++ extension builds.

modelexpress_client/python/modelexpress/vmm/arena.py
```diff
@@ -0,0 +1,355 @@
...
+        rounded = ((total_bytes + self._granularity - 1) // self._granularity) * self._granularity
+        self._total_bytes = rounded
+        self._base = self._backend.reserve(rounded)
+        self._capsule = _vmm_alloc_ext.arena_create(
+            self._base,
+            self._total_bytes,
+            self._granularity,
+            self._perform_cuda_alloc,
+            self._perform_cuda_dealloc,
+        )
```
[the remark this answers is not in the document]
nicolasnoble: Applied verbatim in 465e7bc.

modelexpress_client/python/modelexpress/vmm/hook.py
```diff
@@ -0,0 +1,205 @@
...
+        _active_arena = arena
+        _vmm_alloc_ext.set_active_arena(arena.capsule)
+
+    try:
+        with torch.cuda.use_mem_pool(_mem_pool, device=device):
+            yield
+    finally:
+        with _active_lock:
+            _vmm_alloc_ext.set_active_arena(None)
+            _active_arena = None
```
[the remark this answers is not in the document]
nicolasnoble: Applied verbatim (order swap) in 465e7bc.

modelexpress_client/python/modelexpress/vmm/_alloc_ext.cpp
```diff
@@ -0,0 +1,706 @@
...
+  }
+  Py_END_ALLOW_THREADS
+
+  if (!found) {
+    return;
+  }
+
+  call_dealloc_callback(
+      state, ptr, record.size, record.handle, write_unraisable);
+}
```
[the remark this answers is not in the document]
nicolasnoble: The literal swap (erase-after-dealloc-succeeds) introduces double-free risk: if a caller frees the same ptr twice and the first dealloc raised, the second free would call cuMemUnmap/cuMemRelease again on the same handle. The retry-from-close_and_drain premise also assumes dealloc errors are transient; for cuMem* these are typically permanent (NOT_INITIALIZED, INVALID_VALUE, DEINITIALIZED), so retry doesn't help.

The real bug under this finding was the silent-clear half. Fixed in 465e7bc: py_arena_free now passes write_unraisable=true, so direct-Python arena.free() errors surface via sys.unraisablehook instead of PyErr_Clear, matching mx_free's behavior.

modelexpress_client/python/modelexpress/vmm/README.md
```diff
@@ -0,0 +1,160 @@
...
+collapses the kernel's MR setup cost and reduces RDMA fast-path overhead
+on transfers that touch hundreds or thousands of tensors.
+
+## Quick start
+
+Set one environment variable on the worker process:
+
+```
+MX_VMM_ARENA=1
+```
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in 4df8902.

modelexpress_client/python/tests/test_vmm_arena.py
```diff
@@ -0,0 +1,668 @@
...
+        arena = VmmArena(total_bytes=1 << 30)
+        a = arena.malloc(1024)
+        assert a == arena.base
+
+    def test_size_rounds_up_to_granularity(self):
+        backend = _StubBackend()
+        arena = VmmArena(total_bytes=1 << 30, backend=backend)
+        arena.malloc(1024)
+        # Backend sees the granularity-rounded size, not the raw request.
+        va, size, handle = backend.allocations[0]
```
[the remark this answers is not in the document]
nicolasnoble: Fixed in 4df8902.

modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -86,12 +89,215 @@ def patched(self: VllmConfig) -> None:
...
+_vmm_arenas: dict[int, "object"] = {}
+
 
 # Install eagerly: vllm calls try_verify_and_update_config during engine init,
 # before LoadStrategyChain.run() would lazily import the strategy module.
 _patch_vllm_s3_format_check()
 
 
+@contextmanager
+def _maybe_enter_vmm_arena(ctx: LoadContext) -> Iterator[None]:
```
zhengluo-nv: Should this sharable with sglang, trtllm? If so, this should not be in the vllm loader file

nicolasnoble: Good call. Done in acf90b9: moved into modelexpress/vmm/runtime.py as engine-neutral maybe_enter_vmm_arena + log_arena_post_load. engines/vllm/loader.py just imports and calls now; when SGLang or TRT-LLM land they'll reuse the same helpers instead of cloning the lifecycle.

zhengluo-nv (approved):

PR #291, opened 2026-05-18
feat(lifecycle): add pause_serving / resume_serving for sleep/wake

nv-hwoo:
## Summary

Make MX safe to use under GMS sleep/wake by fixing two distinct bugs in the
NIXL re-registration path. After this lands, GMS can call `mx_teardown()` on
sleep and `mx_bringup()` on wake without the model walker tripping on
post-warmup artifacts.

Ref MX-312.

## Why

GMS drives MX through a sleep/wake cycle on the active engine: on sleep,
`mx_teardown()` shuts down the NIXL agent; on wake, `mx_bringup()` calls
`register_tensors()` again. Two latent bugs on `main` make this path crash:

### 1. Model walker trips on post-warmup artifacts

After vLLM warmup / `torch.compile` / CUDA graph capture, the model carries
`torch._dynamo.aot_compile.AOTCompiledFunction` objects that transitively
reference llvmlite's `_lib_fn_wrapper`, `torch.distributed` deprecation shims,
and HuggingFace `_LazyAutoMapping`. The hidden-tensor discovery walker (added
in #241) does `vars(obj).items()` on attribute subtrees and blows up:

```
AttributeError: '_lib_fn_wrapper' object has no attribute 'items'
```

Tensor discovery is a property of the weights (frozen after
`process_weights_after_loading`), not the model wrapper. Re-walking on wake
is unnecessary *and* unsafe.

### 2. Teardown silently empties the caller's tensor dict

`NixlTransferManager._build_tensor_descriptors()` did `self._tensors = tensors`
(shared reference) and `shutdown()` did `self._tensors.clear()`. Together they
wiped the caller's dict in place. Net effect: even with the architectural fix
from (1), `mx_bringup`'s `if ctx.tensors: <fast path>` guard sees an empty
dict on wake and falls through to the walker — same `_lib_fn_wrapper` crash,
architectural fix bypassed.

Together: (1) gives GMS a safe wake path; (2) makes (1) actually fire.

### Unit tests (this branch)

6 new regression tests in `test_vllm_loader.py`:

- `test_reuse_skips_discovery_when_ctx_has_tensors` — `reuse_discovered=True` skips `discover_tensors` when `ctx.tensors` is populated.
- `test_reuse_falls_back_to_discovery_when_ctx_tensors_empty` — caller-misuse safety: falls back when `ctx.tensors` is empty.
- `test_default_behavior_unchanged` — default `reuse_discovered=False` is byte-identical to pre-PR behavior.
- `test_register_takes_shallow_copy_of_caller_dict` — `_build_tensor_descriptors` produces a dict distinct from the caller's.
- `test_shutdown_does_not_clear_caller_dict` — caller's dict survives a full `register_tensors` → `shutdown()` round trip.
- `test_shutdown_rebinds_internal_containers` — defense-in-depth: even if a future caller bypasses `register_tensors` and aliases `_tensors` directly, `shutdown()` won't mutate the shared container.

Full `test_vllm_loader.py` locally: **61 passed, 8 skipped, 0 failed** (skips are CUDA/NIXL-gated and not exercised by this branch's changes).

### End-to-end on real hardware

Two cluster runs, both green.

#### Qwen3-0.6B TP=2 on `dynamo-aks-exp` — full GMS sleep/wake cycle

- 10/10 sanity (5/5 source + 5/5 target).
- Cross-node RDMA P2P from the wake-recovered source engine produces correct inference on the target.
- Failover green: `SIGKILL` on active engine → kubelet restart → shadow's flock-driven wake → `Remap complete` → `Allocated KV cache on wake`.

#### Kimi-K2.5-NVFP4 TP=4 on `nscale-b200` — P2P stress, exotic config

- **P2P verification:** all 4 ranks registered 2945 tensors / 149.59 GB each;
  cross-node transfer source→target completed in ~20s for ~600 GB total.
- **Target load time:** 20.3s vs source's ~13min from disk — **~38× speedup**.
- **Sanity:** 10/10 PASS on both endpoints with semantically equivalent
  responses to identical prompts, confirming weight integrity through P2P.
- Re-verified on rebased branch tip after pickup of `main`'s VMM arena
  refactor (#282).

## Downstream

The dependent Dynamo PR — [ai-dynamo/dynamo#8218](https://github.com/ai-dynamo/dynamo/pull/8218)
— is pinned against this MX branch's tip. Once this merges, that PR
rebases onto post-merge MX `main` and can move to merge as well.

## Checklist

- [x] Pre-commit clean (`cargo fmt`, `cargo clippy`, ruff, end-of-file fixers).
- [x] 6 new unit tests added; full `test_vllm_loader.py` green locally.
- [x] End-to-end sleep/wake validated on Qwen3-0.6B TP=2.
- [x] P2P-only path re-validated on Kimi-K2.5-NVFP4 TP=4 on the rebased branch tip.
- [ ] CI green on the PR.
- [ ] Code review approved.
- [ ] After merge: rebase dynamo#8218 onto post-merge MX `main`.

## Release Notes

* **New Features**
  * Added support for safe reuse of previously discovered tensors during re-registration operations (e.g., after restore or wake flows).

* **Improvements**
  * Enhanced tensor container handling to prevent unintended mutations of external references.

* **Tests**
  * Added comprehensive unit tests for tensor reuse and container ownership semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/modelexpress/pull/291?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


modelexpress_client/python/modelexpress/lifecycle.py
```diff
@@ -0,0 +1,96 @@
...
+                ctx.global_rank, e,
+            )
+        ctx.nixl_manager = None
+
+
+def resume_serving(
+    ctx: LoadContext,
+    model: nn.Module,
+    *,
+    worker_id: str | None = None,
```
zhengluo-nv: From caller perspective, how does caller know mx worker id? We should make this to preserve the worker id by default like

```
def resume_serving(ctx, model, *, new_worker_id: bool = False):
    if new_worker_id:
        ctx.worker_id = uuid.uuid4().hex[:8]
```

new_worker_id=False preserves ctx.worker_id and republish/reuses the existing worker record. Use for in-place wake after GMS remap.

new_worker_id=True mints a fresh worker id before publishing. Use when the resumed process should appear as a distinct source instance, such as relocation/failover where the old worker may still exist.

modelexpress_client/python/modelexpress/lifecycle.py
```diff
@@ -0,0 +1,96 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: Please add tests for this file

modelexpress_client/python/modelexpress/lifecycle.py
nv-hwoo: Currently the caller doesn't provide worker ID when calling this method so we always mint a fresh worker ID, which I believe is necessary at the moment because there's no way for the waking up process to know if it's waking up from the local GMS failover (from shadow -> main engine) or restoring from snapshotted checkpoint from another pod.

I added an optional argument for any future extension but I can remove if that's more clear.

zhengluo-nv: Shall we make wake up reuse the original worker ID by default? Agent checked for me and our metadata will handle same worker ID correctly. It will update the metadata with the new value. If we use a new ID, it spams a lot of stale metadata records

zhengluo-nv (approved):

PR #297, opened 2026-05-23
ci(k8s): add disaggregated dynamo P2P test + stale-metadata test

tanushriya910:
## Summary

This PR extends the dynamo + p2p CI coverage that #294 introduced. Two new tests land in the main workflow, plus a handful of supporting changes that make both old and new tests more deterministic.

The first new test exercises **disaggregated serving**. Where the aggregated test treats each worker as a self-contained inference unit, disaggregated splits the same workload across a prefill component (computes KV cache, ships it via NIXL) and a decode component (receives KV, generates tokens). The test deploys 1 prefill + 1 decode replica, then scales the decode component to 2 — that single sequence exercises the prefill↔decode KV-transfer path (unique to disagg) as well as the decode↔decode P2P weight transfer path that the aggregated test already covers. 

The second new test addresses **row 18 of the TEST_PLAN** — does the server actually stop returning stale sources after a worker dies? This is meaningful coverage because the reaper logic (heartbeat tracking, status transitions, the client-side `status_filter=READY` filter) is all production code that's never been exercised end-to-end in CI. The test deploys a single source against an mx-server configured with the Redis backend and compressed reaper knobs (heartbeat=30s, scan=10s), force-kills the source pod, then queries `ListSources(status_filter=READY)` via gRPC and asserts zero instances are returned. Matrix over `vllm` / `sglang` so we catch framework-specific publish-path regressions. The Redis backend got pulled in alongside because we wanted the test to be backend-agnostic in spirit — the reaper logic is identical for Redis and K8s CRD, but Redis was untested in CI until now, so we get backend coverage as a side effect.

Along the way, a few related cleanups: the disaggregated test surfaced a race where the Frontend would round-robin to a worker still in CUDA-graph capture, so the dynamo action now gates pytest on every worker pod's `/live` endpoint returning 200; the Frontend inference test now fires 6 concurrent `/v1/completions` requests instead of one, so the scaled-up replica's P2P-loaded weights actually get exercised on the request path.

## Test plan

Successful run at: https://github.com/ai-dynamo/modelexpress/actions/runs/26464529361

## Summary by CodeRabbit

* **New Features**
  * Added support for disaggregated serving topology in Dynamo integration testing
  * Introduced stale metadata detection testing to verify handling after pod failures
  
* **Tests**
  * Enhanced Dynamo P2P tests with parameterized metadata counts and concurrent inference requests
  
* **Chores**
  * Updated CI workflows with serving topology matrix variations
  * Added Redis-backed metadata server manifests for testing

[Review Change Stack](https://app.coderabbit.ai/change-stack/ai-dynamo/modelexpress/pull/297?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ci/k8s/client/test_stale_metadata.py
```diff
@@ -0,0 +1,135 @@
...
+import time
+
+import grpc
+
+from kube_utils import port_forward
+
+# Stubs are generated by the action into the same directory as this test file.
+sys.path.insert(0, "ci/k8s/client")
+import p2p_pb2  # noqa: E402
+import p2p_pb2_grpc  # noqa: E402
```
KavinKrishnan: nit (39-41):

if anyone ever runs
cd ci/k8s/client && pytest test_stale_metadata.py, this prepends a
non-existent directory and you'll get a confusing ImportError because of this

suggest either drop the line entirely (pytest's got you covered) or do the
__file__-relative version:


```
sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
```

ci/k8s/client/test_stale_metadata.py
```diff
@@ -0,0 +1,135 @@
...
+            f"Expected exactly 1 READY source before kill, got {len(live)}. "
+            f"This means the source either never published or there are leftover "
+            f"sources from a previous test run."
+        )
+
+        # Force-delete the source pod. --grace-period=0 + --force simulates
+        # sudden death (no graceful shutdown signal to the modelexpress
+        # client, so it can't deregister itself with the server — exactly
+        # the failure mode the reaper is supposed to handle).
+        result = subprocess.run(
```
KavinKrishnan, line 117: probably need to kill the job iiuc, new pod will spawn otherwise with current command


```
subprocess.run(
    ["kubectl", "-n", namespace, "delete", "job", "mx-source",
     "--grace-period=0", "--cascade=foreground", "--timeout=30s"],
    capture_output=True, text=True, check=True, timeout=60,
)(
```

KavinKrishnan (commented):
overall looks good, CI system is new to me so would be good for others to look as well

ci/k8s/client/test_stale_metadata.py
tanushriya910: good point, I think the test was still passing cause the new pod take time to come up. updated to job deletion now.

ci/k8s/client/test_stale_metadata.py
tanushriya910: thanks, updated

ci/k8s/client/vllm/dynamo/manifest-azure-disaggregated.yaml
```diff
@@ -0,0 +1,261 @@
...
+            rdma/ib: "1"
+      envs:
+        - name: VLLM_PLUGINS
+          value: "modelexpress"
+        - name: MX_SERVER_ADDRESS
+          value: "${DGD_NAME}-modelexpress:8000"
+        - name: MX_RDMA_NIC_PIN
+          value: "auto"
+        - name: UCX_TLS
+          value: "rc_x,rc,dc_x,dc,cuda_copy"
```
zhengluo-nv: Can you remove UCX_TLS from all CI manifests and rerun the test? To align with https://github.com/ai-dynamo/modelexpress/pull/336

tanushriya910: removed in the updated commit.

ci/k8s/client/kube_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        deadline = time.perf_counter() + 60
+        while time.perf_counter() < deadline:
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+                sock.settimeout(2)
+                try:
+                    sock.connect(("localhost", local_port))
+                    break
+                except (ConnectionRefusedError, socket.timeout, OSError):
+                    time.sleep(1)
+        yield local_port
```
nv-hwoo, line 69: In case this times out, should we raise with some kind of TimeoutError? Right now both happy path and timeout all returns with `yield local_port` no matter what, which I think is hard to distinguish what went wrong later.
```python
deadline = time.perf_counter() + 60
ready = False
while time.perf_counter() < deadline:
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
        sock.settimeout(2)
        try:
            sock.connect(("localhost", local_port))
            ready = True
            break
        except (ConnectionRefusedError, socket.timeout, OSError):
            time.sleep(1)
if not ready:
    raise TimeoutError(
        f"kubectl port-forward to {target} (local:{local_port} -> remote:{remote_port}) "
        f"not ready after 60s"
    )
yield local_port

```

zhengluo-nv (approved):

ci/k8s/client/kube_utils.py
tanushriya910: you are right, callers will get a downstream error connection refused on the next operation instead of a clear "port-forward never established"/"timed-out error". I have another PR in the works, I can include a clearer timeout error message in that one.

nv-hwoo: Works for me, thanks

nv-hwoo (approved):

PR #349, opened 2026-05-27
feat(RL/post-2389): MX V2 Support in RL + WeightTransferEngine Support

KavinKrishnan:
## Summary

Adds the **framework-agnostic client-side surface MX needs to support RL weight refit**.

The same primitives are consumed unchanged by every RL framework integration (NeMo RL, PrimeRL, verl) — each framework just provides a thin glue layer on top, no per-framework forking of the MX clients. Introduces both layers as a single deliverable, because RL needs the v2 fat-client API but the v2 layer reuses the v1 building blocks underneath:

**I. v1 base** (`MxTrainingPublisher` / `MxRefitReceiver` in `training_publisher.py`, `refit_receiver.py`)
- Minimal NIXL register + publish + receive primitives.
- Also reusable for the existing v1 cold-start path.

**II. v2 fat-client layer on top** (`MxV2TrainingPublisher` / `MxV2RefitReceiver` / `MxWeightTransferEngine` in `nemo_rl_v2.py`, `engines/vllm/weight_transfer.py`)

Along with four RL primitives:

- **(1) Per-rank shard publish** — via `MxV2TrainingPublisher.add_tensor`
  - Trainer GPUs publish their own shards directly (no allgather to rank 0, which doesn't scale past a few billion params).
- **(2) Metadata envelope** — via `TensorDescriptorV2.extra_parameters` (new map field on the proto)
  - Receivers know what each shard is (column-parallel Q, expert 17, etc.).
- **(3) Version field** — via `SourceIdentity.revision` (new string field on the proto) + `MxV2RefitReceiver.discover_v2_sources(min_version=...)`
  - Receivers can pull "only v5 weights."
- **(4) vLLM adapter** — `MxWeightTransferEngine`
  - Implements vLLM's standardized `WeightTransferEngine` ABC ([vLLM PR #43375](https://github.com/vllm-project/vllm/pull/43375), [native RL APIs blog](https://blog.vllm.ai/2026/05/28/native-rl-apis.html)) so RL frameworks drop MX in without modifying vLLM.

## Status

- ✅ Conflicts resolved (merged `origin/main`; union-merge for `SourceIdentity` field set, regenerated proto stubs via `generate_proto.sh` with pinned `grpcio-tools==1.66.2`).
- ✅ 76/76 unit tests pass on the merged HEAD.
- ✅ `mergeable: MERGEABLE`. DCO + Python Tests + Test Suite (stable/beta/nightly) all green.
- ✅ All 17 CodeRabbit findings addressed in a prior round (1 Critical, 12 Major Quick-wins, 4 Major Heavy-lifts).

## Reviewer guidance

A few things worth pre-empting for reviewers coming in cold:

### Why v1 + v2 are one PR

The v1 base and v2 fat-client layer are released together because:

- v2's `MxV2RefitReceiver` instantiates an `MxRefitReceiver` internally — the v2 layer is a wrapper around v1's NIXL register + publish + receive primitives, not a replacement.
- Cutting v1 alone would land a code path with no downstream consumer (nothing in this PR or `main` actually uses `MxTrainingPublisher` / `MxRefitReceiver` standalone).
- Both layers are required for the downstream PRs that depend on this — splitting them would force two-step coordination for every downstream merge.

The v2 layer adds the four RL primitives listed above; v1 is the substrate it sits on.

### Why `MxWeightTransferEngine` doesn't live under `EngineAdapter`

There are two distinct "engine" abstractions in this codebase, and they target different lifecycle phases:

| Abstraction | Lifecycle | Hook | Where |
|---|---|---|---|
| `EngineAdapter` (existing) | **Cold-start** model load at startup | MX-internal `LoadStrategyChain.run(model, ctx)` | `engines/vllm/adapter.py` |
| `MxWeightTransferEngine` (this PR) | **Per-step RL refit** during training | vLLM upstream `WeightTransferEngine` ABC (vLLM PR #43375) | `engines/vllm/weight_transfer.py` |

Both live under `engines/vllm/` so they're discoverable side-by-side, but they're independent — different contracts, different hook points, different `__init__` shapes. `MxWeightTransferEngine` does not extend `EngineAdapter` because it isn't going through `LoadStrategyChain`; it's plugged directly into vLLM's own factory and called per-refit-cycle, not at startup.

A back-compat shim at `modelexpress/vllm_weight_transfer.py` re-exports the public symbols so existing callers don't break.

### Why `nemo_rl_v2.py` is long (~1350 lines)

The file is the fat-client API surface — one cohesive class (`MxV2RefitReceiver`) with its dataclasses, slice planner, and source picker. Splitting it across modules would split the API across imports for downstream RL framework integrations that already learn one entry point. The length is intentional, not a refactor TODO.

### Benchmarks and demo scripts

`benchmarks/bench_elastic_scaling.py` and `scripts/v2_*_e2e_demo.py` are bundled here because they exercise the v2 client surface in a self-contained way — the benchmark is the validation harness for the new client API, and the demos are the canonical usage examples downstream consumers reference. The benchmark Job manifest is fully parameterized (`<NAMESPACE>`, `<IMAGE>`, `<IMAGE_PULL_SECRET>`); the README documents the env-var protocol.

### Per-tensor `compile_target` + `compile_metadata` on `TensorDescriptorV2`

New per-tensor fields default to `hf_raw` / `{}`; the wire encoder omits them when default so existing payloads stay byte-identical. New constants for canonical-BF16, vLLM-fused, DeepGEMM-FP8, CUTLASS-FP8, and TRT-LLM kernel formats. New helper `compile_target_matches()` for receiver-side filtering with whitelist + required-metadata-subset semantics.

### `compile_target_filter` on `MxV2RefitReceiver.discover_v2_sources`

New kwargs `compile_target_filter` (whitelist set) and `required_compile_metadata` (dict that must be a subset of every tensor's `compile_metadata`). Candidates with no v2 registry are rejected when either filter is set; candidates with mixed compile targets are rejected if any tensor falls outside the allowed set.

### Multi-source slice discovery for mixed trainer/inference TP

New types: `TargetTPLayout`, `SliceSource`, `SliceCoveragePlan`. New methods `MxV2RefitReceiver.discover_v2_sources_for_slice` and `MxV2RefitReceiver.receive_via_plan`. The planner walks v2 candidates per tensor, intersects each publisher's `local_shard_range` against the receiver's requested slice, emits the minimal candidate set covering it; surfaces coverage gaps and shard-axis mismatches in `plan.missing`. `receive_via_plan` orchestrates per-candidate scratch RDMA pulls and stitches via `torch.cat` along the shard axis.

### Proto extensions

- `TensorDescriptorV2.extra_parameters: map<string, string>` — escape hatch downstream RL clients use for per-tensor metadata (parallelism role, kernel format, expert ID, version, etc.). Heavily used by [#429](https://github.com/ai-dynamo/modelexpress/pull/429).
- `SourceIdentity.revision: string` — content-addressed weight version. Non-empty value guarantees two sources with identical `SourceIdentity` have bit-identical weight bytes, enabling decentralized modes (no central coordinator) to use `mx_source_id` as a full content check.
- Redis backend `SourceAttributesJson` carries the new fields plus the union of main's compatibility additions (`backend_framework_version`, `torch_version`, `cuda_version`, `triton_version`, `gpu_arch`, `compile_config_digest`).

## Tests

68/68 client unit tests + cluster-tier green.

## Compatibility

Every new field has a backwards-compatible default and every new method arg is optional. Existing demos and downstream consumers run unchanged. The back-compat shim at `modelexpress/vllm_weight_transfer.py` re-exports the public symbols from the new canonical location at `modelexpress.engines.vllm.weight_transfer`.

## Downstream impact

This PR is the prerequisite for the rest of the RL workstream. Several downstream PRs already in flight import the v2 surface or `nemo_rl_v2` module:

- [modelexpress#421](https://github.com/ai-dynamo/modelexpress/pull/421) — `publish_self_as_source` tree fan-out fix (merged into the `kavink/nemo_rl_moe` integration branch, awaiting this PR to reach main)
- [modelexpress#429](https://github.com/ai-dynamo/modelexpress/pull/429) — Megatron-Core MX clients (slice planner + vendored Bridge math + receiver translator + v1 NIXL sliced-pull). Same integration-branch situation.
- [modelexpress#450](https://github.com/ai-dynamo/modelexpress/pull/450) — `MX_RDMA_NIC_PIN=stripe` multi-NIC mode
- [ai-dynamo/dynamo#10900](https://github.com/ai-dynamo/dynamo/pull/10900) — first-time upstream port of `dynamo.vllm.mx_refit` extension; imports `MxV2RefitReceiver` + Megatron modules
- [ai-dynamo/dynamo#10901](https://github.com/ai-dynamo/dynamo/pull/10901) — buffer-caching perf fix in Dynamo extension
- [jthomson04/RL#2](https://github.com/jthomson04/RL/pull/2) (merged) + [#7](https://github.com/jthomson04/RL/pull/7) — NeMo RL Megatron-MX integration + perf fix

Once this PR lands, `kavink/nemo_rl_moe` (the integration branch holding #421 + #429) can be retired by rebasing those PRs onto main directly.


KavinKrishnan:
**Ready for review — conflicts resolved, all tests green.**

Merged `origin/main` into the branch today (2026-06-25). Resolutions:

* `modelexpress_server/src/p2p/service.rs` — combined main's `worker_tensor_count` helper with this branch's `identity_present` logging. Both wins preserved.
* `modelexpress_server/src/p2p/backend/redis.rs` — `SourceAttributesJson` now carries the union of fields. This branch's `extra_parameters` + `revision` are preserved; main's `backend_framework_version`, `torch_version`, `cuda_version`, `triton_version`, `gpu_arch`, `compile_config_digest` are also retained. Both `::from(&SourceIdentity)` and `::to_source_identity()` construct the full field set.
* `modelexpress_common/proto/p2p.proto` — auto-merged cleanly (no overlap between the two sides' new fields).
* `modelexpress_client/python/modelexpress/p2p_pb2.py` and `modelexpress_client/go/gen/modelexpress/p2p/p2p.pb.go` — regenerated from the merged `.proto` via the existing `generate_proto.sh` scripts. No hand-merging of generated descriptor tables.

Smoke:

* `mergeable: MERGEABLE` (was `CONFLICTING` since May).
* All four client surfaces (`MxV2RefitReceiver`, `MxV2TrainingPublisher`, `MxRefitReceiver`, `MxTrainingPublisher`) import cleanly from top-level `modelexpress`.
* 68/68 unit tests pass across `test_v2_source_picker`, `test_v2_shape_registry`, `test_source_id`, `test_types`, `test_vllm_adapter`.

Pre-merge state preserved on `kavink/post-2389-phase3-4.backup-pre-rebase-20260625` if anyone wants to compare.

The PR description has been refreshed with the current dependency context — several downstream PRs (#421, #429, #450, ai-dynamo/dynamo#10900, jthomson04/RL#2 / #7) have been waiting on this to reach main.

KavinKrishnan:
**Addressed all 17 CodeRabbit findings.** Force-pushed the squashed commit on `kavink/post-2389-phase3-4`.

## Critical (1)

| File | Fix |
|---|---|
| `training_publisher.py:308` | `mark_ready()` now flips every per-layer `mx_source_id` published for the current step, not just the final one. Tracks via `_pending_source_ids` keyed on step boundary. Without this, all but the last layer's metadata stays `INITIALIZING` and READY-only discovery sees a partial weight set. |

## Major Quick-wins (12)

| File | Fix |
|---|---|
| `training_publisher.py:318` | `shutdown()` resets `_registered`, `_publish_mode`, and `_pending_source_ids` so reinit produces a clean publisher. |
| `refit_receiver.py:341/448/496` | NIXL tuple is `(bytes, tensors, duration)` — fixed the 3 call sites that were storing `tensors` as `bytes_skipped`. Derived `bytes_skipped = max(advertised − transferred, 0)`. |
| `refit_receiver.py:505` | `receive_weights_from_metadata` now filters `__mx_`-prefixed sidecar descriptors before NIXL, matching the other paths. |
| `nemo_rl_v2.py:286` (registry) | `_registry` is no longer append-only — re-adding a tensor with the same name now replaces the existing descriptor, preventing stale entries on a long-lived publisher. |
| `nemo_rl_v2.py:286` (DTensor) | `add_tensor` now unwraps DTensors via `to_local()` for the registered tensor, runs `.contiguous()` if needed, and checks CUDA residency on the local shard. Callers no longer need to patch `_registered_tensors` privately. |
| `nemo_rl_v2.py:1151` | Non-divisible default TP slice now returns an explicit `_TensorPlan` with a coverage-gap reason instead of silently truncating via `axis_total // world_size`. Also guards `world_size <= 0` and out-of-range rank. |
| `vllm_weight_transfer.py:314` | Mixed-TP path no longer inherits `same_rank_only=True` from `update_info`; defaults to cross-rank discovery via new optional `mixed_tp_same_rank_only` field. |
| `vllm_weight_transfer.py:491` | `_MxEngineForVllm` MRO reordered to put `MxWeightTransferEngine` first, so its concrete methods resolve before the vLLM ABC's abstract definitions. |
| `bench-elastic.yaml:158` | Per-scenario failures now captured via `SCENARIO_FAILURES` counter; Job exits non-zero after the result-collection sleep if any scenario failed. |
| `bench-elastic.yaml:203` | Added pod-level + container-level `securityContext`: `runAsNonRoot: true`, `runAsUser: 1000`, `allowPrivilegeEscalation: false`, drop all caps then add `IPC_LOCK` + `SYS_RESOURCE` (RDMA pinning), `seccompProfile: RuntimeDefault`. |
| `bench_elastic_scaling.py:418` | Replaced `captured.extend` with a discard callback — benchmark only measures wall + bandwidth, doesn't need to hold tensors. Prevents OOM on real-scale models. |
| `test_bench_elastic_scaling.py:125` | Tree-fanout test fixtures now match the harness — replica receivers get unique `worker_rank=i+1` so `source_worker_rank` distinguishes them from the trainer (rank 0). |

## Major Heavy-lifts (4)

| File | Fix |
|---|---|
| `nemo_rl_v2.py:848` | `pick_best_source` no longer returns trainer candidates unconditionally. Trainers are subject to the same `owned_experts_per_layer` superset check as inference replicas — a trainer rank with EP > 1 only holds a slice of experts. If no single candidate covers the request, returns `None` so the caller multi-sources via `discover_v2_sources_for_slice`. Updated `test_pick_best_falls_back_to_trainer` (renamed to `test_pick_best_requires_ownership_for_trainer`) to reflect the corrected contract. |
| `vllm_weight_transfer.py:365` | Documented + guarded the known limitation: scratch-mode `receive_from_scratch` doesn't populate `_registered_buffers`, so `publish_self_as_source` would silently no-op. Now emits a loud `logger.warning` when `publish_self_as_replica=True` is set with empty `_registered_buffers`. Proper fix (carry scratch buffers into `_registered_buffers`, or wait for vLLM `register_destinations`) tracked as separate work. |
| `shape_descriptors.py:243` | Uneven DTensor sharding (where `global_extent ≠ local_extent × world_size`) now raises `RuntimeError` instead of publishing an incorrect `local_shard_range` for ranks > 0. v2 v0 doesn't support uneven shards; callers can pass a custom `target_range` or rebalance the layout. |
| `bench_elastic_scaling.py:168` | Tree fan-out receivers now spawn with unique `worker_rank=i+1` and `same_rank_only=False`, so `trainer_egress_bytes()` correctly distinguishes trainer pulls (`source_worker_rank == 0`) from replica pulls (non-zero). Plumbed `same_rank_only` through `_run_receiver` → `MxUpdateInfo`. |

## Validation

- 76/76 unit tests pass locally (`test_v2_source_picker`, `test_v2_shape_registry`, `test_source_id`, `test_types` (excl. local protobuf-version test), `test_vllm_adapter`, `test_bench_elastic_scaling`).
- All 7 modified Python files compile cleanly.
- Single squashed signed commit preserved on the PR.

CI is re-running on the new HEAD.

KavinKrishnan:
**Pre-review pass: addressed the anticipated reviewer concerns before they land.** Force-pushed `85dfa4e` on `kavink/post-2389-phase3-4`.

Mapping anticipated concerns to mitigations applied:

| # | Anticipated concern | Mitigation |
|---|---|---|
| 1 | Committed benchmark JSON result files (3 files) were transient per-cluster-session artifacts | Deleted; added `benchmarks/results-*/` to `modelexpress_client/python/.gitignore`. |
| 2 | Internal namespace + private registry hardcoded in `bench-elastic.yaml` | Parameterized as `<NAMESPACE>`, `<IMAGE>`, `<IMAGE_PULL_SECRET>`. `run_cluster_bench.sh` now reads the three values from `MX_BENCH_*` env vars and runs the substitution via `sed`. New "Configuration" section in `benchmarks/README.md` documents the env-var protocol. |
| 3 | `vllm_weight_transfer.py` lived at the top level instead of under the established `engines/<framework>/` pattern | Moved to `modelexpress/engines/vllm/weight_transfer.py` (alongside the existing `adapter.py` + `loader.py`). A back-compat shim at the old path re-exports the public symbols so existing callers don't break. Logger name + tests + benchmarks + scripts all updated to the new canonical path. The PR description's "Reviewer guidance" section explains why this is distinct from `EngineAdapter` (different lifecycle: cold-start vs. per-step refit; different upstream hook: `LoadStrategyChain` vs. vLLM's `WeightTransferEngine` ABC). |
| 4 | Two `@property` defs in `weight_transfer.py` had no return annotations | Added `-> "TransferStats | None"` and `-> "list[TransferStats]"` (forward-string refs to avoid a circular import). |
| 5 | "One concern per PR" — v1 + v2 + benchmarks + demos + protos all bundled | Pre-empted in PR description under "Reviewer guidance / Why v1 + v2 are one PR." The summary: v2 wraps v1's primitives at runtime, so cutting v1 alone lands a code path with no downstream consumer; the benchmarks are the validation harness for the new client API; the demos are the canonical usage references for the downstream RL framework integrations. |
| 6 | 1352-line `nemo_rl_v2.py` might attract "split into modules" feedback | Pre-empted in PR description: it's one cohesive fat-client class + its dataclasses + slice planner + source picker; splitting it would fracture the API surface across imports for downstream consumers that already learn one entry point. |
| 7 | CodeRabbit re-review on force-push | The 17 findings from the previous CR pass remain addressed; the changes since then are structural (move + parameterize + delete), not semantic. CR should converge quickly. |

Other changes worth noting:

- **`run_cluster_bench.sh`** now does the placeholder substitution via `sed` (no `envsubst` dep) and gates on the three required env vars; missing values fail loud with a clear message.
- **Tests still 76/76 green** on the merged HEAD; the post-move test suite (15/15 in `test_vllm_weight_transfer.py`) passes against the new canonical import path.

Diff stat on top of the prior CodeRabbit-response commit: 8 files changed (3 deletions of JSON results, 1 file move with shim, 4 in-place edits for parameterization + path updates). No semantic change to the v1/v2 client surface itself.

yafshar:
@KavinKrishnan I identified a NIXL registration leak in the scratch receive path while testing this branch.
`MxRefitReceiver.receive_weights_scratch()` allocates fresh scratch tensors on each receive and was registering them with NIXL without retaining or deregistering the returned registration descriptors. As a result, registrations accumulated over time. 

In a live NIXL benchmark, the receiver agent metadata grew linearly with repeated scratch receives:

```sh
9878 -> 19190 -> 28502 -> 37814 -> 47126
```

I implemented a fix that scopes scratch tensor registrations to the duration of a single receive and explicitly deregisters them once the transfer completes. Persistent tensor registrations remain unchanged and are still released at shutdown, preserving the expected publisher metadata lifecycle.

With this fix, the same benchmark remains stable and bounded:

```sh
count: 21
first 10: [9882, 9878, 9878, 9878, 9878, 9878, 9878, 9878, 9878, 9878]
last 10: [9878, 9878, 9878, 9878, 9878, 9878, 9878, 9878, 9878, 9878]
first: 9882 last: 9878 delta: -4
```

I pushed a commit to this PR branch with the fix and validation details: #457 

Validation was performed using `modelexpress_client/python/benchmarks/bench_elastic_scaling.py` in live mode with a single receiver and repeated scratch receives.

KavinKrishnan:
Superseded by the consolidated integration branch in #482, which already contains this v2-base + WeightTransferEngine work (now the native "mx" backend). Closing to reduce open-PR surface — the branch and this discussion are preserved, and this will be re-cut as a clean, small review slice (v2 client base) from #482. No work is lost.

PR #363, opened 2026-05-27
fix(vllm): use torch.distributed world rank for MX worker_rank

AndyDai-nv:
## Summary

`worker_rank` in the vLLM adapter now uses `torch.distributed.get_rank()` (world rank) instead of `parallel_config.rank % (tp_size * pp_size)`. Fixes MX P2P weight loading for MoE deployments with DP > 1.

## Validation

Mistral-Small-4-119B-2603, TP=4 DP=2, FP8 MoE, GSM8K 5-shot.

| Condition | Iters | Byte-identical | Mean acc drop |
|---|---|---|---|
| Pre-fix (EP off) | 24 | 4 / 24 | +0.173 (worst +0.70) |
| Post-fix EP off | 32 | 32 / 32 | -0.001 |
| Post-fix EP on  | 10 | 10 / 10 | +0.002 |

## Test plan

- [x] vLLM adapter unit tests updated
- [x] 42 / 42 post-fix GSM8K iters byte-identical with acc within noise vs seed

SGLang adapter not changed in this PR (defer to a follow-up once a SGLang multi-DP MoE deployment is available to validate against).

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -39,7 +39,7 @@ class VllmAdapter(EngineAdapter):
         return build_source_identity(self.vllm_config, self.model_config)
 
     def get_worker_rank(self) -> int:
-        return _get_vllm_worker_rank(self.vllm_config)
+        return _get_vllm_worker_rank(self.vllm_config, self.target_device)
 
     def get_global_rank(self) -> int:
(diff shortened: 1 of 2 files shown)


AndyDai-nv:
/ok to test [d136dfc](https://github.com/ai-dynamo/modelexpress/commit/d136dfcec3450f2b2d9674c2cfc094288dbb769b)

zhengluo-nv (approved):

PR #404, opened 2026-06-01
fix(ngc): download via files-manifest endpoint for all storage/scope combos

AndyDai-nv:
## Summary

The NGC provider (added in #232) could not download artifacts in **any** storage/scope combination. It used the versioned files endpoint (`.../versions/{version}/files`) and branched on V1/V2 storage and org/team scope, which produced three independent failures:

| Scope / Storage | Symptom | Cause |
|---|---|---|
| org & team / **V1** | `NGC artifact has no files` | Versioned endpoint returns `modelFiles` only (no `urls`/`filepath`) → empty list |
| org / **V2** | `HTTP 401 Unauthorized` | Per-file `/files/{path}` needs Bearer, but auth forwarding was keyed on team-vs-org, so org sent none |
| team / **V2** | `HTTP 404` on nested files | Per-file `/files/{path}` URLs can't address subdirectories |

The HuggingFace provider was unaffected because it uses the `hf-hub` crate; only the hand-rolled NGC provider was broken.

## Fix

Switch to the **files-manifest endpoint** `.../v2/org/{org}[/team/{team}]/{type}/{name}/{version}/files` (no `versions/` segment). For both V1/V2 storage and both org/team scopes it returns self-authenticating presigned `urls[]` paired with `filepath[]` for every file, top-level and nested alike. Because the URLs are presigned:

- no per-file URL construction (fixes the nested-path 404),
- no `Authorization` forwarding (fixes the 401),
- V1 is no longer empty (fixes "has no files").

The two scope-specific fetch functions collapse into one `fetch_artifact_files`; the V1/V2 branching, `is_v2_storage`, and the `modelFiles`/`storageVersion` types are removed (net −152 lines). The `checksums.blake3` UAM fallback is preserved for orgs whose listing endpoint is UAM-gated (400/401).

## Verification

Unit tests (44) pass, `clippy` clean, `fmt` clean, server/client compile.

Validated **end-to-end against the live NGC API** for all four cells via `model download --provider ngc --strategy direct`:

| Cell | Model | Result |
|---|---|---|
| org / V1 | `nvidian/aigs_xavier_dla_models/1` | ✅ 442 MB, nested dirs |
| team / V1 | `nvidia/nemo/llama-3_1-8b-instruct-nemo/2.0` | ✅ 15 GB, nested `weights/`, `context/nemo_tokenizer/` |
| org / V2 | `nvidian/cbottle/1.2` | ✅ 11 GB — previously-401 `cBottle-3d/README` downloads |
| team / V2 | `nvidia/earth-2/fourcastnet3/0.1.0` | ✅ 2.7 GB — previously-404 nested `training_checkpoints/best_ckpt_mp0.tar` downloads |

## Summary by CodeRabbit

* **Documentation**
  * Updated NGC Provider architecture documentation with current download flow details and API key resolution configuration.

* **Improvements**
  * Enhanced NGC artifact download handling with improved fallback mechanisms for authentication scenarios.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -503,7 +503,7 @@ pub trait ModelProviderTrait: Send + Sync {
 
 Three implementations:
 - `HuggingFaceProvider` - uses the `hf-hub` crate with high-CPU download mode.
-- `NgcProvider` - downloads from NVIDIA NGC via the V2 artifact API (Bearer-authenticated `/files/{path}` for team artifacts; presigned S3 URLs for org-level artifacts). Falls back to `checksums.blake3` manifest enumeration when bulk file listing returns 400. Resolves the NGC API key from `NGC_API_KEY`, `NGC_CLI_API_KEY`, or `~/.ngc/config`.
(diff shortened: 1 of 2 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/404/commits/114b867975b8ef3852b77bc185d01a065ad6c340

zhengluo-nv (approved):

PR #408, opened 2026-06-02
fix(clear): remove registry record and verify disk on cache hit

AndyDai-nv:
## Problem

`model clear` deletes only the local cache files; it never removes the server-side registry record. The cleared model keeps an orphaned `status=DOWNLOADED` record in Redis. A subsequent default `model download` (smart-fallback) then hits that stale record and returns `✅ SUCCESS / downloaded successfully` without downloading anything — the cache directory is still absent and `model list` still reports the model missing.

### Reproduce
```
modelexpress-cli model download Qwen/Qwen3-0.6B
modelexpress-cli model clear Qwen/Qwen3-0.6B -p hugging-face
modelexpress-cli model download Qwen/Qwen3-0.6B   # ✅ SUCCESS but nothing on disk
```

### Root cause
1. `clear_model` (`handlers.rs`) → `cache.rs` deletes local files only; there was no delete RPC, so the registry record was reachable only via LRU eviction and became orphaned.
2. `ensure_model_downloaded` (`services.rs`) returned the `AlreadyExists(DOWNLOADED)` claim without ever checking that the files exist on disk.

## Fix

**A — `clear` removes the registry record too**
- New `DeleteModel` RPC on `ModelService` (reuses `delete_status`/`delete_model`).
- `model clear` and `model clear-all` now delete the registry record(s) alongside local storage. Best-effort: if the server is unreachable, local files are still cleared and a warning notes the record may remain.

**B — verify disk before honoring a `DOWNLOADED` cache hit**
- `ensure_model_downloaded` checks that the cached files exist before honoring a `DOWNLOADED` claim. If they are missing, it drops the stale record and re-claims so the model is actually re-downloaded. Bounded to two attempts.

Both together: A fixes the clear-side registry leak; B is a backstop for any registry-vs-storage inconsistency.

## Tests
- New server unit tests for `model_files_present` (present / missing / no-cache-dir).
- Updated the 5 mock `ModelService` impls for the new RPC.
- `cargo build`, `cargo clippy --workspace --all-targets` (no warnings), `cargo fmt --check`, and `cargo test --workspace` all pass (server tests 165 → 167).

## Docs
- `docs/ARCHITECTURE.md` — RPC table + behavior note.
- `docs/CLI.md` — `model clear` behavior.

## Summary by CodeRabbit

* **New Features**
  * `model clear` now removes server-side model registry records to prevent false cache hits on future downloads.
  * Extended `model clear --all` to also remove registry records for all cached models.
  * Graceful fallback: local files are cleared even if the server is unreachable, with a warning about potential stale registry records.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -299,8 +299,11 @@ Four proto files define four services, all compiled via `tonic-build` in `modele
 | `EnsureModelDownloaded` | `ModelDownloadRequest` | stream `ModelStatusUpdate` | Trigger download, stream progress |
 | `StreamModelFiles` | `ModelFilesRequest` | stream `FileChunk` | Stream model file contents (1MB chunks) |
 | `ListModelFiles` | `ModelFilesRequest` | `ModelFileList` | List files with sizes |
+| `DeleteModel` | `DeleteModelRequest` | `DeleteModelResponse` | Remove a model record from the registry (used by `model clear`) |
 
 Key message types: `ModelProvider` (HuggingFace, NGC, GCS), `ModelStatus` (Downloading, Downloaded, Error), `ModelStatusUpdate`, `FileChunk`.
+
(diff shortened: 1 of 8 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/408/commits/d5c5746d9e592da042b8026ffb9c00adc7be950b

zhengluo-nv (approved):

PR #411, opened 2026-06-02
feat: add source-typed artifact metadata payloads

zhengluo-nv:
## Overview

This PR adds the metadata contract needed for artifact transfer without changing the runtime byte-transfer path yet.

The main schema change is a source-typed metadata payload:

- `SourceIdentity.mx_source_type` distinguishes weight sources from future artifact sources.
- `WorkerMetadata.source_payload` is a `oneof` with `tensor_source` and `artifact_source`.
- Existing `WorkerMetadata.tensors = 3` remains for backward-compatible readers, but new writers populate `source_payload.tensor_source`.
- Artifact metadata is bounded to discovery fields: `artifact_id`, `total_size`, `file_count`, and `chunk_count`.

This keeps PR 1 focused on the compatibility contract and discovery shape. Artifact manifest RPCs and artifact byte transfer remain future work; this PR only makes sources self-describing enough for later artifact-transfer paths.

## Compatibility

- Weight source IDs are preserved when the new artifact compatibility fields are empty.
- Readers prefer `source_payload.tensor_source` and fall back to the deprecated `tensors` field for older writers.
- Kubernetes `ModelMetadata.spec.sourceType` is an open string, not a CRD enum, so future source types do not require a CRD schema upgrade before rollout.
- Redis and Kubernetes metadata backends both round-trip the new source type and artifact summary fields.

## End-to-End Validation

Manual e2e validation covered the worker P2P metadata path on nscale with vLLM:

- Built and deployed the branch images into the nscale ModelExpress test environment.
- Published a source for `Qwen/Qwen2.5-0.5B-Instruct` with `MX_P2P_METADATA=1`.
- Verified the central metadata entry stayed lightweight while advertising the worker metadata endpoint and worker gRPC endpoint.
- Verified the target used the worker gRPC endpoint to fetch the tensor manifest directly from the source worker.
- Verified the target transferred 267 tensors through the existing P2P tensor path and served inference successfully.

This e2e test proves the new worker metadata payload and read-path migration still support the existing tensor transfer workflow. It does not validate artifact file transfer, which is intentionally outside this PR.

## CI Validation

- Added a CPU-only `Artifact Transfer Contract` CI job for the new artifact metadata contract.
- Added Rust coverage for source identity separation, tensor-vs-artifact payload behavior, Kubernetes artifact metadata shape, and Kubernetes `totalSize` conversion bounds.
- Updated Python tests for source ID canonicalization and tensor payload fallback.
- Regenerated Python and Go protobuf bindings.

## Release Notes

* **New Features**
  * Added artifact source metadata fields to track artifact ID, total size, file count, and chunk information
  * Expanded source type classification system to cover artifact caching and material scenarios
  * Extended compatibility tracking with backend framework and runtime version fields
  * New source type column added to Kubernetes metadata display for enhanced visibility

* **Tests**
  * Added artifact transfer contract testing to CI pipeline for improved reliability


.github/workflows/ci.yml
```diff
@@ -153,6 +153,40 @@ jobs:
...
+          ${{ runner.os }}-cargo-
+
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y protobuf-compiler
+
+    - name: Run artifact transfer contract tests
+      run: |
+        cargo test -p model-express-workspace-tests --test artifact_transfer_contract
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 442cd4e: the artifact-transfer-contract job now has job-level `contents: read`, checkout uses `persist-credentials: false`, and both checkout/cache are pinned to full action SHAs.

modelexpress_common/proto/p2p.proto
```diff
@@ -109,6 +136,28 @@ message TensorDescriptor {
...
+  string artifact_id = 1;
+
+  // Total artifact bytes across all manifest files.
+  uint64 total_size = 2;
+
+  // Number of files in the sealed artifact manifest.
+  uint32 file_count = 3;
+
+  // Number of transfer chunks in the sealed artifact manifest.
+  uint32 chunk_count = 4;
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 442cd4e: updated the proto comments to make this PR explicitly bounded to discovery metadata only. `ArtifactSourceMetadata` no longer implies full file/chunk manifests are served by `WorkerService`, and `worker_grpc_endpoint`/`WorkerService` are documented as tensor-manifest only for this contract.

modelexpress_server/src/p2p/k8s_types.rs
```diff
@@ -93,6 +116,30 @@ pub struct WorkerStatus {
...
+    #[serde(rename = "totalSize")]
+    pub total_size: u64,
+
+    /// Number of files in the sealed artifact manifest.
+    #[serde(rename = "fileCount")]
+    pub file_count: u32,
+
+    /// Number of transfer chunks in the sealed artifact manifest.
+    #[serde(rename = "chunkCount")]
+    pub chunk_count: u32,
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 442cd4e: the Kubernetes-facing `ArtifactSourceStatus.total_size` is now `i64` to match the CRD schema, with checked conversions at the Kubernetes backend boundary to/from the internal/proto `u64`. I also added unit coverage for overflow and negative CRD values.

modelexpress_client/go/generate_proto.sh
```diff
@@ -52,7 +52,10 @@ echo "Generating Go protobuf and gRPC bindings from ${PROTO_DIR}..."
         "${PROTO_FILES[@]}"
 )
 
-mapfile -d '' generated_files < <(find "${GEN_DIR}" -type f \( -name "*.pb.go" -o -name "*_grpc.pb.go" \) -print0 | sort -z)
+generated_files=()
+while IFS= read -r -d '' file; do
+    generated_files+=("${file}")
+done < <(find "${GEN_DIR}" -type f \( -name "*.pb.go" -o -name "*_grpc.pb.go" \) -print0 | sort -z)
```
nv-hwoo, line 60: what is this change?

modelexpress_server/src/p2p/backend.rs
```diff
@@ -179,16 +217,13 @@ impl From<WorkerRecord> for WorkerMetadata {
         Self {
             worker_rank: record.worker_rank,
             backend_metadata,
-            tensors: record
-                .tensors
-                .into_iter()
-                .map(modelexpress_common::grpc::p2p::TensorDescriptor::from)
-                .collect(),
```
nv-hwoo, line 186: If we want to be backward compatible, we should set the deprecated `tensors` field. Currently it's set to empty vec by default, so any old clients will get empty list when talking to the new server. 

Fix: populate both fields when record.artifact_source.is_none():
```rust
  Self {
      ...
      tensors: tensors.clone(),     // keep populated for legacy readers
      source_payload,
      ..Default::default()
  }
```
  Cost is ~one Vec clone of TensorDescriptors per GetMetadata response (which already costs O(n) tensors). Trade-off is a few hundred KB of duplicate payload vs. broken cross-version reads.

  Adjust the state.rs:286 test from assert!(back.tensors.is_empty()) to assert_eq!(back.tensors.len(), 1). And add a cross-version regression test in the contract suite.

modelexpress_server/src/p2p/k8s_types.rs
```diff
@@ -25,6 +25,29 @@ pub struct ModelMetadataSpec {
...
+        match mx_source_type {
+            0 => "weights",
+            1 => "lora",
+            2 => "cuda_graph",
+            3 => "torch_compile_cache",
+            4 => "triton_cache",
+            _ => "unknown",
+        }
+        .to_string()
+    }
```
nv-hwoo, line 51: If someone reorders the proto enum or inserts a value (e.g. MX_SOURCE_TYPE_FLASHINFER_CACHE = 5), this match silently mislabels. Use the generated enum:
```rs
  use modelexpress_common::grpc::p2p::MxSourceType;
  let name = match MxSourceType::try_from(mx_source_type) {
      Ok(MxSourceType::Weights) => "weights",
      Ok(MxSourceType::Lora) => "lora",
      Ok(MxSourceType::CudaGraph) => "cuda_graph",
      Ok(MxSourceType::TorchCompileCache) => "torch_compile_cache",
      Ok(MxSourceType::TritonCache) => "triton_cache",
      Err(_) => "unknown",
  };
```

modelexpress_client/python/modelexpress/metadata/payload.py
```diff
@@ -0,0 +1,26 @@
...
+def tensor_source_metadata(
+    tensors: Sequence[p2p_pb2.TensorDescriptor],
+) -> p2p_pb2.TensorSourceMetadata:
+    return p2p_pb2.TensorSourceMetadata(tensors=list(tensors))
+
+
+def worker_tensor_descriptors(worker: p2p_pb2.WorkerMetadata):
+    if worker.WhichOneof("source_payload") == "tensor_source":
+        return worker.tensor_source.tensors
+    return worker.tensors
```
nv-hwoo, line 25: When source_payload == "artifact_source", this falls through to the deprecated tensors field, which will be empty for artifact sources. Functionally fine, but masks intent. Better:
```python
  def worker_tensor_descriptors(worker):
      which = worker.WhichOneof("source_payload")
      if which == "tensor_source":
          return worker.tensor_source.tensors
      if which == "artifact_source":
          return []     # explicit: artifact sources have no tensors
      return worker.tensors    # legacy writer fallback
```
  Also rename the helper to make the legacy fallback obvious, e.g. tensor_descriptors_or_legacy (bikeshed-able).

modelexpress_server/src/p2p/k8s_types.rs
zhengluo-nv: Updated

modelexpress_client/python/modelexpress/metadata/payload.py
zhengluo-nv: Updated

modelexpress_client/go/generate_proto.sh
zhengluo-nv: Added comment

modelexpress_server/src/p2p/backend.rs
zhengluo-nv: Addressed in 46e0fc7: tensor-source responses now populate both `source_payload.tensor_source` and the deprecated `tensors` field for legacy readers. Artifact-source responses still leave legacy `tensors` empty. I updated the state test and added a contract regression test for the cross-version tensor fallback.

modelexpress_client/python/modelexpress/metadata/payload.py
zhengluo-nv: Addressed in 46e0fc7: `worker_tensor_descriptors()` now branches explicitly on `source_payload`; `artifact_source` returns `[]`, `tensor_source` returns tensor payloads, and only the unset oneof path falls back to deprecated `worker.tensors`. I kept the existing helper name to avoid churn across current call sites.

modelexpress_server/src/p2p/k8s_types.rs
zhengluo-nv: Addressed in 46e0fc7: `source_type_name_from_proto` now converts through the generated `MxSourceType` enum with `MxSourceType::try_from(...)` instead of matching raw integer values, so enum reordering will not silently mislabel known variants.

modelexpress_client/go/generate_proto.sh
zhengluo-nv: Addressed in 46e0fc7: this block collects the actual `.pb.go` / `_grpc.pb.go` files generated by protoc in this run, then applies version-comment normalization, SPDX headers, and gofmt only to that generated file set. I added a short script comment to make that intent explicit.

modelexpress_common/proto/p2p.proto
```diff
@@ -37,14 +37,18 @@ enum BackendFramework {
...
+// Type of reusable material the source is serving.
 enum MxSourceType {
-  MX_SOURCE_TYPE_WEIGHTS = 0;    // Base model weights (default)
-  MX_SOURCE_TYPE_LORA = 1;       // LoRA adapter weights
-  MX_SOURCE_TYPE_CUDA_GRAPH = 2; // CUDA graph artifacts
+  MX_SOURCE_TYPE_WEIGHTS = 0;             // Base model weights (default)
+  MX_SOURCE_TYPE_LORA = 1;                // LoRA adapter weights
+  MX_SOURCE_TYPE_CUDA_GRAPH = 2;          // CUDA graph artifacts
+  MX_SOURCE_TYPE_TORCH_COMPILE_CACHE = 3; // torch.compile cache files
+  MX_SOURCE_TYPE_TRITON_CACHE = 4;        // Triton cache files
```
nv-hwoo, line 46: One last comment: can we also add DeepGEMM into the type in v1? This is the kernel cache that'll give us largest time saving.

nv-hwoo (approved):
Approved with one comment

modelexpress_common/proto/p2p.proto
zhengluo-nv: Sure will add in the next PR

PR #423, opened 2026-06-05
feat: add artifact manifest WorkerService APIs

zhengluo-nv:
## Overview

This PR adds the artifact-manifest discovery layer for file-backed artifact sources. It stays at the metadata/API contract layer; artifact byte transfer is left for the follow-up PR.

## Changes

- Adds `ArtifactManifest`, file metadata, chunk metadata, and artifact manifest WorkerService RPCs for header plus paged chunks.
- Adds DeepGEMM cache as an `MxSourceType` so kernel caches can be represented in v1.
- Adds deterministic Rust and Python artifact manifest helpers. Artifact IDs are SHA-256 over canonical manifest JSON; file and chunk checksums are CRC32C.
- Updates the Python worker server to serve multiple artifact manifests by `artifact_id`.
- Carries artifact source summaries through server metadata while preserving tensor-source compatibility behavior.
- Adds Rust contract tests and Python WorkerService artifact manifest tests. The Python tests are covered by the existing `python-test` CI job, so the artifact contract job only runs the Rust contract test.

## Review Fixes

- Uses SHA-256 for artifact manifest IDs, CRC32C for file/chunk `checksum` fields, and no `checksum_algorithm` field.
- Adds `chunk_size` to `GetArtifactManifestHeaderResponse` so targets can reconstruct and verify the sealed manifest after fetching chunk pages.
- Bounds artifact byte `chunk_size` at 4 GiB in Rust and Python builders to avoid huge contiguous transfer-buffer allocations.
- Aligns Rust and Python directory manifest construction by sorting globally on the canonical manifest path before assigning `file_index` and chunk order. Added a prefix-collision regression fixture (`sub.txt` plus `sub/inner.bin`) on both sides.
- Adds a pinned in-memory artifact manifest ID test in Rust and Python so canonical JSON/hash behavior is cross-checked without depending on temp directory paths.
- Caps `GetArtifactManifestChunks` to a bounded chunk-metadata page size in both Python and Rust. The page is the number of chunk metadata records, not the artifact byte chunk size; `1024` keeps metadata responses bounded while avoiding one RPC per transfer chunk.
- Documents that `artifact_id` is currently install-path-specific because manifest file paths are canonical absolute install paths and are part of the sealed payload.
- Documents `next_page_token` as the decimal string form of the next `start_chunk_index`.
- Documents WorkerService artifact APIs, artifact source summaries, DeepGEMM source type, and the v1 tradeoff that the header returns the full file table while chunk metadata is paged.
- Represents empty files only in the artifact file table, with no zero-length transfer chunk, so the metadata contract matches the follow-up NIXL transfer protocol.
- Skips unsupported Python artifact symlinks during manifest construction instead of failing source publication.
- Makes `ArtifactSource` tensor count handling explicit instead of relying on the legacy tensor fallback.
- Ensures Python artifact manifest fetch helpers close gRPC channels on RPC errors by using `grpc.insecure_channel(...)` as a context manager.
- Fixes Rust artifact chunk pagination so `max_chunks == 0` uses the server default instead of returning a non-progressing empty page.

## CI Fixes

- Fixed the all-features clippy failure in the in-memory backend test fixture by defaulting the newly added `SourceIdentity` fields and avoiding explicit initialization of the deprecated legacy tensor field.
- Removed the duplicate Python test setup from the artifact contract CI job because `python-test` already runs `pytest tests/ -v`.

## E2E Context

- Separately validated the follow-up direction with a file artifact NIXL transfer E2E on the nscale cluster.
- That NIXL byte-transfer implementation is intentionally not part of this PR; this PR provides the manifest discovery contract that the transfer path will consume.

## Validation

- `git diff --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test -p modelexpress-common artifact_manifest --lib`
- `cargo test -p model-express-workspace-tests --test artifact_transfer_contract`
- `.venv/bin/python -m pytest tests/test_artifact_manifest.py -v`


modelexpress_client/python/modelexpress/metadata/worker_server.py
```diff
@@ -60,9 +61,96 @@ def GetTensorManifest(self, request, context):
...
+        )
+        start = request.start_chunk_index
+        if start > len(manifest.chunks):
+            context.abort(
+                grpc.StatusCode.INVALID_ARGUMENT,
+                f"start_chunk_index {start} exceeds chunk_count {len(manifest.chunks)}",
+            )
+        max_chunks = request.max_chunks or _DEFAULT_ARTIFACT_CHUNK_PAGE_SIZE
+        end = min(start + max_chunks, len(manifest.chunks))
+        response = p2p_pb2.GetArtifactManifestChunksResponse(
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in the current head. `GetArtifactManifestChunks` now caps client-requested pages to `_ARTIFACT_CHUNK_METADATA_PAGE_SIZE`; the constant comment also clarifies this is a count of chunk metadata records, not the artifact byte chunk size. Added `test_worker_service_defaults_and_caps_chunk_pages` to cover both default and oversized requests.

modelexpress_client/python/modelexpress/metadata/worker_server.py
```diff
@@ -139,3 +230,54 @@ def fetch_tensor_manifest(
...
+    """Fetch a sealed artifact manifest header directly from a source worker."""
+    channel = grpc.insecure_channel(endpoint)
+    stub = p2p_pb2_grpc.WorkerServiceStub(channel)
+    request = p2p_pb2.GetArtifactManifestHeaderRequest(
+        mx_source_id=mx_source_id,
+        artifact_id=artifact_id,
+    )
+    response = stub.GetArtifactManifestHeader(request, timeout=timeout)
+    response_bytes = response.ByteSize()
+    channel.close()
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in the current head. Both `fetch_artifact_manifest_header` and `fetch_artifact_manifest_chunks` now use `with grpc.insecure_channel(endpoint) as channel`, so the channel is closed on success and exception paths.

modelexpress_common/src/artifact_manifest.rs
```diff
@@ -0,0 +1,579 @@
...
+                self.manifest.chunks.len()
+            );
+        }
+        let max =
+            usize::try_from(max_chunks).context("artifact manifest page size exceeds usize")?;
+        let end = start.saturating_add(max).min(self.manifest.chunks.len());
+        let next_page_token = if end < self.manifest.chunks.len() {
+            end.to_string()
+        } else {
+            String::new()
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in the current head. `to_chunks_response(..., max_chunks=0)` now normalizes zero to `ARTIFACT_CHUNK_METADATA_PAGE_SIZE` before converting to `usize` and computing `end`, so pagination makes forward progress and `next_page_token` advances. Added `chunks_response_uses_default_page_size_for_zero_max_chunks` to cover this path.

zhengluo-nv:
Addressed the review follow-ups in the amended single commit:

- Matched Rust/Python manifest ordering by sorting on canonical manifest path before assigning file/chunk indices.
- Added prefix-collision and pinned cross-language artifact ID tests.
- Capped Rust chunk metadata pagination to the same bounded page size as Python.
- Documented the current install-path-specific `artifact_id` semantics and decimal `next_page_token`.

Validation passed: Rust artifact manifest tests, artifact transfer contract test, all-features clippy, diff check, and Python artifact manifest tests.

modelexpress_common/src/artifact_manifest.rs
```diff
@@ -0,0 +1,717 @@
...
+
+impl ArtifactManifest {
+    pub fn from_directory(
+        root: impl AsRef<Path>,
+        chunk_size: u64,
+        mx_source_type: i32,
+    ) -> Result<Self> {
+        if chunk_size == 0 {
+            bail!("artifact manifest chunk_size must be greater than zero");
+        }
```
nv-hwoo, line 68: Right now we only check for lower bound for chunk size. Should we add one for an upper  bound as well? 

Because chunk_size is caller-provided, a too-large chunk size on a large artifact can request a huge contiguous allocation during manifest construction. That can OOM the source process before metadata is published. This is especially risky because the PR is laying the contract for file-backed cache artifacts, which may contain large files.

modelexpress_client/python/modelexpress/metadata/artifact_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+ARTIFACT_MANIFEST_VERSION = 1
+
+
+def build_artifact_manifest(
+    root: str | Path,
+    chunk_size: int,
+    mx_source_type: int,
+) -> p2p_pb2.ArtifactManifest:
+    if chunk_size <= 0:
+        raise ValueError("artifact manifest chunk_size must be greater than zero")
```
nv-hwoo, line 30: Same as above

modelexpress_common/proto/p2p.proto
```diff
@@ -256,6 +318,67 @@ message GetTensorManifestResponse {
...
 
+message GetArtifactManifestHeaderRequest {
+  // mx_source_id for validation (catches stale discovery)
+  string mx_source_id = 1;
+
+  // Optional artifact_id from ArtifactSourceMetadata for validation.
+  string artifact_id = 2;
+}
+
+message GetArtifactManifestHeaderResponse {
```
nv-hwoo, line 331: Do we want to include `chunk_size` from the `ArtifactManifest` in the response as well so that we can verify the manifest ID later?

modelexpress_common/proto/p2p.proto
```diff
@@ -219,16 +280,17 @@ message WorkerMetadata {
...
-// Sources start this service when MX_P2P_METADATA=1. Targets call
-// GetTensorManifest to fetch tensor descriptors directly from the
-// source worker instead of from the central metadata server.
+// Per-worker gRPC service for direct manifest retrieval. Sources start this
+// service when MX_P2P_METADATA=1. Targets call GetTensorManifest for tensor
+// descriptors or GetArtifactManifestHeader/GetArtifactManifestChunks for sealed
+// file-backed artifact manifests.
 service WorkerService {
   rpc GetTensorManifest(GetTensorManifestRequest) returns (GetTensorManifestResponse);
+  rpc GetArtifactManifestHeader(GetArtifactManifestHeaderRequest) returns (GetArtifactManifestHeaderResponse);
```
nv-hwoo, line 294: Maybe not a blocker for this stage, but do you see any potential issues with returning a single manifest header that returns all file lists at once? 

For example, `GetArtifactManifestHeaderResponse` containing 70,000 small files with long absolute paths serialized to 7,053,571 bytes, which is above the common/default gRPC 4 MiB receive limit.

modelexpress_client/python/pyproject.toml
nv-hwoo, line 1: I think you are missing updated `uv.lock` file with the newly added dependency

modelexpress_common/proto/p2p.proto
nv-hwoo, line 1: Doc needs an update: Public docs still describe WorkerService as tensor-only

Suggested fix:
- Update docs/ARCHITECTURE.md WorkerService table to include GetArtifactManifestHeader and GetArtifactManifestChunks.
- Add a WorkerService subsection in docs/metadata.md with request/response semantics, artifact_id, file table behavior, chunk paging, and next_page_token format.
- Update Redis and Kubernetes examples to show artifact source summaries.
- Update docs/DEPLOYMENT.md to describe tensor and artifact manifest metadata separately from byte transfer.

modelexpress_common/proto/p2p.proto
```diff
@@ -44,6 +44,7 @@ enum MxSourceType {
   MX_SOURCE_TYPE_CUDA_GRAPH = 2;          // CUDA graph artifacts
   MX_SOURCE_TYPE_TORCH_COMPILE_CACHE = 3; // torch.compile cache files
   MX_SOURCE_TYPE_TRITON_CACHE = 4;        // Triton cache files
+  MX_SOURCE_TYPE_DEEP_GEMM_CACHE = 5;     // DeepGEMM kernel cache files
```
nv-hwoo, line 47: DeepGEMM source type and artifact identity fields are missing from metadata docs

modelexpress_client/python/modelexpress/metadata/artifact_manifest.py
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+
+def artifact_manifest_id(manifest: p2p_pb2.ArtifactManifest) -> str:
+    canonical = json.dumps(
+        _manifest_to_dict(manifest),
+        ensure_ascii=False,
+        separators=(",", ":"),
+    ).encode()
+    return hashlib.sha256(canonical).hexdigest()
```
nv-hwoo, line 75: PR body conflicts with the implemented hashing contract:


The PR body says:

```text
Uses CRC32C for artifact manifest IDs and file/chunk checksums, advertises `checksum_algorithm = "crc32c"`, and names manifest file/chunk fields `checksum` instead of `sha256`.
```

The actual implementation uses:

- SHA-256 for `artifact_id` over canonical manifest JSON.
- CRC32C for file and chunk `checksum` fields.
- No `checksum_algorithm` field in the proto or helpers.

The implemented contract is reasonable, but the PR body should not say artifact IDs are CRC32C or that a `checksum_algorithm` field exists.

modelexpress_common/src/artifact_manifest.rs
zhengluo-nv: Added a upper bound of 4 GB.

modelexpress_client/python/modelexpress/metadata/artifact_manifest.py
zhengluo-nv: Added an upper bound of 4 GB.

modelexpress_common/proto/p2p.proto
zhengluo-nv: Updated

modelexpress_client/python/pyproject.toml
zhengluo-nv: Updated

modelexpress_common/proto/p2p.proto
zhengluo-nv: I will use tar to bundle those small files into one file, which is also optimal for transfering

modelexpress_common/proto/p2p.proto
zhengluo-nv: Updated the doc

modelexpress_common/proto/p2p.proto
zhengluo-nv: Updated the doc

modelexpress_client/python/modelexpress/metadata/artifact_manifest.py
zhengluo-nv: Will update PR body

modelexpress_client/python/pyproject.toml
zhengluo-nv: That crc32c lib is already in the uv.lock

modelexpress_common/proto/p2p.proto
zhengluo-nv: In that case the manifest only stores one file which is the tar file, and the tar will be extracted to the target worker's path as is

modelexpress_common/src/artifact_manifest.rs
```diff
@@ -0,0 +1,727 @@
...
+        let path = entry.path();
+        let file_type = entry
+            .file_type()
+            .with_context(|| format!("failed to inspect artifact path {}", path.display()))?;
+        if file_type.is_symlink() {
+            bail!(
+                "artifact manifest does not support symlink {}",
+                path.display()
+            );
+        }
```
nv-hwoo, line 288: One more small comment: There seems to be a slight mismatch between the Python and Rust on handling symlinks.

In Python `modelexpress_client/python/modelexpress/metadata/artifact_manifest.py:94-96`, it skips the symlinks
```python
if path.is_symlink():
    logger.warning("Skipping artifact symlink %s", path)
    continue
```
while in the Rust code here, it's directly rejecting it. This creates a small cross-language semantic drift, so I would match both to skip the symlinks (or both reject it).

zhengluo-nv: Sounds good. Updated the Rust side to skip symlink. This change was supposed to be in the next PR

nv-hwoo (approved):
LGTM

PR #424, opened 2026-06-05
feat: Kubernetes ServiceAccount auth for gRPC

wseaton:
Opt-in ServiceAccount auth for the gRPC server. Off by default; existing deployments are unaffected.

- AuthN: verifies the caller's projected ServiceAccount token via the Kubernetes TokenReview API, in-process (no sidecar or mesh). Requested audiences must overlap the reviewed token's audiences.
- AuthZ: exact-match `namespace:serviceaccount` allowlist.
- Modes `off` (default) and `enforce`. `enforce` fails config validation without a token audience and a non-empty allowlist; the Helm chart fails rendering on the same misconfiguration.
- Health service stays ungated so kubelet probes need no token.
- Verified tokens and definitive rejections are cached (bounded, default 60s TTL). TokenReview backend errors return `UNAVAILABLE`, not `UNAUTHENTICATED`, and are never cached.
- Rust and Python clients attach the token when a projected token file is present and send nothing otherwise, so the same client works off-cluster against a server with auth off.
- Helm: `security.enabled` injects the server env and, with `serviceAccount.create`, binds the SA to `system:auth-delegator` for TokenReview access.

Config via `MODEL_EXPRESS_SECURITY_*` env vars or `security.*` in the config file / Helm values. Covered by unit tests and an end-to-end test driving the real client against a real tonic server.

## Summary by CodeRabbit

* **New Features**
  * Added optional Kubernetes ServiceAccount authentication for gRPC services, with `off` vs `enforce` modes.
  * Configurable token audiences, allowed service account allowlist, and server-side review result caching (TTL).
  * Python and Rust clients now automatically attach bearer tokens from projected ServiceAccount files when present.
  * Helm charts now enable auth settings and add required RBAC for TokenReview delegation.
* **Documentation**
  * Added ServiceAccount Authentication deployment and client guidance, including relevant environment variables and Helm examples.
* **Tests**
  * Expanded unit coverage for token reading, refresh, TTL/env handling, and authorization metadata.
  * Added end-to-end authentication test validating success, deny, and unauthenticated paths.


wseaton:
Smoke-tested live on a Kubernetes cluster (real apiserver TokenReview, `enforce` mode, allowlist `weaton-dev:mx-auth-client`, audience `modelexpress`). Three probe pods calling a gated RPC (`modelexpress-cli api send ping`):

| Probe | Result |
|---|---|
| Allowlisted SA, projected token | `{"message":"pong"}`, exit 0 |
| No token file | `UNAUTHENTICATED`, exit 1 |
| Authenticated SA not in allowlist | `PERMISSION_DENIED`, exit 1 |

Client without a token warns and degrades as designed:

```
WARN modelexpress auth: token file not found; RPCs sent without a bearer token (server must have auth off)
ERROR Command execution failed: gRPC error: status: Unauthenticated, message: "missing or invalid service account token"
```

Server-side denials:

```
WARN modelexpress_server::auth::layer: auth denied reason=missing or invalid service account token path=/model_express.api.ApiService/SendRequest
WARN modelexpress_server::auth::layer: auth denied reason=service account weaton-dev:mx-auth-intruder is not in the allowlist path=/model_express.api.ApiService/SendRequest
```

Health stays reachable without a token, so kubelet probes are unaffected.

modelexpress_server/src/auth/layer.rs
```diff
@@ -0,0 +1,201 @@
...
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let clone = self.inner.clone();
+        let mut inner = std::mem::replace(&mut self.inner, clone);
+        let state = self.state.clone();
+
+        Box::pin(async move {
+            match state.verify(req.headers()).await {
```
nicolasnoble, line 71: Auth runs once when the stream opens, so a long-lived server-streaming RPC keeps flowing even after the token expires or the SA is revoked. Fine for a first cut. Might be worth a doc line noting auth is enforced at RPC start, not per message.

modelexpress_server/src/auth.rs
```diff
@@ -0,0 +1,396 @@
...
+                .max_capacity(Self::TOKEN_CACHE_MAX_ENTRIES)
+                .build(),
+        }
+    }
+
+    pub(crate) async fn verify(&self, headers: &HeaderMap) -> Result<CallerIdentity, Denial> {
+        let token = extract_bearer(headers).ok_or(Denial::Unauthenticated)?;
+        let key = TokenCacheKey::new(token.expose_secret());
+
+        if self.negative_cache.get(&key).await.is_some() {
```
nicolasnoble, line 115: The negative cache is keyed on the raw token, so a caller cycling unique bad tokens skips it and every one lands a fresh TokenReview on the apiserver. Fine for an opt-in feature. If you ever want to harden it, a global or per-source TokenReview rate limit would cap the amplification.

nicolasnoble (approved):
Looks solid, a few non-blocking notes inline.

One cross-cutting thing worth handling before merge: with auth on but no TLS, the projected token rides on the wire in cleartext. The Python client attaches it to an insecure channel, so it's capturable and replayable until the token expires. Can we get a clear "enable auth only over TLS" warning in the deployment docs, and maybe a client-side warn when a token is attached to an insecure channel? Fine to merge without the client guard, but I'd like the docs warning in.

modelexpress_server/src/auth.rs
wseaton: Sounds good, happy do a follow-up PR for this.

modelexpress_server/src/auth/layer.rs
wseaton: Added doc comments

zhengluo-nv:
/ok to test [4898067](https://github.com/ai-dynamo/modelexpress/pull/424/commits/4898067d02ca5ae21d8a054863d711cad04e14b0)

zhengluo-nv (approved):

wseaton:
@coderabbitai review

athreesh:
thanks @wseaton !!

PR #425, opened 2026-06-05
fix(ci): avoid user-controlled data in run blocks for vCluster actions

nv-hwoo:
SonarQube flagged three composite actions for interpolating action inputs
directly into shell `run:` blocks via `${{ }}`.

Fix: move each untrusted input into the step's `env:` block and reference it
as a normal shell variable, so the shell treats the value as literal data.

- `check-vcluster-exists`: `kubeconfig_base64` moved to an env var.
- `install-vcluster-cli`: `vcluster_version` moved to an env var (both the
  log-group echo and the release download URL).
- `setup-mx-vcluster`: `kubeconfig_base64` and `vcluster_k8s_version` moved to
  env vars; also replaced `${{ github.workspace }}` inside run blocks with the
  built-in `$GITHUB_WORKSPACE` so no run block expands a `${{ }}` expression.
  The `env:`-block `KUBECONFIG` value keeps `${{ github.workspace }}` since the
  runner must expand it there.

## Summary by CodeRabbit

* **Chores**
  * Improved GitHub Actions workflow configuration for internal CI/CD processes, including refinements to kubeconfig handling, vCluster CLI installation, and environment variable management.

### DIFF
--- a/.github/actions/check-vcluster-exists/action.yml
+++ b/.github/actions/check-vcluster-exists/action.yml
@@ -32,13 +32,14 @@ runs:
       env:
         VCLUSTER_NAME: ${{ inputs.vcluster_name }}
         NAMESPACE: ${{ inputs.vcluster_namespace }}
+        KUBECONFIG_BASE64: ${{ inputs.kubeconfig_base64 }}
       run: |
         # Use a unique temp file so concurrent runs on the same self-hosted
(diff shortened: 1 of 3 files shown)


tanushriya910 (commented):
Changes look good but looks like one of the CI checks is failing although it doesn't look like its caused by this PR

AndyDai-nv (approved):

PR #430, opened 2026-06-09
ci: add SGLang P2P test over Mooncake TransferEngine transport

AndyDai-nv:
## What

Adds a CI P2P weight-transfer test for **SGLang over the Mooncake TransferEngine transport** (`transport=transfer_engine`), mirroring the existing SGLang NIXL test (`test-p2p` row 3) but routing the transfer through SGLang's Mooncake TransferEngine instead of NIXL. mx-server still coordinates discovery via `transfer_engine_session_id` metadata.

## Changes

- **`ci/k8s/client/sgl/Dockerfile.mooncake`** — same ModelExpress-client overlay as the NIXL image, plus a build-time canary that fails the build if the base SGLang lacks the `transfer_engine` hook.
- **`ci/k8s/client/sgl/manifest-azure-mooncake.yaml`** — mirrors `manifest-azure.yaml`; seed/client branch on `JOB_NAME` because the launch args differ (seed: `--remote-instance-weight-loader-start-seed-via-transfer-engine`; client: `--load-format remote_instance --remote-instance-weight-loader-backend modelexpress`).
- **`conftest.py` / `test_p2p_k8s.py`** — add `--transport`; skip the NIXL-only per-rank-agent assertion for `transfer_engine`. Transfer marker is the loader's `Receiving N tensors via TransferEngine` log line.
- **`modelexpress-ci-tests.yml`** — add `sglang-mooncake` to `build-client` and `test-p2p`.
- **`TEST_PLAN.md`** — row 3.1.

## Known unknowns (expected to surface on the first run)

1. The base SGLang image in `Dockerfile.mooncake` currently reuses the NIXL variant's pin and is **not yet confirmed** to carry the transfer_engine hook. The build-time canary fails loudly with a "repin `SGLANG_IMAGE`" message if it's absent.
2. Whether Mooncake initializes at TP=1 on a single-GPU `a100a` node, and whether it needs `MOONCAKE_DEVICE` pinned to that node's IB NIC (left to auto-discovery for now, following the NIXL variant's `MX_RDMA_NIC_PIN=auto` philosophy).

## Summary by CodeRabbit

* **Chores**
  * Extended CI infrastructure to support SGLang with Mooncake TransferEngine as an alternative P2P weight-transfer transport mechanism.
  * Added containerization and Kubernetes manifests for new transport variant testing on Azure A100 GPUs.
  * Implemented configurable transport backend selection in test suite to support multiple transfer strategies.


.github/workflows/modelexpress-ci-tests.yml
tanushriya910: you can remove this max-parallel: 3 constraint so all 4 jobs can run in parallel

ci/k8s/server/mx-server.yaml
```diff
@@ -42,6 +42,20 @@ spec:
...
+            # unaffected. But SGLang's in-tree (v0.5.12+) transfer_engine source
+            # publishes READY exactly once and never heartbeats, so a target
+            # that cold-starts slowly (model init + flashinfer + TransferEngine
+            # init can take >90s to reach the weight-load step) would otherwise
+            # find the source already reaped to STALE -> "no READY source found".
+            # 900s matches pod_ready_timeout_seconds: the source stays READY for
+            # as long as the test is willing to wait for the target anyway.
+            # The stale-metadata test uses mx-server-redis.yaml (its own short
+            # reaper knobs), so this does not weaken that coverage.
+            - name: MX_HEARTBEAT_TIMEOUT_SECS
```
tanushriya910, line 60: > so a slow-cold-starting target was finding the source already reaped to STALE past the 90s default; mx-server.yaml now sets `MX_HEARTBEAT_TIMEOUT_SECS=900` to cover cold-start.

but wouldn't this be an issue for other inference servers that spin up after 900s timeout since they won't see any non-stale sources to do P2P transfer from?

.github/workflows/modelexpress-ci-tests.yml
AndyDai-nv: Heads-up before dropping it: in this PR's own run, `P2P TP test (vllm TP=2)` failed once with `FailedScheduling: Insufficient nvidia.com/gpu` (the target sat Pending the full 900s) while test-p2p was already capped at 3 and running alongside test-tp + the dynamo jobs; the rerun passed once GPUs freed up. Each TP=1 P2P entry holds 2 GPUs (source + target on separate hosts), so removing the cap takes peak test-p2p demand from 6 -> 8 concurrent GPUs, on top of test-tp (4) and dynamo. Happy to remove it if the a100a pool has that headroom, or bump to 4 and watch — but flagging the contention we just hit this run. WDYT?

ci/k8s/server/mx-server.yaml
AndyDai-nv: Good question — this manifest is CI-only: run-mx-p2p-test deploys it into a per-test namespace and tears it down at the end, and the 900s equals that test's own `pod_ready_timeout_seconds`, so the source stays READY for exactly as long as the harness will wait for the target — they can't drift apart within a run. It's not a production setting.

The reason a bump is needed at all is that SGLang's in-tree `transfer_engine` source publishes READY once and never heartbeats, unlike the ModelExpress `MxModelLoader` path (used by vllm/trt/sglang-nixl) which refreshes a HeartbeatThread well inside 90s. A real inference server using the MxModelLoader path keeps heartbeating, so it stays non-stale regardless of when it spun up. So this only affects the SGLang native-transfer_engine seed, and only in CI cold-start timing. I can file the SGLang no-heartbeat gap separately if useful.

tanushriya910: right, I got the reasoning behind why it was updated. but I was thinking more in lines for customers using this same setting in production cause this seems like a real gap. yeah we can create a separate issue for that.

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: oh that one needs a separate pool of non-MIG GPUs which we don't have as many. For your test case as long as its a100a pool it should be fine

ci/TEST_PLAN.md
```diff
@@ -12,6 +12,7 @@ Coverage matrix for the ModelExpress CI infrastructure. Reflects what's currentl
...
...e --remote-instance-weight-loader-backend modelexpress` + config `{"transport": "transfer_engine"}`. mx-server coordinates via `transfer_engine_session_id` metadata. Transfer marker is the loader's `Receiving N tensors via TransferEngine` log; the NIXL-only per-rank-agent assertion is skipped via `--transport transfer_engine`. First PR run (commit d3688d6) failed the `sglang-mooncake` build: the canary caught that the NIXL base predates the transfer_engine loader. Base repinned to `lmsysorg/sglang:v0.5.12.post1-cu129` (the hook landed in v0.5.10; cu129 keeps it in the `nixl[cu12]` family) and `mooncake-transfer-engine` is now pip-installed in the image | Bring-up findings: Mooncake initializes fine at TP=1 on a single-GPU a100a node (auto-discovers the IB HCAs, `MOONCAKE_DEVICE` not needed). The seed publishes READY but SGLang's in-tree transfer_engine source publishes once and never heartbeats (unlike the ModelExpress MxModelLoader path), so a slow-cold-starting target was finding the source already reaped to STALE past the 90s default; mx-server.yaml now sets `MX_HEARTBEAT_TIMEOUT_SECS=900` to cover cold-start. Still validating the end-to-end transfer + inference after that fix |
```
tanushriya910: >  Still validating the end-to-end transfer + inference after that fix

I think document needs update

.github/workflows/modelexpress-ci-tests.yml
tanushriya910: non blocking comment, we can change this later as well

ci/TEST_PLAN.md
AndyDai-nv: Updated — the test now passes end-to-end, so I flipped row 3.1 from "In CI — unverified" to "In CI", dropped the "still validating" line, and also fixed a stale bit in there (the marker text said "Receiving N tensors via TransferEngine" but the actual assertion uses SGLang's client-side "ModelExpress: weight transfer complete"). The SGLang native-source no-heartbeat gap is noted as tracked separately.

.github/workflows/modelexpress-ci-tests.yml
AndyDai-nv: Thanks for confirming the a100a pool has the headroom — dropped the `max-parallel: 3` cap so all test-p2p framework entries run concurrently, and updated the comment to note the scarce non-MIG pool that constrains test-tp is a separate concern. Both in a8a28fe.

ci/k8s/client/sgl/manifest-azure-mooncake.yaml
```diff
@@ -0,0 +1,173 @@
...
+            - name: MODEL_EXPRESS_URL
+              value: "mx-server:8000"
+            - name: MX_SERVER_ADDRESS
+              value: "mx-server:8000"
+            - name: MX_RDMA_NIC_PIN
+              value: "auto"
+            - name: UCX_RNDV_SCHEME
+              value: "get_zcopy"
+            - name: UCX_RNDV_THRESH
+              value: "0"
```
tanushriya910, line 144: nit: are these needed for transfer engine mode

tanushriya910 (approved):

PR #433, opened 2026-06-10
feat: add NIXL artifact transfer protocol

zhengluo-nv:
## Summary

Adds a NIXL-backed file artifact transfer path for sealed cache artifacts, with a shared Python helper interface for the cache use cases we want to support:

- `torch_compile_cache_artifact_transfer()`
- `triton_cache_artifact_transfer()`
- `deep_gemm_cache_artifact_transfer()`

The tar-backed implementation packages a cache directory into one uncompressed tar file before publishing the artifact manifest. Targets transfer the tar through NIXL into their own target-local bundle path, then call `install()` to untar into the runtime cache directory. This keeps the published manifest sealed while avoiding any requirement that source and target share the same absolute staging path.

This PR intentionally adds the helper methods, transfer protocol, tests, and docs only. Loader integration is left to a follow-up PR.

## What Changed

- Adds artifact manifest transfer RPCs on `WorkerService`:
  - `GetArtifactManifestHeader`
  - `GetArtifactManifestChunks`
  - `PrepareArtifactChunk`
  - `ReleaseArtifactChunk`
- Adds `P2PArtifactTransfer` and tar-backed cache artifact helpers.
- Adds NIXL DRAM receive support for file-backed artifact chunks.
- Uses CRC32C for file and chunk checksums.
- Publishes lightweight artifact source metadata through the existing MX metadata backends.
- Supports target-local artifact staging before install.
- Adds focused Python tests for manifest building, chunk transfer, tar install safety, source publication/discovery, lifecycle cleanup, and target-local bundle paths.
- Updates metadata, architecture, deployment, and proto comments for the artifact transfer flow.

## E2E Validation

Tested on nscale in namespace `zheng` using a current-worktree no-deps client image:

`nvcr.io/nvidian/dynamo-dev/model-express-dev-containers:client-artifact-pr3-8d1c3ea-pathfix-e2e`

Image digest:

`sha256:3c62fd8f660f817d51832e05b04d7a6ba7f30efa336a79daf27cd592c73782bb`

### Redis Backend

Source:

- Fixture: `20,000` files, `2,147,483,648` bytes
- Fixture create: `12.820s` (not counted in transfer timing)
- `prepare_source()`: `2.488s`
- Bundle: `1` file, `33` chunks, `2,160,773,120` bytes
- Source bundle path: `/artifact-cache/source-bundle/artifact.tar`
- Source status reached `READY`

Target:

- Discovery wait after source READY: `0.060s`
- Transfer: `1.564s`
- Install / untar: `1.657s`
- Transfer + install: `3.221s`
- Exact relative path validation: `true`
- Installed files: `20,000`
- Installed bytes: `2,147,483,648`
- Target bundle path: `/artifact-cache/target-bundle/artifact.tar`

Note: the first Redis target job started before the server/source were reachable and failed with gRPC `UNAVAILABLE`; the retry after source readiness completed successfully. Transfer timing above is from the completed retry.

### Kubernetes CRD Backend

Source:

- Fixture: `20,000` files, `2,147,483,648` bytes
- Fixture create: `13.337s` (not counted in transfer timing)
- `prepare_source()`: `2.535s`
- Bundle: `1` file, `33` chunks, `2,160,773,120` bytes
- Source bundle path: `/artifact-cache/source-bundle/artifact.tar`
- Source status reached `READY`

Target:

- Discovery wait: `16.678s` (target started before source READY)
- Transfer: `1.636s`
- Install / untar: `1.752s`
- Transfer + install: `3.388s`
- Exact relative path validation: `true`
- Installed files: `20,000`
- Installed bytes: `2,147,483,648`
- Target bundle path: `/artifact-cache/target-bundle/artifact.tar`

CRD metadata check:

- `ModelMetadata` Ready condition: `True`
- `status.worker.artifactSource.artifactId`: present
- `status.worker.artifactSource.fileCount`: `1`
- `status.worker.artifactSource.chunkCount`: `33`
- `status.worker.artifactSource.totalSize`: `2160773120`

## Local Validation

```bash
uv run --project modelexpress_client/python --extra dev pytest \
  modelexpress_client/python/tests/test_artifact_transfer.py \
  modelexpress_client/python/tests/test_artifact_manifest.py
```

Result: `34 passed`

```bash
git diff --check
```

Result: clean

## Release Notes

* **New Features**
  * Added P2P artifact transfer support for file-backed caches (torch compile, Triton, DeepGEMM), enabling efficient peer-to-peer distribution of compiled artifacts across nodes.
  * Made artifact chunk size configurable via `MX_ARTIFACT_TRANSFER_CHUNK_SIZE` environment variable for tuning transfer performance.

* **Documentation**
  * Updated coordination protocol and deployment guidance for artifact transfer workflows.


modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
```diff
@@ -0,0 +1,998 @@
...
+    def _expire_leases_locked(self) -> None:
+        now = time.monotonic()
+        expired = [
+            lease_id
+            for lease_id, lease in self._leases.items()
+            if lease.expires_at <= now
+        ]
+        for lease_id in expired:
+            lease = self._leases.pop(lease_id)
+            self._free_slots.append(lease.slot_index)
```
AndyDai-nv, line 634: Lease expiry returns the slot straight to `_free_slots` with no fencing that the remote NIXL READ on that buffer has finished. The default lease TTL (300s) is larger than the default transfer timeout (120s), so a single healthy transfer won't hit this. But if a transfer ever exceeds the TTL (slow link, large chunk, a target that stalls after `PrepareArtifactChunk`), the source can hand this same DRAM buffer to another `prepare()` and overwrite bytes that a remote agent is still reading. The target's per-chunk CRC32C catches the corruption, so it surfaces as an intermittent transfer failure rather than bad data — but it's hard to diagnose.

Could we isolate an expired slot instead of immediately recycling it (or make the TTL configurable and document that it must comfortably exceed the worst-case transfer time)?

modelexpress_client/python/modelexpress/metadata/worker_server.py
```diff
@@ -64,6 +67,97 @@ def GetTensorManifest(self, request, context):
...
+            )
+        except OSError as exc:
+            context.abort(
+                grpc.StatusCode.INTERNAL,
+                f"failed to prepare artifact chunk {chunk.chunk_index}: {exc}",
+            )
+        except ValueError as exc:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, str(exc))
+        except RuntimeError as exc:
+            context.abort(grpc.StatusCode.RESOURCE_EXHAUSTED, str(exc))
```
AndyDai-nv, line 108: This `RESOURCE_EXHAUSTED` exposes an implicit coupling between three values that nothing validates:
- target batch concurrency = `max_inflight_chunks` (default 4)
- source DRAM buffers = `max_buffers` passed at publish (default 4)
- the worker gRPC thread pool, hardcoded `max_workers=4` in `WorkerGrpcServer.start()`

They all happen to be 4, so the default path works. But if a target sets `max_inflight_chunks > 4`, the source either runs out of buffers and aborts here with `RESOURCE_EXHAUSTED` — and the target doesn't retry, since `_prepare_chunk_batch` raises `first_error` and the whole transfer fails — or the prepares just queue behind the 4-thread pool and lose the intended concurrency.

Two suggestions: (1) derive the gRPC thread-pool size from config / `max_inflight` rather than hardcoding 4, and (2) have the target back off and retry on `RESOURCE_EXHAUSTED` instead of failing the transfer, so source and target concurrency don't have to be hand-matched.

modelexpress_client/python/tests/test_artifact_manifest.py
```diff
@@ -229,6 +254,213 @@ def test_worker_service_defaults_and_caps_chunk_pages():
...
+        tmp_path,
+        chunk_size=3,
+        mx_source_type=p2p_pb2.MX_SOURCE_TYPE_TORCH_COMPILE_CACHE,
+    )
+
+    assert [(file.path, file.size) for file in manifest.files] == [
+        ((tmp_path / "empty.bin").resolve().as_posix(), 0)
+    ]
+    assert list(manifest.chunks) == []
+
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in `442c013` by removing the duplicate second definition of `test_build_artifact_manifest_keeps_empty_file_without_transfer_chunk`.

Validation: focused artifact tests passed (`35 passed`) and `git diff --check` is clean.

modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
```diff
@@ -0,0 +1,1006 @@
...
+            logger.warning(
+                "Expired artifact chunk lease %s for chunk %d; released buffer slot %d",
+                lease_id,
+                lease.chunk.chunk_index,
+                lease.slot_index,
+            )
+
+
+def transfer_artifact_from_worker(
```
nv-hwoo, line 643: I think we need target-side validation of the fetched artifact manifest before writing. Today the target verifies each received chunk CRC, but it still does not validate the complete artifact contract: expected `artifact_id`, chunk page identity, `chunk_count`, chunk indexes, range bounds, overlap, or full per-file coverage.

Because `_prepare_target_files()` pre-sizes files, missing or partial coverage can still look successful with zero-filled holes. Can we reconstruct the manifest from the header plus fetched chunks, compare `artifact_manifest_id()`, verify complete non-overlapping coverage, and then verify final file size/CRC after writes?

modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
zhengluo-nv: For your concern on the inflight transfer case, We don’t have source-side visibility into whether the target NIXL READ is still in flight, but the target verifies every received chunk against the manifest CRC32C before writing it. If a TTL-expired buffer is reused and the target reads overwritten bytes, the checksum mismatch aborts the artifact transfer and cleans up partial files. So if this happens, it will fail the transfer and fallback to generating cache at the runtime.

modelexpress_client/python/modelexpress/metadata/worker_server.py
zhengluo-nv: This one should be addressed. The target now treats RESOURCE_EXHAUSTED during prepare as source-side backpressure: if at least one chunk in the batch was prepared, it transfers/releases that partial batch and then continues with the remaining chunks. If zero chunks can be prepared, it retries briefly and then fails. I also added a test where the target asks for 4 inflight chunks but the source only allows 2 active leases; the transfer completes while respecting the source limit.

docs/metadata.md
nv-hwoo, line 1: The docs describe `k8s-service` as the decentralized P2P metadata backend, but `MxK8sServiceClient.get_metadata()` still only calls `GetTensorManifest` and returns `tensor_source`. `discover_artifact_source()` requires `artifact_source`, so artifact discovery does not work through this backend today.

Can we either document artifact transfer as central-backend-only for now, or add a `k8s-service` artifact discovery path using `GetArtifactManifestHeader`?

modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
```diff
@@ -0,0 +1,1009 @@
...
+            raise ValueError(
+                f"artifact bundle must contain exactly one file, got {len(header.files)}"
+            )
+        tar_file = Path(header.files[0].path).resolve(strict=True)
+        output_path = self.target_root.resolve()
+        output_path.mkdir(parents=True, exist_ok=True)
+        _validate_tar_members(tar_file)
+        _run_tar(["-xf", str(tar_file), "-C", str(output_path)])
+
+    def _target_tar_path(self) -> Path:
```
nv-hwoo, line 269: Could we reject an existing symlink at the target bundle path before calling `resolve()` or opening it? Right now, if `bundle_root/artifact.tar` already exists as a symlink, `resolve()` follows it and `_prepare_target_files()` can truncate/write the symlink target outside `bundle_root`.

I think this should use a no-follow temp file under `bundle_root`, then atomic rename. It would also be good to add a regression test where `target-bundle/artifact.tar` points to an outside file and the outside file is not truncated.

modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
```diff
@@ -0,0 +1,1009 @@
...
+        response, _ = fetch_artifact_manifest_chunks(
+            endpoint,
+            mx_source_id=mx_source_id,
+            artifact_id=artifact_id,
+            start_chunk_index=start,
+        )
+        chunks.extend(response.chunks)
+        if not response.next_page_token:
+            return chunks
+        start = int(response.next_page_token)
```
nv-hwoo: nit: `_fetch_all_chunks()` trusts `next_page_token` and feeds it back as `start` without a progress check. A buggy source can return the same token repeatedly, which makes the target loop forever while accumulating chunks in memory.

docs/DEPLOYMENT.md
nv-hwoo, line 1: `MX_ARTIFACT_TRANSFER_CHUNK_SIZE` is implemented and tested, but I do not see it in the deployment/config docs. Can we document the default, max, and memory/performance tuning implications?

modelexpress_common/proto/p2p.proto
```diff
@@ -196,7 +199,8 @@ message ArtifactManifestFile {
   // File index used by ArtifactManifestChunk.file_index.
   uint32 file_index = 1;
 
-  // Canonical absolute install path using "/" separators.
+  // Canonical absolute publisher path using "/" separators. Transfer targets
+  // may rewrite this to a target-local staging path before installation.
```
nv-hwoo: (nit) Small conflict in the comment? The top-level manifest comment says paths are install paths, while `ArtifactManifestFile.path` says publisher paths that targets may rewrite.

modelexpress_common/proto/p2p.proto
```diff
@@ -388,6 +395,78 @@ message GetArtifactManifestChunksResponse {
...
+  uint64 addr = 1;
+
+  // Prepared range length in bytes.
+  uint64 length = 2;
+
+  // NIXL device id for the prepared range. DRAM buffers use 0.
+  uint32 device_id = 3;
+
+  // NIXL memory type for prep_xfer_dlist, for example "DRAM".
+  string mem_type = 4;
```
nv-hwoo: Do we use anything other than DRAM?

modelexpress_common/proto/p2p.proto
```diff
@@ -388,6 +395,78 @@ message GetArtifactManifestChunksResponse {
...
+
+  // Metadata for the prepared chunk.
+  ArtifactManifestChunk chunk = 4;
+
+  // NIXL source descriptor for this chunk.
+  ArtifactChunkTransferDescriptor source = 5;
+
+  // Self-describing worker info, matching GetArtifactManifestHeaderResponse.
+  string metadata_endpoint = 6;
+  string agent_name = 7;
```
nv-hwoo: Should this be part of `ArtifactChunkTransferDescriptor` if it's NIXL specific?

docs/DEPLOYMENT.md
zhengluo-nv: Documented

docs/metadata.md
zhengluo-nv: Document that artifact transfer is central-backend-only

modelexpress_common/proto/p2p.proto
zhengluo-nv: Updated

modelexpress_common/proto/p2p.proto
zhengluo-nv: Probably not. To keep it simple I remove mem_type parameter and keep it DRAM only. The only extension I can imagine in the future is to use TCP instead of RDMA

modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
zhengluo-nv: Addressed. The target now validates the fetched header/chunk contract before writing: expected artifact id, file/chunk counts, chunk index order, bounds, complete non-overlapping per-file coverage, and reconstructed `artifact_manifest_id()`. Each chunk is still CRC32C-verified before write, so corrupted bytes fail before reaching the staged file.

I kept the final whole-file CRC out intentionally to avoid a second full read after transfer; the manifest coverage check plus per-chunk CRC covers the zero-filled-hole case this comment called out.

modelexpress_client/python/modelexpress/metadata/artifact_transfer.py
zhengluo-nv: Addressed. `_target_tar_path()` now rejects an existing `bundle_root/artifact.tar` symlink before transfer, and there is a regression test where `target-bundle/artifact.tar` points outside the bundle directory. The outside file is not opened/truncated.

modelexpress_common/proto/p2p.proto
zhengluo-nv: Addressed. I removed `metadata_endpoint`, `agent_name`, and `worker_rank` from `PrepareArtifactChunkResponse` instead of moving them into `ArtifactChunkTransferDescriptor`. The header already carries the worker identity, and the prepare response only needs the prepared chunk metadata, the byte-range transfer descriptor, lease id, and NIXL source metadata.

nv-hwoo (approved):

PR #436, opened 2026-06-11
test(ci): track P2P peak VRAM usage

zhengluo-nv:
## Summary

- Add a dedicated `P2P VRAM test (vllm)` CI job that compares target P2P load VRAM against source disk-load VRAM.
- Track both peak VRAM during end-to-end initialization and final VRAM after `/health` returns 200.
- Run the VRAM job on the full-GPU A100b pool with `deepseek-ai/DeepSeek-V2-Lite`, TP=2, and the existing vLLM worker image.
- Keep the existing functional P2P matrix separate from the VRAM parity test while reusing the existing TP=2 A100b manifest.

## Test Details

- VRAM sampling is enabled by setting `vram_tolerance_percent`; unset jobs keep the existing non-VRAM path.
- The comparison allows target-source delta within 1% of source VRAM.
- The dedicated VRAM pytest reuses the shared P2P transfer-marker and per-rank source-agent assertions.
- The dedicated VRAM pytest fails if target peak or final-after-initialization VRAM exceeds source beyond the configured tolerance.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/modelexpress-ci-tests.yml"); YAML.load_file("ci/k8s/client/vllm/manifest-azure-tp2.yaml"); puts "yaml ok"'`
- `python3 -m py_compile ci/k8s/client/test_p2p_vram_k8s.py ci/k8s/client/test_p2p_k8s.py ci/k8s/client/conftest.py`
- `git diff --check`

```
  {
    "source": {
      "peak_mib": 39580,
      "final_mib": 39580
    },
    "target": {
      "peak_mib": 39596,
      "final_mib": 39596
    },
    "tolerance_percent": 0.5
  }
  Target peak VRAM: 39596 MiB across 10 sample(s).
  Target final VRAM after /health: 39596 MiB.
```


.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -416,6 +416,66 @@ jobs:
             kill "${VCLUSTER_PF_PID}" 2>/dev/null || true
           fi
 
+  # ---------------------------------------------------------------------------
+  # Dedicated P2P VRAM parity test.
+  # Runs on the full-GPU A100b pool with TP=2 and a DeepSeek MLA + MoE model so
+  # the VRAM signal is meaningful and separate from the functional framework
+  # P2P matrix.
+  # ---------------------------------------------------------------------------
+  test-p2p-vram:
```
tanushriya910: you probably want to include it under the umbrella ci-status-check otherwise this failing test won't block PR merges to main: https://github.com/ai-dynamo/modelexpress/blob/main/.github/workflows/modelexpress-ci-tests.yml#L721

zhengluo-nv: Updated

.github/actions/run-mx-p2p-test/action.yml
```diff
@@ -296,39 +330,52 @@ runs:
         # ships curl (vllm/vllm-openai + sglang base images both do).
         deadline=$((SECONDS + TIMEOUT))
         TARGET_POD=""
-        while [ $SECONDS -lt $deadline ]; do
-          # Refresh pod name each iteration — covers the Pending → Running
-          # transition where the pod name doesn't exist yet at first.
-          TARGET_POD=$(kubectl get pods -n ${NAMESPACE} \
-            -l job-name=mx-target \
-            -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
+        if [ -n "${VRAM_TOLERANCE_PERCENT}" ]; then
```
tanushriya910: nit: we can actually assert this for all p2p tests and remove the if else branching so its cleaner as well

.github/actions/run-mx-p2p-test/vram-sampling.sh
```diff
@@ -0,0 +1,107 @@
...
+      -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)"
+  fi
+  if [ -z "${pod}" ]; then
+    return 0
+  fi
+
+  local sample output err err_file
+  err_file="$(mktemp)"
+  output="$(kubectl exec -n "${NAMESPACE}" "${pod}" -c "${container}" -- \
+      nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits 2>"${err_file}" \
```
tanushriya910: question: so this will include the kv cache allocation as well not just weights, just wondering if vllm allocates a fixed percentage of GPU memory for weights + kv cache then that might hide weight transfer discrepancies?

zhengluo-nv: I think so, as long as the input is the same, the allocated VRAM should be the same. I tested a few times and always see 0 diff between source and target

tanushriya910: right, so the concern I have is are we really testing if RDMA path used more weight memory compared to source. so if the target duplicated/leaked weights, the KV cache just shrinks to compensate and you'd still see 0 diff thereby defeating the point of this test.

zhengluo-nv: Make sense. vLLM can set a fixed size of KV cache, will do that instead

.github/actions/run-mx-p2p-test/action.yml
zhengluo-nv: Enabled it for all p2p

zhengluo-nv: Actually nvidia-smi does not work in the MIG GPUs, so I can only enable it on the test cases that run on the A100 pool

.github/workflows/modelexpress-ci-tests.yml
```diff
@@ -416,6 +420,67 @@ jobs:
...
+          vcluster_name: ${{ env.VCLUSTER_NAME }}
+          vcluster_namespace: ${{ env.VCLUSTER_NAMESPACE }}
+
+      - name: Run P2P MLA model test
+        uses: ./.github/actions/run-mx-p2p-test
+        with:
+          kubeconfig_base64: ${{ steps.connect-vcluster.outputs.kubeconfig_base64 }}
+          namespace: mx-ci-vllm-mla-${{ github.run_id }}
+          server_image: ${{ env.MX_IMAGE_REPO }}/mx-server:${{ github.sha }}
+          manifest: ci/k8s/client/vllm/manifest-azure-tp8.yaml
```
tanushriya910: this might be a problem cause there is also tp=2 single node test that requires full GPUs to run so its also placed in the a100b pool (non MIG pool), both the tp=2 single node test and this test cannot run in parallel cause we only have 2 A100 (each with 8 GPUs) in the a100b node pool and this test alone will take all the 16 GPUs. so we can either:

1. remove the existing test-tp with single node (keep the multi-node one though) in favor of this one since this is a stronger test
2. reduce the TP size for this test if possible so both tests can run in parallel.

zhengluo-nv: I switched to a small MLA model with TP=2

tanushriya910 (approved):

PR #438, opened 2026-06-12
feat: add AcceleratorBackend boundary and CUDA capability gates

yafshar:
## Summary

Introduce an `AcceleratorBackend` boundary to centralize accelerator-specific torch device operations and capability checks.

- Replace direct `torch.cuda.*` usage with a backend abstraction  
- Introduce capability gating for CUDA-only fast paths:
  - pool registration
  - VMM arena
  - GDS loading  
- Wire the backend through:
  - engine adapters
  - NIXL registration
  - RDMA receive synchronization
  - load strategies  

NIXL memory registration remains the canonical `VRAM` segment; the backend is used strictly for local runtime operations and policy decisions.

CUDA behavior is unchanged:
- CUDA remains the default backend  
- All capability gates return `true`  
- Non-CUDA paths remain inactive until additional backends are implemented  

Mock backend tests are added to validate fallback behavior without requiring a GPU.

---

## Why

`NixlTransferManager`, load strategies, and engine adapters previously called `torch.cuda.*` directly, creating scattered CUDA dependencies.

`AcceleratorBackend` introduces a clear abstraction boundary by:
- centralizing device-specific operations  
- consolidating policy for accelerator-specific optimizations  

Capability checks are enforced consistently at:
- the arena lifecycle boundary  
- the NIXL registration boundary  

`supports_vmm_arena()` acts as the single source of truth for VMM enablement.

This design removes CUDA-specific assumptions at the call sites and enables future support for non-CUDA accelerators by modifying only backend implementations.

**No functional change is expected on CUDA.**

---

## Scope

This PR is strictly scoped to introducing the backend abstraction and capability gating.

- ✅ Adds `AcceleratorBackend`  
- ✅ Replaces direct CUDA calls with backend interface  
- ✅ Introduces capability gates for CUDA-specific fast paths  

---

## Testing

```bash
python -m pytest \
  modelexpress_client/python/tests/test_accelerator_backend.py \
  modelexpress_client/python/tests/test_vllm_adapter.py \
  modelexpress_client/python/tests/test_sglang_loader.py -v
# 36 passed

python -m pytest \
  modelexpress_client/python/tests/test_nixl_backend.py \
  modelexpress_client/python/tests/test_pool_registration.py \
  modelexpress_client/python/tests/test_gds_loader.py \
  modelexpress_client/python/tests/test_vmm_hook.py -v
# 67 passed
```

## Summary by CodeRabbit

* **New Features**
  * Introduced accelerator backend abstraction layer enabling flexible device-specific operation handling.
  * Standardized memory segment naming to use VRAM across all transfer operations.
  * Enhanced feature capability detection for GDS and VMM arena support through the accelerator backend.

* **Bug Fixes**
  * Improved fallback behavior when optional accelerator features are unsupported.

### DIFF
--- a/modelexpress_client/python/modelexpress/accelerator_backend.py
+++ b/modelexpress_client/python/modelexpress/accelerator_backend.py
@@ -64,7 +64,7 @@ class AcceleratorBackend(Protocol):
         """Return whether allocation-level NIXL pool registration is supported."""
         ...
 
-    def supports_vmm_arena(self) -> bool:
+    def supports_vmm(self) -> bool:
         """Return whether the CUDA VMM arena fast path is supported."""
         ...
 
@@ -116,7 +116,7 @@ class CudaAcceleratorBackend:
     def supports_pool_reg(self) -> bool:
         return True
 
-    def supports_vmm_arena(self) -> bool:
+    def supports_vmm(self) -> bool:
         return True
 
     def supports_gds(self) -> bool:
--- a/modelexpress_client/python/modelexpress/accelerator_backend.py
+++ b/modelexpress_client/python/modelexpress/accelerator_backend.py
@@ -1,0 +1,127 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Accelerator backend abstraction for device-specific operations."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Protocol
+
+import torch
+
+
+NIXL_ACCELERATOR_MEM_TYPE = "VRAM"
+
+
+class AcceleratorBackend(Protocol):
+    """Boundary for torch device control and accelerator capabilities."""
+
+    @property
+    def name(self) -> str:
+        """Backend family name for logs and capability policy, for example ``cuda``."""
+        ...
+
+    @property
+    def torch_device_type(self) -> str:
+        """Torch device type used to construct tensors, which may differ from ``name``."""
+        ...
+
+    @property
+    def nixl_mem_type(self) -> str:
+        """NIXL memory segment for accelerator memory."""
+        ...
+
+    def set_device(self, device_id: int) -> None:
+        """Make ``device_id`` current for this backend."""
+        ...
+
+    def current_device(self) -> int:
+        """Return the current local device ordinal."""
+        ...
+
+    def synchronize(self, device_id: int | None = None) -> None:
+        """Synchronize backend work on ``device_id`` or the current device."""
+        ...
+
+    def empty_cache(self) -> None:
+        """Release backend allocator cache where supported."""
+        ...
+
+    def torch_device(self, device_id: int) -> torch.device:
+        """Return a torch device object for ``device_id``."""
+        ...
+
+    def is_accel_tensor(self, tensor: torch.Tensor) -> bool:
+        """Return whether ``tensor`` lives on this backend's accelerator memory."""
+        ...
+
+    def supports_pool_reg(self) -> bool:
+        """Return whether allocation-level NIXL pool registration is supported."""
+        ...
+
+    def supports_vmm_arena(self) -> bool:
+        """Return whether the CUDA VMM arena fast path is supported."""
+        ...
+
+    def supports_gds(self) -> bool:
+        """Return whether GPUDirect Storage loading is supported."""
+        ...
+
+
+@dataclass(frozen=True)
+class CudaAcceleratorBackend:
+    """CUDA implementation of the accelerator backend boundary."""
+
+    @property
+    def name(self) -> str:
+        return "cuda"
+
+    @property
+    def torch_device_type(self) -> str:
+        return "cuda"
+
+    @property
+    def nixl_mem_type(self) -> str:
+        return NIXL_ACCELERATOR_MEM_TYPE
+
+    def set_device(self, device_id: int) -> None:
+        torch.cuda.set_device(device_id)
+
+    def current_device(self) -> int:
+        return int(torch.cuda.current_device())
+
+    def synchronize(self, device_id: int | None = None) -> None:
+        if device_id is None:
+            torch.cuda.synchronize()
+        else:
+            torch.cuda.synchronize(device_id)
+
+    def empty_cache(self) -> None:
+        torch.cuda.empty_cache()
+
+    def torch_device(self, device_id: int) -> torch.device:
+        return torch.device(self.torch_device_type, device_id)
+
+    def is_accel_tensor(self, tensor: torch.Tensor) -> bool:
+        return bool(tensor.is_cuda)
+
+    def supports_pool_reg(self) -> bool:
+        return True
+
+    def supports_vmm_arena(self) -> bool:
+        return True
+
+    def supports_gds(self) -> bool:
+        return True
+
+
+def accelerator_backend_for(device: torch.device | str) -> AcceleratorBackend:
+    """Return the backend implementation for ``device``.
+
+    Only CUDA devices are currently supported by this factory.
+    """
+    torch_device = torch.device(device)
+    if torch_device.type == "cuda":
+        return CudaAcceleratorBackend()
+    raise ValueError(f"Unsupported accelerator backend for torch device {torch_device!s}")
--- a/modelexpress_client/python/modelexpress/accelerator_backend.py
+++ b/modelexpress_client/python/modelexpress/accelerator_backend.py
@@ -56,6 +56,10 @@ class AcceleratorBackend(Protocol):
         """Return whether ``tensor`` lives on this backend's accelerator memory."""
         ...
 
+    def supports_rdma_p2p(self) -> bool:
+        """Return whether this backend supports NIXL RDMA P2P transfers."""
+        ...
+
     def supports_pool_reg(self) -> bool:
         """Return whether allocation-level NIXL pool registration is supported."""
         ...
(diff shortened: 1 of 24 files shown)


zhengluo-nv:
Hi Yaser, thanks for contributing. We are reviewing this and get back to you soon

yafshar:
> Hi Yaser, thanks for contributing. We are reviewing this and get back to you soon

Hi @zhengluo-nv, thanks again for taking a look. Just wanted to gently follow up and ask whether there’s anything I can add, clarify, split out, or test to make the review easier.

I have a few related changes queued up, so I’d be happy to adjust this PR in whatever direction is most helpful before opening more. No rush, and thanks again.

zhengluo-nv:
Hi Yaser, can you join this ext slack channel? We can discuss the design of supporting XPUs there
**ext-intel-xpu-dynamo**
https://nvidia.enterprise.slack.com/archives/C0AB0HDKP6D

modelexpress_client/python/modelexpress/accelerator_backend.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def is_accel_tensor(self, tensor: torch.Tensor) -> bool:
+        """Return whether ``tensor`` lives on this backend's accelerator memory."""
+        ...
+
+    def supports_pool_reg(self) -> bool:
+        """Return whether allocation-level NIXL pool registration is supported."""
+        ...
+
+    def supports_vmm_arena(self) -> bool:
```
zhengluo-nv: nit: Should be named `supports_vmm`

modelexpress_client/python/modelexpress/accelerators/base.py
```diff
@@ -0,0 +1,127 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhengluo-nv, line 1: Here are the places we also need to adapt accelerator backend:

Source identity needs accelerator compatibility
modelexpress_common/proto/p2p.proto
SourceIdentity does not include accelerator family/runtime/transfer ABI, so CUDA and XPU workers for the same model can share the same mx_source_id. Before non-CUDA backends are added, source discovery should be partitioned by accelerator compatibility, otherwise an XPU target can discover a CUDA source and attempt an invalid transfer path.

Tensor discovery is still CUDA-specific
modelexpress_client/python/modelexpress/tensor_utils.py
The backend interface adds is_accel_tensor(), but discovery still filters with tensor.is_cuda in capture_tensor_attrs, hidden tensor adoption, and iter_module_tensors. That means an XPU backend could exist and still publish no tensors through the vLLM path. The tensor discovery helpers should take/use the active accelerator backend.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
zhengluo-nv, line 55: This should check accelerator backend

modelexpress_client/python/modelexpress/accelerators/base.py
```diff
@@ -0,0 +1,127 @@
...
+from dataclasses import dataclass
+from typing import Protocol
+
+import torch
+
+
+NIXL_ACCELERATOR_MEM_TYPE = "VRAM"
+
+
+class AcceleratorBackend(Protocol):
```
zhengluo-nv, line 16: nit: let's move this to a accelerators folder and separate this file into 2: one for the base class and one for cuda

modelexpress_client/python/modelexpress/vmm/runtime.py
```diff
@@ -107,6 +107,17 @@ def maybe_enter_vmm_arena(ctx: "LoadContext") -> Iterator[None]:
         yield
         return
 
+    accelerator_backend = getattr(ctx, "accelerator_backend", None)
```
zhengluo-nv: The other places assume ctx.accelerator_backend is never None. Why do we have fallback here?

modelexpress_client/python/modelexpress/gds_loader.py
```diff
@@ -64,9 +65,10 @@ class MxGdsLoader:
             process(name, tensor)
     """
 
-    def __init__(self):
+    def __init__(self, accelerator_backend: AcceleratorBackend | None = None):
```
zhengluo-nv: I think a required accelerator_backend is better. This is internal code and no backward-compatibility concern. Same as the other classes where accelerator_backend is passed

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
yafshar: Good catch. `RdmaStrategy.is_available()` currently checks NIXL availability, server/decentralized metadata configuration, and transfer policy, but it does not ask the selected accelerator backend whether NIXL RDMA P2P is supported.

I'll add this as an accelerator capability, consistent with the existing `supports_pool_reg()` / `supports_vmm_arena()` / `supports_gds()` gates:

```
def supports_rdma_p2p(self) -> bool:
    """Return whether this backend supports NIXL RDMA P2P transfers."""
    ...
```

CUDA returns `True`. Test/mock backends default to `False` so CI can cover the skip path. Any future backend on the same NIXL VRAM RDMA path opts in by returning `Tru`e.

`RdmaStrategy.is_available()` will then check this right after the NIXL-availability check and before the server/config and transfer-policy checks, with a log message naming the backend. I'll add focused coverage beside the existing RDMA availability tests.

modelexpress_client/python/modelexpress/vmm/runtime.py
yafshar: Good point. `LoadContext` always provides an `accelerator_backend`, and the other strategy paths already treat it as required. The use of `getattr(..., None)` here was only a defensive fallback for older or minimal test stubs, but it can inadvertently bypass the backend VMM capability check when `MX_VMM_ARENA=1`.

I’ll switch this to a direct call and update the VMM test stubs to explicitly include an accelerator backend.

modelexpress_client/python/modelexpress/accelerators/base.py
yafshar: Thanks, your comments are exactly what the next two PRs in this stack address.
I split the work intentionally so this PR focuses on introducing the backend
boundary without changing CUDA runtime behavior.

**Source identity / accelerator compatibility**

This is covered in the follow-up PR for accelerator runtime metadata and
compatibility filtering. That change adds an accelerator field to
WorkerMetadata / GetTensorManifestResponse and filters RDMA candidates in
RdmaStrategy, so an XPU target will skip CUDA sources.

One design point I’d really appreciate your opinion on: I’m keeping accelerator
out of `SourceIdentity` / the `mx_source_id` hash, and instead enforcing
compatibility at discovery time via filtering.

The reasoning is:

- `mx_source_id` currently reflects model/source identity. Adding accelerator
  would change hashes for existing CUDA deployments and break pinned source-id
  cross-check tests.
- We need to remain backward compatible with older workers where accelerator is
  missing or unknown.

From a safety perspective, the outcome is the same: an XPU target won’t transfer
from a CUDA source. The difference is where that constraint is applied.

If you feel strongly that accelerator should be part of `SourceIdentity`, that’s
a bigger compatibility shift, but I’m happy to pull it forward.

**CUDA-specific tensor discovery**

This is handled in the follow-up PR for backend-aware tensor discovery. It
replaces `tensor.is_cuda` in `capture_tensor_attrs`, hidden tensor adoption, and
`iter_module_tensors` with `backend.is_accel_tensor()`, threads the active
backend through the discovery helpers, and renames `_find_hidden_cuda_tensors`
to `_find_hidden_accel_tensors`.

This PR intentionally keeps discovery identical to CUDA. Until the follow-up
lands, an XPU backend wouldn’t publish tensors through the vLLM path, which is
why the XPU backend PR is stacked on top of both.

modelexpress_client/python/modelexpress/accelerators/base.py
yafshar: I’ll move the backend code into an `accelerators/` package with
`accelerators/base.py` for the protocol/shared constants and
`accelerators/cuda.py` for `CudaAcceleratorBackend`. I’ll update the existing
import sites rather than leave a top-level shim, so there is a single canonical
location.

modelexpress_client/python/modelexpress/accelerator_backend.py
yafshar: I’ll also rename `supports_vmm_arena()` to `supports_vmm()`.

modelexpress_client/python/modelexpress/gds_loader.py
yafshar: Agreed. In the current runtime path, `MxGdsLoader` is constructed by
`GdsStrategy`, and `GdsStrategy` already passes `ctx.accelerator_backend`, so the
default CUDA fallback is unnecessary.

I'll make `accelerator_backend` required in `MxGdsLoader.__init__()` and update
the tests and docstring examples to pass it explicitly. I'll also make the same
change for `GdsTransferManager`, since `MxGdsLoader` already passes its backend
through and the remaining no-backend call sites are tests and docstring examples.

I'm leaving `NixlTransferManager` optional for this PR because the TRT-LLM path
still constructs it directly and is not backend-aware yet. I'll make that one
required when the TRT-LLM backend-awareness change lands, or I can pass
`CudaAcceleratorBackend()` explicitly there now if you prefer.

yafshar:
@zhengluo-nv Thanks for bearing with the force-push. I rebased onto the latest upstream/main and resolved the docs/ARCHITECTURE.md conflict by keeping both the GDS accelerator-backend condition and the new TP section.  New review-fix commits are stacked on top.

zhengluo-nv (approved):

PR #441, opened 2026-06-17
fix(sglang): honor ModelExpress model name override

Broduker:
## Summary
- Honor `load_config.modelexpress_model_name` when SGLang builds `SourceIdentity.model_name`.
- Preserve the existing fallback to `model_config.model_path` / `model_config.model`.
- Add a regression test for the SGLang ModelExpress model-name override.

## Problem

Without this, SGLang uses the local model path as `SourceIdentity.model_name`, which changes the computed `mx_source_id` and can prevent ModelExpress from matching the intended source.

## Tests
- `uv run --extra dev pytest tests/test_sglang_loader.py -q`

## Summary by CodeRabbit

* **New Features**
  * SGLang engine adapter now supports optional model name override via configuration.

* **Bug Fixes**
  * Improved tensor discovery in SGLang adapter to include hidden tensors and buffers, with proper deduplication for shared storage.
  * Enhanced tensor attribute capture during weight processing.

* **Tests**
  * Added test coverage for model name override functionality and tensor collection behavior.


modelexpress_client/python/modelexpress/engines/sglang/adapter.py
```diff
@@ -268,7 +272,14 @@ def build_sglang_source_identity(model_config: ModelConfig) -> p2p_pb2.SourceIde
     )
 
 
-def _get_model_name(model_config: ModelConfig) -> str:
+def _get_model_name(
+    model_config: ModelConfig,
+    load_config: LoadConfig | None = None,
+) -> str:
+    if load_config is not None:
+        override = getattr(load_config, "modelexpress_model_name", None)
```
zhengluo-nv, line 280: This override does not appear reachable from the current SGLang runtime contract. Upstream SGLang `main` only carries `modelexpress_url` and `modelexpress_transport` on `LoadConfig`, and `--modelexpress-config` only parses `url` / `transport`, so the documented launch path never sets `load_config.modelexpress_model_name`. That means `SourceIdentity.model_name` still falls back to `model_config.model_path`, leaving the local-path mismatch described in the PR unresolved unless a matching SGLang change also adds and passes this field. Please either add/land that SGLang plumbing and document the key, or derive the intended model name from an existing runtime field.

zhengluo-nv (commented):
I found one integration-contract issue to address.

modelexpress_client/python/modelexpress/engines/sglang/adapter.py
Broduker: Thanks, @zhengluo-nv.

Originally, we wanted to solve the SGLang same-configuration loading issue for ModelExpress: source and target instances should be able to start with the same SGLang/Dynamo configuration, without explicitly separating source/target roles or requiring seed instance settings. In our custom SGLang branch, this was implemented by delegating the adaptive loading policy to the ModelExpress package, and `modelexpress_config.model_name` was passed down as `load_config.modelexpress_model_name` to keep source/target identity stable when their local model paths differed.

This PR was created to support that custom SGLang behavior on the ModelExpress side, by making the SGLang adapter honor `load_config.modelexpress_model_name` when building `SourceIdentity`.

However,  SGLang has now been addressed that delegates ModelExpress loading to the ModelExpress package. The upstream implementation also supports same-configuration adaptive loading for the ModelExpress backend, but it does not expose or pass `modelexpress_model_name`; it derives identity from `model_config.model_path`.

So this PR is no longer necessary for the upstream SGLang path.

PR #448, opened 2026-06-23
docs(architecture): document GDS full-tensor read limitation under TP

AndyDai-nv:
## What

Documents a known issue of the GDS loading path (`GdsStrategy` / `MxGdsLoader`) in `docs/ARCHITECTURE.md`.

## Mechanism

`MxGdsLoader` issues NIXL/GDS reads using the full `(file_offset, size)` from each tensor's `data_offsets`, then hands the full checkpoint tensor to the engine through the normal `model.load_weights()` path. vLLM's parameter loaders apply TP sharding themselves via `loaded_weight.narrow(...)` before copying into the final parameter buffer.

With TP > 1, each rank reads the entire checkpoint tensor through GDS and keeps only its local shard, so GDS/disk bytes read scale with the TP degree instead of staying constant. The default disk loader leans on mmap and lazy page faults and skips bytes it would discard. End-to-end numbers per TP degree are still to be measured.

## Scope

Docs only, no code change. The writeup also notes that addressing this is a load-contract change (a TP-aware iterator, or a vLLM-specific path that plans file ranges from parameter metadata and reads into the final param buffers) rather than a one-liner.


zhengluo-nv (approved):

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/448/commits/980138af76734108339986033b685a5b60e2b14e

PR #452, opened 2026-06-25
test(ci): add vLLM data-parallel P2P test

AndyDai-nv:
## What

Adds a `test-dp` CI job that exercises ModelExpress P2P RDMA weight transfer under **vLLM native data parallelism** (`--data-parallel-size 2`), closing TEST_PLAN row 5.2.

It mirrors the existing `test-tp` job structurally but targets a different parallelism axis.

## Why tp_size=1, not 2 (the key design point)

DP is fundamentally different from TP for ModelExpress:

- **TP** shards one model's weights across GPUs — each rank owns a distinct slice, source publishes one ModelMetadata CR per rank, target pairs rank-for-rank.
- **DP** runs N *full* model copies. They are interchangeable, and `build_source_identity` carries no DP dimension (tp/pp/ep only), so every DP core resolves to `worker_rank` 0.

So the source publishes a single rank-0 CR, and the target's DP cores each pull the full rank-0 weight set over RDMA — **N concurrent pulls from one source**, not rank-paired transfers. Running with `tp_size=1` means the per-rank agent assertion correctly expects exactly one distinct source agent (rank 0). No changes needed to `run-mx-p2p-test`, `test_p2p_k8s.py`, or `conftest.py`.

## Changes

- `ci/k8s/client/vllm/manifest-azure-dp2.yaml` (new) — copy of `manifest-azure-tp2.yaml` with `--data-parallel-size 2 --tensor-parallel-size 1`; same non-MIG `a100b` pool, 2 GPUs / 2 IB devices (one full GPU per DP core).
- `.github/workflows/modelexpress-ci-tests.yml` — new `test-dp` matrix job (vLLM, DP=2, single-node); wired into `ci-status-check` `needs:`.
- `ci/TEST_PLAN.md` — row 5.2 flipped to **In CI (vLLM single-node DP=2)**, with the per-replica-inference limitation noted (vLLM native DP exposes a single load-balanced endpoint, so individual cores can't be addressed externally).

## Follow-ups

- TRT-LLM / SGLang DP variants.
- A true per-replica inference assertion would need a router-fronted multi-Deployment topology (or Dynamo DP).

## Summary by CodeRabbit

* **New Features**
  * Added CI coverage for vLLM single-node data parallelism (DP=2) weight-transfer testing.
  * Introduced a new Kubernetes test job manifest for the DP-based P2P RDMA setup.

* **Bug Fixes**
  * Expanded the main CI status check so the new DP test is now required before merges.

* **Documentation**
  * Updated the CI test plan to reflect the current DP test coverage and status.


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/452/commits/5a1ff7f3372cd55e9e7599ba80d4af4ce6961f22

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/5a1ff7f3372cd55e9e7599ba80d4af4ce6961f22

zhengluo-nv (approved):

PR #454, opened 2026-06-25
feat: route tensor discovery through AcceleratorBackend predicate

yafshar:
## Summary

Makes tensor discovery backend-aware so future non-CUDA workers can correctly discover publishable tensors instead of emitting empty sets. Builds on the AcceleratorBackend boundary (#438). Behavior remains unchanged for CUDA.

## Changes

- Replace hardcoded `tensor.is_cuda` checks with `AcceleratorBackend.is_accel_tensor()`.
- Thread the active backend from engine adapters into `tensor_utils` helpers (`capture_tensor_attrs`, `adopt_hidden_tensors`,
  `iter_module_tensors`, `collect_module_tensors`) and the SGLang `collect_sglang_tensors` path.
- Ensure both vLLM `capture_tensor_attrs` call sites (post-load and model-specific finalizers) pass `self.accelerator_backend`.
- Rename `_find_hidden_cuda_tensors` → `_find_hidden_accel_tensors`; update docstrings and logs to use backend-agnostic terminology. 
- Default to the CUDA backend when none is provided, preserving existing runtime behavior.

## Testing

- Discovery helpers now execute under a CPU-mock backend in CI (hidden-tensor adoption and collection no longer skip in GPU-less environments).
- Existing no-backend tests confirm default CUDA behavior is unchanged (CPU tensors are still ignored).
- Python test suite:
  - Local: `python -m pytest modelexpress_client/python/tests -v` — 405 passed
  - GPU: `python -m pytest modelexpress_client/python/tests -v` — 426 passed, 1 skipped
- `pre-commit run --all-files` passes cleanly (cargo fmt/clippy/check).

## Summary by CodeRabbit

* **New Features**
  * Tensor discovery and collection are now accelerator-backend aware, including filtering by the active backend.
  * Hidden tensors can be detected and registered across broader module/object graphs for supported backends.
* **Bug Fixes**
  * Improved accuracy by excluding non-accelerator-resident tensors from results.
  * Enhanced handling of non-contiguous storage views and tied-weight deduplication during registration.
* **Documentation**
  * Updated architecture documentation to use accelerator-generic terminology.
* **Tests**
  * Added and updated coverage for backend-aware tensor discovery and hidden-tensor adoption.


yafshar:
@coderabbitai

modelexpress_client/python/modelexpress/engines/sglang/adapter.py
```diff
@@ -211,19 +215,25 @@ def _call_sglang_post_load_weights(model: torch.nn.Module) -> None:
...
     match the upstream integration.
     """
+    backend = accelerator_backend or CudaAcceleratorBackend()
     tensors: dict[str, torch.Tensor] = {}
     seen_ptrs: set[int] = set()
 
     for name, param in model.named_parameters():
         t = param.data
+        if not backend.is_accel_tensor(t):
+            continue
```
zhengluo-nv: nit: This check should skip the whole method. Currently the for loop is still executed

modelexpress_client/python/modelexpress/engines/sglang/adapter.py
```diff
@@ -211,19 +215,25 @@ def _call_sglang_post_load_weights(model: torch.nn.Module) -> None:
             post_load_weights()
 
 
-def collect_sglang_tensors(model) -> dict[str, torch.Tensor]:
+def collect_sglang_tensors(
+    model,
+    accelerator_backend: AcceleratorBackend | None = None,
```
zhengluo-nv: nit: Move this method under the SglangAdapter class so you don't have to pass in this extra arg. Same as the other methods

modelexpress_client/python/modelexpress/tensor_utils.py
```diff
@@ -76,57 +86,82 @@ def capturing_setattr(self, name, value):
         nn.Module.__setattr__ = original_setattr
 
 
-def _find_hidden_cuda_tensors(
-    obj: object, visited: set[int], depth: int = 0,
+def _find_hidden_accel_tensors(
+    obj: object,
+    visited: set[int],
+    accelerator_backend: AcceleratorBackend | None = None,
```
zhengluo-nv: Make this a required arg

modelexpress_client/python/modelexpress/tensor_utils.py
```diff
@@ -76,57 +86,82 @@ def capturing_setattr(self, name, value):
...
         return []
     visited.add(id(obj))
 
     results: list[tuple[str, torch.Tensor]] = []
 
-    if isinstance(obj, torch.Tensor) and obj.is_cuda and obj.numel() > 0:
-        results.append(("t", obj))
+    if isinstance(obj, torch.Tensor):
+        tensor = cast(torch.Tensor, obj)
+        if backend.is_accel_tensor(tensor) and tensor.numel() > 0:
```
zhengluo-nv: backend is always the same, so why do you still run the whole recursion? We should skip

modelexpress_client/python/modelexpress/tensor_utils.py
```diff
@@ -35,8 +44,8 @@ def safe_checksum(tensor: torch.Tensor) -> str:
 
 
 @contextmanager
-def capture_tensor_attrs():
-    """Intercept bare CUDA tensor assignments during process_weights_after_loading.
+def capture_tensor_attrs(accelerator_backend: AcceleratorBackend | None = None):
```
zhengluo-nv: Don't use optional argument here

modelexpress_client/python/modelexpress/engines/sglang/adapter.py
yafshar: `is_accel_tensor(t)` checks each param's device, not the backend, so it can't gate the whole method. Here, params can be mixed-device. Under CPU offload, `named_parameters()` returns host- and device-resident params together; skipping the method (or bailing on the first CPU param) would drop the device params that must be registered, giving a truncated manifest and a broken transfer. The per-param continue filters exactly the CPU ones and keeps the rest. Same filter runs in the vLLM path too. I'd leave it as-is.

modelexpress_client/python/modelexpress/engines/sglang/adapter.py
yafshar: Done! moved into `SglangAdapter._collect_tensors`, reads `self.accelerator_backend`, no extra arg. `discover_tensors` calls it. Matches the other methods.

modelexpress_client/python/modelexpress/tensor_utils.py
yafshar: The recursion isn't performing backend checks; it's traversing arbitrary nested structures to locate tensors at unknown depths. Since `is_accel_tensor()` operates on individual tensors, we can't know whether a nested tensor belongs on the accelerator until we've actually reached it. That discovery process is the reason for the recursive walk, so there isn't a backend-related condition that can be pulled out and evaluated once up front. I'd keep the implementation as-is unless I'm missing something in the feedback.

modelexpress_client/python/modelexpress/tensor_utils.py
yafshar: Done! `accelerator_backend` is now a required arg, dropped the `None` default and the resolve fallback.

modelexpress_client/python/modelexpress/tensor_utils.py
yafshar: Done!

zhengluo-nv (approved):

PR #455, opened 2026-06-25
test(ci): add vLLM expert-parallel (EP=2) P2P test

AndyDai-nv:
## What

Adds a `test-ep` CI job that exercises ModelExpress P2P RDMA weight transfer under **vLLM expert parallelism** (`--tensor-parallel-size 2 --enable-expert-parallel`) on an MoE model (`deepseek-ai/DeepSeek-V2-Lite`), closing TEST_PLAN row 5.3.

Follows the same structure as `test-tp` / `test-dp` / the MLA model test.

## Why tp_size=2 (rank-paired, like TP)

In vLLM, EP layers on top of the TP world: attention stays TP-sharded while the MoE experts are distributed expert-parallel across the same ranks. So each rank owns **distinct** weights (its attention slice + its expert subset), the source publishes one ModelMetadata CR per rank, and the target pairs rank-for-rank — identical in shape to the TP test. Hence `tp_size=2` and the per-rank agent assertion expects two distinct source agents (ranks 0, 1).

## Coverage beyond test-tp

Enabling EP sets `parallel_config.expert_parallel_size > 0`, which `build_source_identity` folds into `mx_source_id`. This is the **first job to exercise `rdma_strategy` rank pairing with `expert_parallel_size > 0`** (the hash was computed but never tested). A source/target mismatch in that field would force a disk fallback and fail the RDMA marker assertion.

## Changes

- `ci/k8s/client/vllm/manifest-azure-ep2.yaml` (new) — mirrors `manifest-azure-tp2.yaml` with the added `--enable-expert-parallel` arg; same non-MIG `a100b` pool, 2 GPUs / 2 IB devices.
- `.github/workflows/modelexpress-ci-tests.yml` — new `test-ep` matrix job (vLLM, EP=2, single-node, DeepSeek-V2-Lite, `tp_size=2`, 3600s pod-ready timeout); wired into `ci-status-check` `needs:`.
- `ci/TEST_PLAN.md` — row 5.3 flipped to **In CI (vLLM single-node EP=2)**.

Reuses `run-mx-p2p-test` + `test_p2p_k8s.py` unchanged.

## Follow-ups

- Pure `TP=1 × EP=N` shape (vLLM expresses EP over the TP/DP world, so EP>1 needs `TP*DP>1`; TP=2+EP is the simplest single-pod shape).
- Per-expert-subset transfer assertion (current coverage confirms rank-paired transfer + correct identity hash, not that rank N pulled only expert subset N).
- TRT-LLM / SGLang EP variants.

## Summary by CodeRabbit

* **New Features**
  * Added CI coverage for expert-parallel weight-transfer testing on a multi-GPU MoE model.
  * Introduced a new Azure-based validation job with expanded parallelism support and host-placement checks.

* **Tests**
  * Updated the CI test plan to mark expert-parallel validation as in progress in CI.
  * Broadened branch-protection gating so the new expert-parallel test must pass before changes merge.


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/455/commits/a13356af655a34a7439ef06f1290975049490a15

KavinKrishnan (approved):

PR #461, opened 2026-06-29
feat(client): add P2P source-selection layer (random, rendezvous_hash)

yixinhuang48:
## Summary

Adds P2P source-selection **observability + a pluggable selection module** to the
Python client. This is an extension point and measurement tooling, **not** a
balance/throughput change in itself.

- Replaces the hardcoded `random.shuffle(candidates)` in the RDMA load path with
  a small, stateless, client-side `SourceSelector` abstraction (name registry +
  `MX_P2P_SOURCE_SELECTOR` env var) so future load-/topology-aware policies are
  drop-ins on the selection path.
- Ships two policies: `random` (behavior-preserving default) and
  `rendezvous_hash` (deterministic HRW spreading, opt-in).
- Adds structured selection logs and an opt-in Prometheus metrics collector
  (`mx_p2p_*`) so selection schemes can be measured on real workloads.
- `random` stays the default, so there is **no behavior change** unless a policy
  is explicitly enabled.

## Why

The RDMA loader picks P2P sources by `ListSources(READY)` -> filter by
`worker_rank` -> `random.shuffle` -> try the top `MAX_SOURCE_RETRIES=3`. Random
ordering is simple but gives no predictable balancing during large fan-out or
RL-style weight sync, cannot be made deterministic for tests, and offers no
extension point for load/topology signals. There were also no metrics for P2P
source selection or transfers — only logs.

## What Changed

- ✅ **`modelexpress/source_selection.py`** — `SourceSelector` Protocol +
  `ScoredSelector` base (orders by descending score), a name->factory registry
  (`register_selector`/`get_selector`) with graceful fallback to `random`, and
  two policies:
  - `random` — local-RNG shuffle (does not touch process-global RNG state).
  - `rendezvous_hash` — HRW hashing (blake2b, not Python's salted `hash()`) of
    target identity + each candidate identity; deterministic, stable across
    restarts, minimally disrupted by source-set changes.
  - Selectors read the few fields they need (`worker_rank`, `worker_id`,
    `identity.model_name`) directly off the live `LoadContext`, so there is no
    parallel context type.
- ✅ **Integration** — `RdmaStrategy._find_source_instances()` orders via the
  configured selector; the SGLang TransferEngine path (same central-coordinator
  `list_sources` pattern) is wired identically. The `MAX_SOURCE_RETRIES` slice
  and the "no retry after transfer start" rule are unchanged.
- ✅ **Config** — `MX_P2P_SOURCE_SELECTOR` (`random` | `rendezvous_hash`);
  invalid values warn and fall back to `random`.
- ✅ **Observability** — always-on structured selection logs; a generic opt-in
  Prometheus collector (`metrics.py`, `MX_METRICS_ENABLED=1`) with a pull
  `/metrics` endpoint (auto-start) and Pushgateway flush on exit. The P2P
  source-selection metrics are the first group; the collector is feature-agnostic
  so other client metrics can be added later.
- ✅ **Benchmark methodology + results** — documented in `docs/ARCHITECTURE.md`
  (Source Selection -> "Selection efficacy (measured)").

| Variable | Default | Values |
|---|---|---|
| `MX_P2P_SOURCE_SELECTOR` | `random` | `random`, `rendezvous_hash` |
| `MX_METRICS_ENABLED` | `0` | `1` to enable the `mx_p2p_*` metrics (needs the `metrics` extra) |

**Non-goals (deferred):** proto/server-schema changes; NIXL/RDMA protocol
changes; load-/topology-/weighted policies (the scoring interface accepts them
later); k8s-service pod-level routing; an always-on metrics pipeline. A
`load_aware`/`topology_aware` policy is a drop-in on the scoring interface, but
additionally needs a per-source load/topology **signal delivered to the
selector** — a new `SourceInstanceRef` field (`p2p.proto` + server change, plus
the source reporting it). That signal plumbing, not the policy, is the bulk of
the follow-up and is deferred with it.

## Testing (Local Validation)

```bash
cd modelexpress_client/python
uv pip install -e ".[dev]"
.venv/bin/python -m pytest tests/    # 419 passed, 32 skipped
```

- New `tests/test_source_selection.py` — registry/config fallback, random
  set-preservation + local-RNG isolation, rendezvous determinism + pinned
  cross-process hash + spreading + remove-source stability + tie-stability, the
  RdmaStrategy integration cases (rank filter, max-retry slice, metadata-miss
  fallback, no-retry-after-transfer-start), `configured_policy_label`, and the
  load()-path metric outcomes (selection/success/fallback/metadata-miss).
- New `tests/test_metrics.py` — locks the two safety guarantees: recorders are
  no-ops when disabled, and never raise into the load path.
- `tests/test_sglang_loader.py` — selector wiring on the TransferEngine path
  (rank filter, selector-driven iteration order, oneof gating, skip-not-found).
- Updates the 5 `tests/test_vllm_loader.py` cases that patched the removed
  module-level `random` to use a deterministic identity selector.

(The 7 `test_pool_registration` errors in a non-GPU env are the pre-existing
missing optional `cuda-python`/`vmm` dep, unrelated to this change.)

## E2E Validation (8×B200 nodes, InfiniBand RDMA)

vLLM `--load-format modelexpress`, TP=1 (1 GPU/replica), central-coordinator
(`kubernetes` CRD backend), P2P-endpoint path (`MX_P2P_METADATA=1`). The 0.5.0
client overlay runs over the stock runtime image (no rebuild).

**Source.** Loads the model, registers tensors with NIXL, publishes a READY
ModelMetadata CRD with its `worker_grpc_endpoint`:
`Published metadata to MX server (mx_source_id=..., worker_id=...)` -> `Status -> READY`.

**Target.** Selects a source via the policy and pulls from it cross-node:
```
Source selection: source_selector=rendezvous_hash source_candidates_total=2 source_candidates_rank_matched=2
Source attempt: source_attempt_index=0 source_worker_id=6459bef1 mx_source_id=...
P2P mode: fetching tensor manifest from <peer>:6555  (267 tensors)
RDMA transfer complete: 267 tensors, 0.99 GB, 181.7 Gbps
```

**Prometheus (`MX_METRICS_ENABLED=1`, scraped from a target `/metrics`):**
```
mx_p2p_source_selections_total{policy="rendezvous_hash",source_worker_id=...} 1
mx_p2p_source_attempts_total{result="success"} 1
mx_p2p_transfer_seconds_count{outcome="success"} 1
mx_p2p_candidates{stage="rank_matched"} sum=2
mx_p2p_source_selection_seconds ~ 3e-5s
```

**Larger-scale runs** (each policy against 4 pre-warmed sources; the pool grows
live as targets become sources):

| Model (per-transfer) | M×N | policy | RDMA success | mean bandwidth |
|---|---|---|---|---|
| Qwen2.5-0.5B (0.99 GB) | 4×12 | rendezvous_hash | 12/12 | 179.7 Gbps |
| Qwen2.5-0.5B (0.99 GB) | 4×12 | random | 12/12 | 181.2 Gbps |
| Qwen2.5-7B (~15 GB) | 4×8 | rendezvous_hash | 8/8 | 210.8 Gbps |
| Qwen2.5-7B (~15 GB) | 4×8 | random | 7/8 | 211.2 Gbps |

42 real cross-node NIXL RDMA transfers across both models/policies (one 7B
target fell back to disk cleanly when a just-joined source's endpoint wasn't
ready — the fallthrough path behaving correctly). **Bandwidth is
policy-independent**; 7B's larger transfers amortize NIXL setup (~211 vs
~180 Gbps).

**Offline efficacy** (selector logic only — deterministic in policy and
identity, so it reproduces the on-cluster selection distribution without
transfers; methodology + full table now in `docs/ARCHITECTURE.md`):

| M,N | max-source share (rand → rdv, ideal) | re-pick churn (rand → rdv) | 1-src disruption (rdv) |
|---|---|---|---|
| 4,20 | 0.380 → 0.366 (0.250) | 60% → 0% | 0.0% |
| 4,40 | 0.330 → 0.320 (0.250) | 65% → 0% | 0.0% |
| 8,32 | 0.223 → 0.226 (0.125) | 84% → 0% | 0.0% |

In the equidistant, fixed-pool case **balance is equal** (both uniform-hash);
`rendezvous_hash`'s edge is **determinism** (0% re-pick churn vs random's
~(M-1)/M) and **minimal disruption** (0% of unaffected targets move when a
source leaves), at no balance/bandwidth cost. Under live fan-out (a growing
pool, not a fixed-pool comparison) deterministic hashing can pile onto an
always-present source while random self-averages — which is the gap the
deferred `load_aware` policy closes, not a ranking of the two Phase 1 policies.

## Release Notes

**New Features**
- Configurable client-side P2P source selection (`MX_P2P_SOURCE_SELECTOR`):
  `random` (default) and `rendezvous_hash`.
- Opt-in Prometheus metrics collector for the client
  (`MX_METRICS_ENABLED=1`); ships the P2P source-selection group (`mx_p2p_*`).

**Documentation**
- `docs/DEPLOYMENT.md` — new P2P selection + metrics environment variables.
- `docs/ARCHITECTURE.md` — Source Selection section (incl. measured selection
  efficacy), strategy-table update, and module tree.


modelexpress_client/python/modelexpress/engines/sglang/loader.py
```diff
@@ -217,7 +217,22 @@ def _find_transfer_engine_source(self, ctx: LoadContext):
         candidates = [
             inst for inst in response.instances if inst.worker_rank == ctx.worker_rank
         ]
-        random.shuffle(candidates)
+        selection_ctx = SourceSelectionContext(
```
zhengluo-nv: Why not just use LoadContext?

yixinhuang48:
Addressed the CodeRabbit comments in 5955c8b:
- `configured_policy_label` now resolves through the registry (matches `get_selector`), so a registered-but-raising factory labels as `random` like selection.
- Pushgateway pushes use a per-host `grouping_key` so concurrent workers don't overwrite each other.
- bench: fail fast on a failed Redis flush; gate target readiness on vLLM `/health` (load-complete) so makespan/collection don't race container start; ideal max-share uses the discrete `ceil(total/M)/total` baseline; logs collected under `mktemp -d`.
- tests: deterministic sglang not-found ordering + failing-factory label fallback.

examples/p2p_transfer_k8s/source_selection_bench/run_bench.sh
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env bash
```
zhengluo-nv: Let's not put these files as examples. Examples folder are external customer facing. We can document the benchmark result in the docs

modelexpress_client/python/modelexpress/source_selection_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+ENV_PORT = "MX_P2P_METRICS_PORT"
+ENV_PUSHGATEWAY = "MX_P2P_METRICS_PUSHGATEWAY"
+ENV_SCHEME = "MX_P2P_METRICS_SCHEME"
+
+
+def _enabled() -> bool:
+    return os.environ.get(ENV_ENABLED, "0") in ("1", "true", "True")
+
+
+class _Metrics:
```
zhengluo-nv: Let's try to make this a generic metric collector that can later collect all useful metrics, like model load latency, p2p transfer latency, etc.

modelexpress_client/python/modelexpress/source_selection_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+import atexit
+import logging
+import os
+
+logger = logging.getLogger("modelexpress.source_selection.metrics")
+
+ENV_ENABLED = "MX_P2P_METRICS_ENABLED"
+ENV_PORT = "MX_P2P_METRICS_PORT"
+ENV_PUSHGATEWAY = "MX_P2P_METRICS_PUSHGATEWAY"
+ENV_SCHEME = "MX_P2P_METRICS_SCHEME"
```
zhengluo-nv: nit: Name it MX_METRICS_ENABLED because modelexpress does not only have the p2p feature.

modelexpress_client/python/modelexpress/engines/sglang/loader.py
yixinhuang48: Good call — done in 0babc1f. Dropped `SourceSelectionContext` entirely; selectors now take the live `LoadContext` and read just `worker_rank` / `worker_id` / `identity.model_name`. Factories are now zero-arg (the per-target signal flows in via `order(candidates, ctx)`), which also collapsed `configured_policy_label()` to `get_configured_selector().name`. To keep `source_selection.py` from importing the load_strategy package, the annotation is a `TYPE_CHECKING` forward ref — no import cycle, and the unit tests pass a tiny duck-typed stand-in.

examples/p2p_transfer_k8s/source_selection_bench/run_bench.sh
yixinhuang48: Done in 0babc1f — removed `examples/p2p_transfer_k8s/source_selection_bench/` and folded the methodology + measured results into `docs/ARCHITECTURE.md` (Source Selection → "Selection efficacy (measured)"). The property claims the offline sim checked (determinism, bounded disruption, first-choice spreading) are already covered by unit tests in `test_source_selection.py`.

modelexpress_client/python/modelexpress/source_selection_metrics.py
yixinhuang48: Done in 0babc1f — `source_selection_metrics.py` is now `metrics.py` with a generic `MetricsCollector`. The enable / pull-`/metrics` / Pushgateway / no-op plumbing is feature-agnostic, and the metric families are grouped (`# --- P2P source-selection group ---`) so model-load latency, transfer latency, etc. slot in as new groups without touching the exposition. I kept the existing `mx_p2p_*` metric names since those are genuinely P2P metrics; happy to add the next group(s) in a follow-up.

modelexpress_client/python/modelexpress/source_selection_metrics.py
yixinhuang48: Done in 0babc1f — renamed to `MX_METRICS_ENABLED` / `MX_METRICS_PORT` / `MX_METRICS_PUSHGATEWAY` / `MX_METRICS_SCHEME` (dropped the `_P2P_` scope), updated across code, docs, and pyproject. Since the feature is new in this PR there's no back-compat alias to keep.

yixinhuang48:
@zhengluo-nv thanks for the review — addressed all four in 0babc1f:

- **Use `LoadContext`**: removed `SourceSelectionContext`; selectors take the live `LoadContext` directly (read only `worker_rank` / `worker_id` / `identity.model_name`). Factories are now zero-arg and `configured_policy_label()` collapses to `get_configured_selector().name`. A `TYPE_CHECKING` forward ref avoids any import cycle.
- **Generic metric collector**: `source_selection_metrics.py` → `metrics.py` with a generic `MetricsCollector`; the enable/pull/push/no-op plumbing is feature-agnostic and the P2P metrics are now the first grouped family, so model-load/transfer latency etc. drop in later.
- **Env var rename**: `MX_P2P_METRICS_*` → `MX_METRICS_*` across code, docs, and pyproject.
- **Benchmark out of `examples/`**: deleted `examples/p2p_transfer_k8s/source_selection_bench/`; methodology + measured results now live in `docs/ARCHITECTURE.md` (Source Selection → "Selection efficacy (measured)").

Full suite green locally (419 passed, 32 skipped; the 7 `test_pool_registration` errors are the pre-existing missing-`cuda` optional dep). PR description updated to match.

modelexpress_client/python/modelexpress/engines/sglang/loader.py
```diff
@@ -217,7 +217,16 @@ def _find_transfer_engine_source(self, ctx: LoadContext):
         candidates = [
             inst for inst in response.instances if inst.worker_rank == ctx.worker_rank
         ]
-        random.shuffle(candidates)
+        selector = get_configured_selector()
```
zhengluo-nv, line 223: Why is this only applied to sglang not vllm?

yixinhuang48: It is applied to vLLM — just at a shared layer rather than here. Source ordering lives in `RdmaStrategy._find_source_instances()` (`load_strategy/rdma_strategy.py`), which is the NIXL/RDMA load path **vLLM** runs via `LoadStrategyChain.run()` (`engines/vllm/loader.py:100`). SGLang's *default* `nixl` transport goes through that same chain (`engines/sglang/loader.py:97`), so it's covered there too.

This call (line 220) is only for SGLang's separate `transfer_engine` transport (`_load_model_via_transfer_engine`), which discovers sources with its own `list_sources`/`get_metadata` instead of going through `RdmaStrategy` — so the selector has to be wired in independently. vLLM has no equivalent second path, so there's nothing vLLM-specific to add; it already selects via the shared strategy (the `test_vllm_loader.py` cases patch `rdma_strategy.get_configured_selector` to prove it).

Added a comment at this spot in 1d6fed2 so it doesn't read as sglang-only.

zhengluo-nv (approved):

yixinhuang48:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/ada0c08dfab970de1c48d30d96a0c6691571c30c

zhengluo-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/ada0c08dfab970de1c48d30d96a0c6691571c30c

AndyDai-nv (approved):
Can wait for a new CI rounds

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/ada0c08dfab970de1c48d30d96a0c6691571c30c

KavinKrishnan:
/ok to test [31cb84a](https://github.com/ai-dynamo/modelexpress/pull/461/commits/31cb84a439e45185f31949353acb84538220b0dd)

PR #474, opened 2026-07-01
test(ci): make per-rank source-agent assertion DP-aware

AndyDai-nv:
## Summary

Fixes the flaky **P2P DP test (vllm DP=2 single-node)** →
`test_per_rank_source_agents`, which fails ~50% of runs with e.g.:

```
Expected 1 distinct source agent(s), got 2:
{mx-auto-worker0-45390a7e, mx-auto-worker0-caf48b7d}  source_ranks=[0]
```

## Root cause

`test_per_rank_source_agents` asserted `len(distinct_agents) == tp_size`.
That equality is only correct for **pure TP**, where each `worker_rank`
owns a distinct shard and the source publishes exactly one candidate per
rank.

Under **data parallelism** the source runs `dp_size` interchangeable full
copies at the **same** `worker_rank` (0). Each DP core self-assigns a
random `worker_id` (`engines/vllm/adapter.py`), so the source publishes
`dp_size` distinct rank-0 NIXL agents (`mx-auto-worker0-<uuid>`). The
target's `dp_size` cores each pick one at random (default `random`
selector), so they land on **1..dp_size** distinct sources. The
distinct-agent count is therefore non-deterministic in
`[tp_size, tp_size * dp_size]`; the DP=2 job passed only when both target
cores happened to converge on the same source (a coin flip).

This is not a weight-correctness bug: the dense DP replicas are identical,
so any of them is a valid pull. `source_ranks` was always `[0]` (correct).

## Change

- Replace the equality with a range check
  `tp_size <= distinct_agents <= tp_size * dp_size`.
- Keep `source_ranks == range(tp_size)` — the deterministic invariant that
  still catches TP collapse and real source-rank mis-pairing (the class of
  bug #363 fixed via world-rank pairing).
- Add a `--dp-size` pytest option + `dp_size` fixture; thread `dp_size`
  through the `run-mx-p2p-test` action; set `dp_size: "2"` on the DP job.
- `dp_size` defaults to **1**, so the range collapses to exactly `tp_size`
  and every TP-only job (TP=1, TP=2, EP tp_size=2) is unaffected.
- Correct the `manifest-azure-dp2.yaml` header, which wrongly stated a DP
  source publishes a single rank-0 CR / the assertion expects exactly one
  agent.

## Summary by CodeRabbit

* **New Features**
  * Added configurable data-parallel test sizing, letting CI and tests run with a chosen `dp_size` value.
  * Updated automated checks to account for multiple data-parallel replicas.

* **Bug Fixes**
  * Improved per-rank agent assertions so they work correctly when data-parallelism is greater than 1.
  * Clarified expected agent-selection behavior in the Azure DP2 test setup.

* **Documentation**
  * Expanded inline test guidance to better explain how data-parallel settings affect CI expectations.

### DIFF
--- a/.github/actions/run-mx-p2p-test/action.yml
+++ b/.github/actions/run-mx-p2p-test/action.yml
@@ -90,6 +90,15 @@ inputs:
       Catches silent TP collapse and rank-pairing regressions.
     required: false
     default: '1'
+  dp_size:
+    description: |
+      Data-parallel size of the target deployment. Forwarded to pytest as
+      --dp-size. DP replicas are interchangeable copies at the same
+      worker_rank, so the per-rank agent assertion accepts a distinct-agent
+      count in [tp_size, tp_size * dp_size] instead of exactly tp_size.
+      Defaults to 1 (pure TP / single copy).
+    required: false
+    default: '1'
   measure_vram:
     description: 'When true, sample source/target VRAM and enable the pytest VRAM parity assertions.'
     required: false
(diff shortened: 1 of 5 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/474/commits/d0bfad4b0ad7cb4bd65c65a5965a7894353d001a

yixinhuang48 (approved):

tanushriya910 (approved):

PR #479, opened 2026-07-04
feat(weight_transfer): add trainer-inference weight sync with M2N and 2-D TP sharding

ganeshku1:
## Summary

- Extends `TrainerShard` with `col_start`/`col_end` fields (default `0`/`-1` = full width, fully backward-compatible) enabling 2-D tile sharding of parameter tensors
- Updates the region router to split element runs at both row and column shard boundaries, covering all trainer-side parallelism configurations that were previously unsupported
- Adds `TrainerTensor.shard_for_elem(row, col)` alongside the existing `shard_for_row(row)`
- **58 new tests across 3 files; 88/88 passing**

## Motivation

The previous `TrainerShard` protocol only modeled dim-0 (row) sharding, covering FSDP/DP trainers. Three common configurations were unsupported:

| Configuration | Before | After |
|---|---|---|
| Column-parallel TP — row-parallel weights (`o_proj`, `down_proj`) split along dim-1 | X | ✓ |
| 2-D tile (TP x FSDP hybrid) — rectangular tile per rank | X | ✓ |
| Expert Parallelism with column-split experts | X | ✓ |

### `protocol/types.py`
- `TrainerShard`: adds `col_start: int = 0` and `col_end: int = -1` (`-1` resolves to full tensor width at query time)
- `TrainerTensor`: adds `shard_for_elem(row, col)` for 2-D lookup; existing `shard_for_row(row)` unchanged

### `planner/router.py`
- `_is_row_only()`: detects when all shards span the full column range, fast-path to original behaviour
- `_shard_for_elem_2d()`: locates the owning tile and computes the local row-major offset within shard-local contiguous storage
- `_split_run_2d()`: advances element runs one column-segment at a time, splitting at column-shard and row-end boundaries
- `route_regions()`: dispatches to the correct splitting strategy per tensor

### `tests/test_weight_transfer.py` (existing, +7 cases, 30/30)
- `TestRouteRegions2D`: column-parallel TP=2, 2-D tile TP x FSDP, local offset arithmetic, backward compatibility, `shard_for_elem`

### `tests/test_rl_weight_sync.py` (new, 14 tests)
RL training loop integration (GRPO/PPO/PrimeRL pattern):
- Plan caching and invalidation across gradient steps
- GRPO N-worker rollout independence
- TP=2 column-sharded and MoE EP routing across RL steps
- M2nPlanner server fallback, per-step invalidation, worker rank forwarding

### `tests/test_e2e_nccl_m2n.py` (new, 22 tests)
NCCL M2N transport end-to-end:
- Byte accounting: no bytes dropped or double-counted across all shard configurations
- Multi-worker M2N: N workers get consistent, non-overlapping descriptor sets
- TP=2 routing: col-sharded descriptors reference both TP agents with balanced byte splits
- M2nDescriptor wire-protocol encode/decode roundtrip including 64-bit GPU address preservation
- M2nPlanner server-coordination pipeline and LocalPlanner fallback under server failure

### `tests/test_e2e_primerl.py` (new, 22 tests)
PrimeRL-integrated weight sync end-to-end:
- Full BakeRecorder -> resolve -> route pipeline with byte-exact data_ptr verification
- 10-step RL loop with plan invalidate+rebuild; multi-worker independence
- TP=2 hybrid (q_proj row-sharded + o_proj col-sharded) across 5 RL steps
- 1-D params (biases, LayerNorms) and large-param pipeline correctness

All 58 new tests are pure-Python (no NIXL, no GPU) and complete in 0.11s.

## Validation

End-to-end on devdesktop (2x NVIDIA RTX 6000 Ada, PyTorch 2.9.0+cu128) with Qwen2.5-0.5B (290 params, 988.1 MB bfloat16):
- 48 row-parallel layers split into contiguous [R, C/2] shard tensors with TP=2
- 86,378 RDMA descriptors generated by 2-D router
- 988.1 MB transferred; all 290 parameters match byte-for-byte
- 88/88 unit tests pass

### Transport benchmark (NCCL M2N vs NIXL proxy, PCIe/PHB no NVLink)

| Transport | Mean latency | Throughput |
|---|---|---|
| NCCL send/recv (2 processes) | 85.7 ms | 11.53 GB/s |
| CUDA memcpy (NIXL RDMA proxy) | 96.2 ms | 10.28 GB/s |
| torch .copy_() built-in | 98.6 ms | 10.02 GB/s |

NCCL is 12% faster than raw cudaMemcpy on PCIe-only hardware.

## Notes

The router assumes each shard is stored contiguously (as it is on every real TP rank). Strided views into a full tensor are not valid shard inputs.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -715,6 +715,99 @@ while agent.check_xfer_state(handle) not in ("DONE", "SUCCESS"):
 agent.release_xfer_handle(handle)
 ```
 
+## Trainer-Inference Weight Sync (WeightSyncService)
+
+ModelExpress includes a `WeightSyncService` gRPC service that coordinates live
+weight synchronization from a sharded trainer to inference workers.  The Python
+client exposes `PullRole` (inference worker pulls from trainer) and `PushRole`
+(trainer pushes to inference workers).
+
+### Protocol Overview
+
+```mermaid
+sequenceDiagram
+    participant T as Trainer
+    participant MX as MX Server (WeightSyncService)
+    participant W as Inference Worker
+
+    T->>MX: PublishTrainerTable(model_key, table)
+    W->>MX: BuildPlan(plan_key, model_key, regions)
+    MX-->>W: BuildPlanResponse(plan_id)
+    W->>MX: GetPlan(plan_id)
+    MX-->>W: GetPlanResponse(ready, descriptors)
+    W->>T: NIXL READ (RdmaDescriptors)
+```
+
+### Planner Abstraction
+
+Three planner implementations live in `modelexpress/weight_transfer/planner/`:
+
+| Class | When to use |
+|-------|-------------|
+| `LocalPlanner` | No MX server available; each worker routes independently. Per-worker cache only. |
+| `ServerPlanner` | MX server available. One worker routes; all workers with the same `plan_key` reuse the cached plan. Falls back to `LocalPlanner` on server error. |
+| `M2nPlanner` | All workers register simultaneously. MX routes all workers' regions in one pass and returns a globally-consistent M2N plan. Falls back to `LocalPlanner` on error. |
+
+### Key Data Flow (PULL)
+
+1. **Bake pass**: `PullRole.initialize()` drives the engine's weight loader with
+   `LazyWeight` tensors that record op chains without materializing data.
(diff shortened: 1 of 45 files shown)


ganeshku1:
Closing to reopen as a clean PR with reduced LOC refactoring included.

PR #480, opened 2026-07-04
feat: filter P2P weight sources by runtime accelerator

yafshar:
## Summary

Expose the active accelerator backend in P2P worker metadata and filter
incompatible RDMA source candidates before target preparation, preventing
cross-accelerator weight transfers from being attempted.

- Add an `accelerator` field to `WorkerMetadata` and
  `GetTensorManifestResponse` (protobuf), and persist it through both the Redis
  and Kubernetes metadata backends.
- Filter RDMA source candidates on the target side by comparing the published
  accelerator value with the local `AcceleratorBackend.name`
  (`RdmaStrategy._accelerator_compatible`).
- Keep accelerator compatibility independent of `SourceIdentity`, preserving
  the stability of `mx_source_id` and pinned source-ID hashes.
- Treat missing or empty accelerator metadata as compatible to support rolling
  upgrades and sources that predate this field.
- Regenerate the checked-in Python and Go P2P protobuf bindings.
- Update the documentation to describe the new metadata field and target-side
  compatibility filtering behavior.

## Motivation

In mixed-accelerator fleets, targets must not attempt RDMA weight transfers from
sources running on incompatible accelerator backends. Transfers between
compatible accelerators are unaffected; only incompatible source-target pairs
are filtered out.

## Testing

- `cargo test -p modelexpress-server p2p`
- `cd modelexpress_client/python && pytest tests/`

Additionally, validated end-to-end on a two-node RDMA setup:

- Verified that the source published its accelerator in worker metadata (the
  field was absent before this change, confirming the new publication path).
- Verified that the target selected the source based on accelerator
  compatibility and did not trigger an accelerator-mismatch skip.
- Verified successful RDMA weight transfer, followed by target startup and
  serving.

## Summary by CodeRabbit

* **New Features**
  * Added accelerator-aware worker metadata throughout P2P publishing and discovery.
  * Compatibility checks now use runtime accelerator info to help select matching sources.

* **Bug Fixes**
  * Improved source selection to skip incompatible workers before transfer begins.
  * RDMA loading now falls back more predictably when a source is missing metadata or does not match the target accelerator.
  * Discovery now better handles stale or outdated worker information.

* **Documentation**
  * Updated architecture and metadata docs to describe accelerator-based compatibility and the revised loading flow.


yafshar:
@zhengluo-nv cargo audit failing! it is a new RUSTSEC-2026-0204 on crossbeam-epoch, needs a dep bump on main.

zhengluo-nv (approved):

PR #481, opened 2026-07-05
feat(weight_transfer): add trainer-inference weight sync with M2N and 2-D TP sharding

ganeshku1:
## Summary

- Adds a modular `weight_transfer` sub-package for trainer → inference weight sync with two transfer modes: **PULL** (inference workers read from trainer) and **PUSH** (trainer writes to inference workers)
- Implements **2-D tile sharding** for TP × FSDP hybrid trainers: row-parallel layers (o_proj, down_proj) are column-sharded so each TP rank holds a contiguous `[R, C/tp]` tile
- Adds **M2N many-to-many transport** via the MX server as a barrier coordinator: all workers register simultaneously, the server routes once, each worker gets its descriptor slice
- Includes **LocalPlanner** (pure-math, client-side) and **ServerPlanner** (offloads routing to MX WeightSyncService in Rust) with automatic fallback

## Tests

**88 tests, 0 GPU required for CI** (pure-Python stubs):

| File | Tests | Coverage |
|---|---|---|
| `test_rl_weight_sync.py` | 14 | RL loop lifecycle, GRPO multi-worker, MoE EP, M2nPlanner fallback |
| `test_e2e_nccl_m2n.py` | 22 | M2N descriptor accounting, wire-protocol roundtrip, TP=2 routing |
| `test_e2e_primerl.py` | 22 | Full bake→resolve→route pipeline, PrimeRL sync lifecycle |
| `tests/gpu/test_e2e_primerl_real_model.py` | 17 | Real Qwen/Qwen2.5-0.5B (290 params, 988 MB bfloat16), 2× CUDA GPU |

**Real-model GPU tests** (`tests/gpu/`, skipped in CI, run on devdesktop with 2× RTX 6000 Ada):
- FSDP: byte-exact transfer of all 290 params
- TP=2: 48 row-parallel params col-sharded across both agents, verified byte-exact
- 3-step RL loop: noise injection → sync → verify → plan invalidate

## Test plan

- [ ] `python -m pytest tests/test_rl_weight_sync.py tests/test_e2e_nccl_m2n.py tests/test_e2e_primerl.py` — 58 tests, no GPU
- [ ] `python -m pytest tests/gpu/test_e2e_primerl_real_model.py` — 17 tests, requires 2× CUDA GPUs
- [ ] CI Python Tests check passes

## Summary by CodeRabbit

* **New Features**
  * Added trainer-to-inference weight sync support, including live weight updates, coordinated routing, and multi-worker plan handling.
  * Added support for both pull-based and push-based sync flows, with caching and automatic plan refresh when weights change.
  * Expanded configuration options for sync timeouts and server/Redis endpoints.

* **Bug Fixes**
  * Improved routing for both row-sharded and 2D-sharded model weights to better match tensor-parallel layouts.

* **Tests**
  * Added broader end-to-end coverage for routing, sync behavior, and descriptor serialization.


modelexpress_server/src/weight_sync/router.rs
```diff
@@ -0,0 +1,245 @@
...
+        })
+        .collect()
+}
+
+// ---------------------------------------------------------------------------
+// Internal helpers
+// ---------------------------------------------------------------------------
+
+fn unpack_runs(flat: &[i64]) -> Vec<(i64, i64)> {
+    flat.chunks(2).map(|c| (c[0], c[1])).collect()
```
nicolasnoble: `chunks(2)` yields a 1-element final chunk on odd-length input, so `c[1]` panics, and this comes straight off the wire. `chunks_exact(2)`, or validate even length and return `invalid_argument`.

modelexpress_server/src/weight_sync/service.rs
```diff
@@ -0,0 +1,387 @@
...
+    ) -> Result<Response<InvalidatePlanResponse>, Status> {
+        let req = request.into_inner();
+        let mut state = self.state.write().await;
+        if let Some(plan_id) = state.plan_key_to_id.remove(&req.plan_key) {
+            state.plans.remove(&plan_id);
+        }
+        Ok(Response::new(InvalidatePlanResponse { ok: true }))
+    }
+
+    async fn register_m2n_worker(
```
nicolasnoble, line 288: No timeout or reaper on this barrier. If a worker never registers, everyone else polls forever and the pending entry leaks. The p2p service has a reaper; this one doesn't. Can we stamp entries with a creation time and reap stale ones?

modelexpress_server/src/weight_sync/service.rs
```diff
@@ -0,0 +1,387 @@
...
+
+            entry.registrations.insert(
+                req.worker_rank,
+                M2nWorkerRegistration {
+                    regions: req.regions,
+                    _nixl_metadata: req.nixl_metadata,
+                },
+            );
+
+            if entry.registrations.len() < entry.total_workers as usize {
```
nicolasnoble, line 352: `total_workers` is an int32 off the wire; a negative value casts to a huge `usize` and the barrier can never fire. A late worker also gets a plan_id but no slice and polls forever. Reject `total_workers <= 0` and unknown-rank late registrations.

modelexpress_server/src/weight_sync/router.rs
```diff
@@ -0,0 +1,245 @@
...
+        dst_rem = count;
+    }
+
+    while let Some(shard) = cur_shard {
+        if dst_rem == 0 {
+            break;
+        }
+
+        let count = src_rem.min(dst_rem);
+        let src_addr = shard.device_addr + (src_rel * element_size) as u64;
```
nicolasnoble: This address math runs on unvalidated wire ints, so it overflow-panics in debug and wraps in release into a bad descriptor. `checked_add`/`checked_mul` and reject on overflow.

modelexpress_client/python/modelexpress/load_strategy/trainer_pull_strategy.py
```diff
@@ -0,0 +1,234 @@
...
+
+        try:
+            self._pull_role.sync()
+        except Exception as e:
+            raise StrategyFailed(f"Initial RDMA pull failed: {e}", mutated=True) from e
+
+        return result
+
+    def update_weights(self, ctx: LoadContext) -> None:
+        """Execute a weight sync using the pre-built static plan.
```
nicolasnoble, line 165: No rollback here. `sync()` writes into live param memory, so if it throws partway the model's left with mixed old/new weights and keeps serving. There's a `rollback()` on the strategy but nothing calls it on this path. Stage into scratch and commit atomically, or force a reinit on any sync failure?

modelexpress_client/python/modelexpress/weight_transfer/transport/nixl_executor.py
```diff
@@ -0,0 +1,132 @@
...
+                remote_xfer_side=src_prepped if operation == "READ" else dst_prepped,
+                remote_indices=indices,
+                backends=backends,
+            )
+            agent.transfer(handle)
+            handles.append(handle)
+            total_bytes += sum(d.nbytes for d in descs)
+
+        wait_start = time.monotonic()
+        for handle in handles:
```
nicolasnoble: On timeout/ERR we release only the current handle and raise; the rest of `handles` leak and their transfers keep writing GPU memory after we've bailed. Needs a `try/finally` that releases everything and aborts outstanding transfers.

nicolasnoble (commented):
Went through this properly. Inline comments are the code-level stuff, roughly blocker-first. Two I'd really want fixed before this merges: the M2N barrier has no timeout or reaper, so one inference worker that never registers hangs the whole collective and leaks the session, and the pull refit writes into live param memory with no rollback, so a mid-sync failure leaves the engine serving mixed old/new weights.

Few housekeeping things that don't need inline threads:
- `protocol/test_weight_transfer.py` is byte-identical to `tests/test_weight_transfer.py` and lives inside the package, so `include=["modelexpress*"]` ships it in the wheel and the suite double-counts it. Drop the copy, pin `testpaths`.
- `tests/test_e2e_nccl_m2n.py` is named for NCCL e2e transport but it's fully mocked, no NCCL and no transport. Rename or drop, especially now that #497 has real ones.
- `protocol/router.py` is a dead byte-identical copy of `planner/router.py`, and `_shard_width` looks unused.
- The new `WeightLoaderAdapter` + `_CtxEngineAdapter` duplicate `EngineAdapter` without adding a capability it lacks - could fold in.
- The four new env vars (`MX_TRAINER_TABLE_KEY`, `MX_TRAINER_SYNC_TIMEOUT`, `MX_REDIS_URL`, `MX_WEIGHT_SYNC_SERVER`) aren't in `docs/DEPLOYMENT.md`.

Happy to talk through any of it.

modelexpress_client/python/modelexpress/weight_transfer/planner/router.py
```diff
@@ -0,0 +1,234 @@
...
+    """Route resolved element-run regions to NIXL RDMA descriptors.
+
+    Maps every element offset to a trainer shard GPU address and pairs it
+    with the destination vLLM parameter address.  The same algorithm is
+    implemented in Rust in modelexpress_server/src/weight_sync/router.rs.
+    """
+    descriptors: list[RdmaDescriptor] = []
+
+    for region in regions:
+        trainer_tensor = table.tensor_by_name(region.tensor_name)
```
nicolasnoble, line 239: A few silent-drop paths: a missing tensor `continue`s here, a missing src shape skips in `resolver.py` (~L94), and the src/dst zip stops when either side exhausts with no total-equality check. Each produces a partial plan that transfers as success and leaves the param stale. Assert coverage + `src==dst` at plan build and make these hard errors? PUSH validates sizes, PULL doesn't.

modelexpress_client/python/modelexpress/weight_transfer/planner/router.py
```diff
@@ -0,0 +1,234 @@
...
+    For 2-D tile shards the local offset is:
+        (row - row_start) * shard_width + (col - col_start)
+    This invariant ensures device_addr + local_off*elem_size is the correct
+    byte address for any element in a tile.
+    """
+    for shard in shards:
+        col_end = _resolved_col_end(shard, full_width)
+        if shard.row_start <= row < shard.row_end and shard.col_start <= col < col_end:
+            w = col_end - shard.col_start
+            local_off = (row - shard.row_start) * w + (col - shard.col_start)
```
nicolasnoble, line 75: This assumes physical row stride equals tile width, but `row_bytes` says otherwise for a padded/column-view tile, so every row after the first addresses wrong. Use `row_bytes` as the stride, or assert `row_bytes == width * elem_size`.

modelexpress_client/python/modelexpress/weight_transfer/planner/resolver.py
```diff
@@ -0,0 +1,123 @@
...
+    A fully-contiguous tensor produces exactly one run.
+    """
+    if len(shape) == 0:
+        return [(storage_offset, 1)]
+    if math.prod(shape) == 0:
+        return []
+
+    runs: list[tuple[int, int]] = []
+
+    def _collect(base: int, dim: int) -> None:
```
nicolasnoble, line 59: Builds one tuple per element before merging; a contiguous `lm_head`/embedding is hundreds of millions of tuples in a Python loop. Worth a contiguous fast path so it doesn't OOM on real tensors.

modelexpress_client/python/modelexpress/weight_transfer/engine/lazy.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def bake_model(
+    model: Any,
+    weight_iter: Iterator[tuple[str, LazyWeight]],
+) -> list[RecordedCopy]:
+    """Run model.load_weights() with LazyWeights and return recorded copies."""
+    recorder = BakeRecorder()
+    with recorder:
+        if hasattr(model, "load_weights"):
```
nicolasnoble: When the model has no `load_weights()` this returns `[]` and the whole sync reports success having moved nothing, so the engine serves uninitialized weights. Should raise on a zero-byte plan.

modelexpress_client/python/modelexpress/weight_transfer/roles/pull.py
```diff
@@ -0,0 +1,121 @@
...
+
+        self._descriptors = self._planner.build(regions, table, self._plan_key)
+        logger.info(
+            "[Worker %d] Plan built: %d RDMA descriptors, %.2f GB total",
+            self._worker_rank,
+            len(self._descriptors),
+            sum(d.nbytes for d in self._descriptors) / 1e9,
+        )
+
+        remote_agents: dict[int, str] = {}
```
nicolasnoble, line 88: `add_remote_agent` runs every `initialize()` but `teardown()` never deregisters, so a trainer that reshards repeatedly accumulates registrations. Track and remove them? (Same on the push side, `push.py` ~L75.)

modelexpress_client/python/modelexpress/load_strategy/trainer_pull_strategy.py
```diff
@@ -0,0 +1,234 @@
...
+from ..weight_transfer.planner.server import ServerPlanner
+
+if TYPE_CHECKING:
+    from ..weight_transfer.protocol.types import TrainerTable
+    from ..weight_transfer.engine.base import WeightLoaderAdapter as WtAdapter
+
+logger = logging.getLogger("modelexpress.strategy_trainer_pull")
+
+
+def _use_server_planner() -> bool:
```
nicolasnoble, line 54: `bool(os.environ.get("MX_WEIGHT_SYNC_SERVER", ""))` treats `=0` and `=false` as enabled, since any non-empty string is truthy. Parse through the `envs` registry with real truthy semantics.

modelexpress_server/src/weight_sync/service.rs
```diff
@@ -0,0 +1,387 @@
...
+/// Pending M2N plan accumulating worker registrations until the barrier fires.
+struct M2nPlanEntry {
+    total_workers: i32,
+    registrations: HashMap<i32, M2nWorkerRegistration>,
+    /// Set once all workers have registered and the plan is built.
+    plan_id: Option<String>,
+}
+
+/// Shared server state (plan cache + raw table blobs).
+struct WeightSyncState {
```
nicolasnoble, line 67: These maps are insert-only with no TTL and superseded plans get orphaned, so it's a slow leak over a long run. Evict on retarget/invalidate.

nicolasnoble (approved):

modelexpress_server/src/weight_sync/service.rs
nicolasnoble: Added eviction on retarget via invalidate_model, so superseded plans go when the trainer table is republished. A general time-based TTL across plans and the trainer and inference tables I left out. It needs a lifetime policy and an MX_ env knob to be tunable, which is more than this PR should decide. The m2n barrier does have a timeout now, currently a module-local 600s const.

PR #494, opened 2026-07-07
fix(client): register SGLang P2P target tensors in post-processed layout

nv-hwoo:
## Problem

SGLang P2P weight transfer silently corrupts DeepSeek-V4-Pro on the target.
The target reaches Ready, passes `/health`, and answers requests, but emits
token salad. The source (disk-loaded) is coherent with identical flags
(TP=8, EP=8, `--moe-a2a-backend deepep`, fp8 block-quant checkpoint).

## Root cause

The SGLang adapter registers tensors at different lifecycle stages on the
two roles, and the `.__storage` naming convention makes that visible as a
name mismatch:

- `_collect_tensors` registers contiguous parameters under their plain name
  and non-contiguous parameters as a byte view of their underlying storage
  named `<name>.__storage`.
- On the **source**, registration happens after a full disk load. SGLang's
  `DeepseekV4ForCausalLM.post_load_weights()` (called from the tail of
  `load_weights`, `sglang/srt/models/deepseek_v4.py:2290` at 0.5.13.post1)
  runs `_setup_fp8_wo_a_scales`, which replaces every layer's MLA
  `wo_a.weight_scale_inv` with a DeepGEMM-layout
  (`transform_sf_into_required_layout`) tensor that is **non-contiguous**.
  The source therefore publishes these as `...weight_scale_inv.__storage`.
- On the **target**, `before_rdma_receive` replayed only the per-module
  `quant_method.process_weights_after_loading` sweep; `post_load_weights`
  was deferred to `after_rdma_receive`. At registration time the same
  tensors were still **contiguous** and registered under their plain names.
- `match_tensors` matched by name and silently skipped both sides: 61
  fp8 scale tensors per rank (one `wo_a.weight_scale_inv` per layer) were
  never transferred and kept their random init values, mis-scaling every
  layer's attention output projection.

vLLM is unaffected (its pipeline registers both roles in the same state);
MoE / expert parallelism / DeepEP are uninvolved despite the initial
suspicion: the corrupted tensors are attention-side.

## Fix

1. **`modelexpress/engines/sglang/adapter.py`**: `before_rdma_receive` now
   runs `model.post_load_weights()` **before** the per-module quant sweep,
   mirroring the disk path's order, so the target registers tensors in the
   exact state the source published them from. The `after_rdma_receive`
   override is removed (`post_load_weights` is not idempotent for DSV4:
   `_setup_fp8_wo_a_scales` re-views an already-transformed tensor).
2. **`modelexpress/nixl_transfer.py`**: `match_tensors` now raises
   `ManifestMismatchError` when the source manifest and local registration
   name sets diverge, instead of silently skipping unmatched tensors and
   serving partially transferred weights.

## Validation

DeepSeek-V4-Pro, TP=8 EP=8 DeepEP, fp8 block-quant (ue8m0), B200 x2,
source + target pair, image pinned by digest:

| Gate | Before | After |
|---|---|---|
| Tensors matched per rank | 2199 / 2260 | 2260 / 2260 |
| Unmatched name sets | 61 local-only + 61 source-only | 0 |
| RDMA transfer | 113.03 GB/rank, ~4.1 s, ~220 Gbps | unchanged (61 extra small tensors ride along) |
| Disk fallback lines | 0 | 0 |
| Target output | token salad | coherent (sanity 10/10, verified by reading responses) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved weight-loading reliability during transfers by validating tensor names on both sides and failing fast when there’s a mismatch, instead of silently skipping missing entries.
  * Updated the load flow so all weight post-processing now happens in a single, consistent step after receiving weights.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/sglang/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/sglang/adapter.py
@@ -111,10 +111,8 @@ class SglangAdapter(EngineAdapter):
         return tensors
 
     def before_rdma_receive(self, result: LoadResult) -> LoadResult:
+        result = self._post_load_weights(result)
         return self._process_weights_after_loading(result)
-
-    def after_rdma_receive(self, result: LoadResult) -> LoadResult:
-        return self._post_load_weights(result)
 
     def apply_weight_iter(
         self,
--- a/modelexpress_client/python/modelexpress/nixl_transfer.py
+++ b/modelexpress_client/python/modelexpress/nixl_transfer.py
@@ -569,6 +569,19 @@ class NixlTransferManager:
         matched_tensors = len(remote_descs)
         match_time = time.perf_counter() - match_start
 
+        # Name-set diff between the source manifest and the locally registered
+        # tensors.
+        src_names = {s.name for s in source_tensors}
+        local_only = sorted(set(self._tensors) - src_names)
+        source_only = sorted(src_names - set(self._tensors))
+        if local_only or source_only:
+            raise ManifestMismatchError(
+                f"Tensor name mismatch between source manifest and local "
+                f"registration: {len(local_only)} local-only, "
+                f"{len(source_only)} source-only. Refusing to serve "
+                f"partially transferred weights. "
+            )
+
         if not remote_descs:
             logger.warning("No matching tensors found for transfer")
             return 0, 0, 0.0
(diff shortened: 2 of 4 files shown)


AndyDai-nv (commented):
seems like there are ci failures

nv-hwoo:
Yeah seems like there was some glitch in the CI system. Rerunning passed.

AndyDai-nv (approved):

PR #497, opened 2026-07-09
feat(weight_transfer): nccl_m2n collective reshard transport

kwen2501:
# nccl_m2n reshard transport for weight_transfer

Stacked on #481 (`ganeshku/weight-transfer-trainer-inference-sync`). This PR adds
a collective **nccl_m2n** transport backend to the `weight_transfer` sub-package
and proves it end-to-end on 2xH100.

## Motivation

PR #481 syncs trainer -> inference weights and structures the pipeline around a
transport `Executor` behind `PushRole` (trainer = source) and `PullRole`
(generator = destination). It leaves room for a native many-to-many transfer
primitive that moves the whole redistribution in a single collective call.

`nccl_m2n` (`~/nccl-extensions/nccl_m2n`) is that primitive.
`ncclReshardWithWindow(comm, window, src, dst, stream)` reshards a global tensor
from a source mesh to a destination mesh in one zero-copy GPU collective over
NCCL's user-window API. This PR wires it in as a first-class transport backend.

## Key design finding: a collective, mesh-driven transport

`ncclReshardWithWindow` takes both `src` and `dst` `ncclDistTensor_t` on **every**
rank and is **collective** -- trainer and generator ranks enter it together:

- Trainer (source mesh): `src.dataPtr = param tile`, `dst.dataPtr = NULL`
- Generator (destination mesh): `dst.dataPtr = weight buffer`, `src.dataPtr = NULL`
- Each rank carries *both* meshes (topology only on the non-owned side)

So the trainer/generator identity from #481 is preserved, but the transport is
collective rather than one initiator acting alone. Rather than a new `SyncMode`,
`PushRole` and `PullRole` each gain a **`reshard()`** method that both sides
co-call in the same step.

```mermaid
sequenceDiagram
    participant T as Trainer (PushRole, src mesh)
    participant G as Generator (PullRole, dst mesh)
    Note over T,G: shared ncclComm_t (ranks [0,tp_src) src, [tp_src,world) dst)
    T->>T: stage param tile -> symmetric window
    T-->>G: ncclReshardWithWindow(src=tile, dst=NULL)
    G-->>T: ncclReshardWithWindow(src=NULL, dst=window)
    G->>G: window -> live weight buffer
```

Properties of this backend:

- `nccl_m2n` **routes internally** from the src/dst meshes, so no client-side
  descriptor routing is needed on this path. Large simplification.
- Needs one `ncclComm_t` spanning all src+dst ranks (**bootstrapped fresh**) and
  symmetric `ncclMemAlloc` buffers with an `ncclCommWindowRegister` window
  (zero-offset contract).

## Integration point

The **transport `Executor`** is the seam. `NcclM2nExecutor` is a sibling backend
that consumes *meshes* instead of byte descriptors:

| Aspect | nccl_m2n backend (this PR) |
|---|---|
| Executor | `NcclM2nExecutor` (per-param meshes) |
| Routing | internal to `nccl_m2n` (from src/dst meshes) |
| Connection | shared `ncclComm_t`, fresh bootstrap over torch.distributed |
| Buffers | symmetric window + `param <-> window` staging |
| Call style | collective (trainer and generator co-call) |

## What's implemented

New files under `modelexpress/weight_transfer/`:

- **`transport/_nccl_m2n_bind.py`** -- ctypes binding over `libnccl_m2n.so` +
  `libnccl.so` (vendored from the nccl MR-2984 PoC), plus `bootstrap_comm_from_torch`
  (fresh `ncclComm_t` via `ncclGetUniqueId` broadcast over torch.distributed +
  `ncclCommInitRank`) and a device-to-device staging copy.
- **`transport/nccl_m2n_executor.py`** -- `NcclM2nExecutor`: per-param mesh
  construction, symmetric-window staging (`param -> window -> reshard -> window ->
  param`, honoring the zero-offset contract), and `ncclReshardWithWindow`. Plus
  `ReshardParam`, `build_reshard_params`, `torch_dtype_to_nccl`.
- **`planner/mesh.py`** -- pure-math (torch-free) mesh mapping: `Mesh`,
  `build_tp_meshes`, `tile_shape`, and `shard_dim_from_trainer_tensor` (reads
  #481's `TrainerTensor` row/col shard layout into `SHARD(dim)` / `REPLICATE`).

Modified:

- **`roles/pull.py`**, **`roles/push.py`** -- add the collective `reshard()`.
- **`transport/__init__.py`** -- export `NcclM2nExecutor`, `ReshardParam`.

## Testing

- **No-GPU golden** (`tests/test_nccl_m2n_reshard.py`, 7 tests) -- mesh inference,
  tile shapes, and gather+reslice byte-parity. Passes under real torch 2.12.
- **2-GPU e2e** (`tests/gpu/run_reshard_e2e.py`) -- fresh-comm bootstrap ->
  symmetric window -> `ncclReshardWithWindow` -> byte-compare dst tile vs. golden.

Validated on EOS (2xH100), building the vendored NCCL from source (CUDA 12.9) and
`libnccl_m2n.so` (sm_90), running through the ModelExpress binding + executor:

```
[rank 0] comm bootstrapped, world=2 tp_src=1 tp_dst=1
[rank 1] RESHARD PASS dst_index=0 got[0,:4]=[0.0, 1.0, 2.0, 3.0] exp[0,:4]=[0.0, 1.0, 2.0, 3.0]
```

`LD_PRELOAD` of the vendored `libnccl.so.2` unifies torch and `libnccl_m2n.so` on
the same from-source NCCL, avoiding a dual-NCCL ABI clash; a gloo control-plane
process group keeps torch from standing up its own NCCL communicator.

## Scope and follow-ups

This is an **executor spike** -- deliberately the smallest slice that proves the
path. Deferred:

- The e2e ran `tp_src=tp_dst=1` (full path, full-tensor move). The harness is
  parameterized by `TP_SRC`/`TP_DST`; a 4-GPU 2->2 run exercises real sharded M->N.
- Cross-dim (transpose) reshards, EP/MoE per-expert, and non-matching dtypes.
- MX-server-brokered uniqueId broadcast via the existing `RegisterM2nWorker`
  barrier (replacing the torch.distributed bootstrap for true disaggregation).
- Strategy env switch in `load_strategy/trainer_pull_strategy.py`.
- Packaging `libnccl_m2n.so` (build/vendor) so ctypes resolves it without manual
  `LD_LIBRARY_PATH`.

## Summary by CodeRabbit

- **New Features**
  - Added NCCL-based tensor resharding between different tensor-parallel configurations.
  - Added support for sharded and replicated mesh layouts, including shape calculation and validation.
  - Added push and pull reshard operations with transfer metrics.
  - Added public transport support for NCCL M2N execution.

- **Tests**
  - Added CPU coverage for resharding layout and tensor-shape calculations.
  - Added a two-GPU end-to-end validation for NCCL resharding.


modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+            )
+
+            rc = self._m2n.reshard(self._comm, self._window, src_t, dst_t, self._stream)
+            if rc != binding.ncclSuccess:
+                raise RuntimeError(f"ncclReshardWithWindow({p.name!r}) rc={rc}")
+
+            # Copy the resharded tile out of the window into the live param.
+            if not self._is_src:
+                self._m2n.device_synchronize()
+                self._m2n.memcpy_dtod(p.local_ptr, self._window_buf, p.local_nbytes)
```
nicolasnoble: This writes each resharded tile straight into live `p.local_ptr` in the loop; a failure at L143 leaves tiles 0..k-1 new and k..N old in the running model with no rollback. Stage into scratch and commit atomically.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    from math import prod
+
+    named = dict(model.named_parameters())
+    params: list[ReshardParam] = []
+    window_bytes = 0
+
+    for tt in table.tensors:
+        param = named.get(tt.name)
+        if param is None:
```
nicolasnoble, line 297: The plan is intersected with this rank's `named_parameters()` (params skipped here) and `window_bytes` is computed from local params, but `reshard`/`window_register` are collectives needing identical calls and window sizes on every rank. Different param subsets across workers deadlock or corrupt the registration. Build the plan and window size from the shared `TrainerTable`, not local params.

nicolasnoble (commented):
Went through this too. Inline comments are the code-level things, roughly production-gating first; the two biggest for me are the torn-refit with no rollback and the cross-rank window-size divergence, both of which can corrupt or hang a real multi-worker run.

Couple of things at the suite level rather than inline:
- The only byte-exactness check (`torch.equal`) is in `gpu/run_reshard_e2e.py`, which is a standalone script rather than a pytest test, so CI never runs it - the no-GPU golden tests validate tile shapes and torch's own chunk/cat inverse, not that a weight lands correctly. Would be good to get a GPU-gated CI test on the byte path.
- `tests/test_nccl_m2n_reshard.py` (real) now sits next to #481's `tests/test_e2e_nccl_m2n.py` (fully mocked, no NCCL) - two files whose names both read as nccl-m2n but validate very different things. Worth renaming the mock so nobody trusts it as an e2e.
- `reshard()` is wired only through the GPU test - nothing in shipped code constructs `NcclM2nExecutor` or calls `teardown()`, so the comm/window lifecycle is the future caller's job. Fine for a first slice, just flagging it.

Happy to talk any of it through.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+            dst_mesh = binding.make_mesh(
+                dst_mesh_dc.dims, dst_mesh_dc.start_rank, dst_mesh_dc.placement
+            )
+
+            src_local = tile_shape(p.global_shape, src_mesh_dc)
+            dst_local = tile_shape(p.global_shape, dst_mesh_dc)
+
+            # Stage the owned tile into the window base before the collective.
+            if self._is_src:
+                self._m2n.memcpy_dtod(self._window_buf, p.local_ptr, p.local_nbytes)
```
nicolasnoble: Staging (here) and copy-out (L148) run on the legacy default stream, but `reshard` (L141) runs on `self._stream`, with no event sync between window reuses. A caller passing a real non-blocking torch stream gets the window overwritten mid-collective, silently. Async copies on `self._stream` + an event sync, or double-buffer.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+            src_ptr = self._window_buf if self._is_src else 0
+            dst_ptr = self._window_buf if not self._is_src else 0
+            src_t = binding.make_tensor_desc(
+                src_ptr, src_local, p.ndims, p.dtype_nccl, src_mesh
+            )
+            dst_t = binding.make_tensor_desc(
+                dst_ptr, dst_local, p.ndims, p.dtype_nccl, dst_mesh
+            )
+
+            rc = self._m2n.reshard(self._comm, self._window, src_t, dst_t, self._stream)
```
nicolasnoble: This rc is the enqueue result, not completion, and there's no `ncclCommGetAsyncError` anywhere, so an async NCCL failure copies an un-completed window into live params while `rc == ncclSuccess`. Query the async error before trusting the tile.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+        max_cta: int | None = None,
+    ) -> None:
+        self._m2n = m2n
+        self._comm = comm
+        self._rank = rank
+        self._tp_src = tp_src
+        self._tp_dst = tp_dst
+        self._device_id = device_id
+        self._stream = stream
+        self._is_src = rank < tp_src
```
nicolasnoble: Roles are assigned by `rank < tp_src` here and the groups are hardcoded 0-based in `mesh.py` (`start_rank` 0 / `tp_src`, ~L84-88). A participant set that's dense but not `[0, tp_src)` inverts producer/consumer silently. Derive roles and groups from the actual participant ranks.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def _ensure_window(self, nbytes: int) -> None:
+        """(Re)allocate the symmetric window if it is too small.
+
+        The window is collectively registered, so every rank must size it to the
+        same value -- callers pass a world-consistent worst case.
+        """
+        if self._window is not None and self._window_bytes >= nbytes:
+            return
+        if self._window is not None:
```
nicolasnoble: If `mem_alloc`/`window_register` (L99) raises after the deregister here, `self._window`/`self._window_buf` still point at the freed buffer, so the next `teardown` double-deregisters. Null them after freeing.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+        logger.info(
+            "reshard complete: %d params, %.2f GB in %.3fs (%.1f Gbps)",
+            len(params),
+            total_bytes / 1e9,
+            elapsed,
+            gbps,
+        )
+        return total_bytes, elapsed
+
+    def teardown(self) -> None:
```
nicolasnoble, line 268: `teardown` deregisters the window and finalizes but never `comm_destroy`s, so the bootstrapped `ncclComm` leaks every resync cycle. And `execute` (L110+) has no `try/finally`, so a mid-loop failure leaks the window too. Wrap execute and destroy the comm here.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,239 @@
...
+            self._window_buf = None
+            self._window_bytes = 0
+        self._m2n.finalize()
+
+
+def torch_dtype_to_nccl(dtype) -> int:
+    """Map a torch dtype to its ncclDataType_t enum (src and dst must match)."""
+    import torch
+
+    table = {
```
nicolasnoble: `torch_dtype_to_nccl` omits the fp8 types the binding defines (`ncclFloat8e4m3/e5m2`), so a valid fp8 layout raises here. Worth adding them.

modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_m2n_bind.py
```diff
@@ -0,0 +1,295 @@
...
+        rc = self.m2n.ncclM2nInit(ctypes.byref(cfg))
+        assert rc == ncclSuccess, f"ncclM2nInit rc={rc}"
+
+    def finalize(self):
+        self.m2n.ncclM2nFinalize()
+
+    def mem_alloc(self, nbytes: int) -> int:
+        p = ctypes.c_void_p()
+        rc = self.nccl.ncclMemAlloc(ctypes.byref(p), nbytes)
+        assert rc == ncclSuccess, f"ncclMemAlloc rc={rc}"
```
nicolasnoble: All the FFI status checks in this file use `assert rc == ncclSuccess`, which the interpreter strips under `python -O`, so in an optimized build a failed alloc/register returns a garbage pointer and sails through (same class as the Rust side's `unwrap()` ban). Also `reshard` (L224) returns its rc instead of raising like the others, and `cudaSetDevice` (L281) is unchecked. Suggest explicit raises throughout.

modelexpress_client/python/modelexpress/weight_transfer/planner/mesh.py
```diff
@@ -0,0 +1,120 @@
...
+def shard_dim_from_trainer_tensor(tensor: TrainerTensor) -> int:
+    """Infer the single sharded tensor dim from a ``TrainerTensor``'s shards.
+
+    Returns ``REPLICATE`` when a single shard covers the full tensor, ``0`` when
+    the shards partition rows (row-parallel-loaded / column-parallel weights),
+    or ``1`` when they partition columns (row-parallel weights, 2-D-col tiles).
+    Raises when the layout shards both dims at once (a later 2-D slice).
+    """
+    shards = tensor.shards
+    if not shards or len(shards) == 1:
```
nicolasnoble, line 101: Single shard collapses to `REPLICATE` on a pure count check here with no coverage check, so a slot owning a partial region lowers with the full shape against a partial buffer (OOB, silent). And the inferred geometry (L92-120) is never reconciled with the configured `tp_src`/`tp_dst`. Add an extent/coverage check even on the replicate path.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
kwen2501: Addressed in `d2f8e56` / the runtime refactor: destination tiles stage into scratch first, then commit after every lane synchronizes and async-error checks pass.

modelexpress_client/python/modelexpress/refit/reshard/transport/nccl_m2n/executor.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    from math import prod
+
+    named = dict(model.named_parameters())
+    params: list[ReshardParam] = []
+    window_bytes = 0
+
+    for tt in table.tensors:
+        param = named.get(tt.name)
+        if param is None:
```
kwen2501: Addressed in `2b1223e` and follow-ups: `build_reshard_params` now raises if a trainer-table parameter is missing locally, and dtype is taken from the shared table with a local match check (`92710f2`).

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
kwen2501: Addressed in `2a167e3` / runtime lanes: staging, reshard, and commit all run on the lane CUDA stream with explicit synchronize boundaries.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
kwen2501: Addressed: `_M2nRuntime.check_async_error` runs after staging and after commit before treating tiles as complete.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
kwen2501: Lane communicators use ranks `[0, tp_src)` / `[tp_src, …)` by construction (`build_tp_meshes` + lane bootstrap). Role is derived from `lane.comm_rank`, not a global torch process group layout. Documented in the executor.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
kwen2501: Addressed earlier and superseded by the runtime refactor: window/comm lifecycle now lives in `_M2nRuntime.close()`.

modelexpress_client/python/modelexpress/refit/reshard/transport/nccl_m2n/executor.py
```diff
@@ -0,0 +1,239 @@
...
+        logger.info(
+            "reshard complete: %d params, %.2f GB in %.3fs (%.1f Gbps)",
+            len(params),
+            total_bytes / 1e9,
+            elapsed,
+            gbps,
+        )
+        return total_bytes, elapsed
+
+    def teardown(self) -> None:
```
kwen2501: Superseded by the runtime split: executor `teardown` only releases staging; `_M2nRuntime.close()` finalizes and destroys every lane communicator.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
kwen2501: Superseded: ctypes dtype mapping was removed when migrating to public `nccl.m2n.reshard`, which consumes torch dtypes directly.

modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_m2n_bind.py
kwen2501: Addressed in `20c8265`, then the ctypes binding was deleted in `e11c3c4` in favor of NCCL4Py / nccl-extensions.

modelexpress_client/python/modelexpress/weight_transfer/planner/mesh.py
kwen2501: Fixed in `92710f2`: a single shard returns `REPLICATE` only when it covers the full `[0, rows) x [0, cols)` extent; partial coverage raises.

modelexpress_client/python/modelexpress/refit/reshard/transport/nccl_m2n/mesh.py
```diff
@@ -0,0 +1,120 @@
...
+    """Infer the single sharded tensor dim from a ``TrainerTensor``'s shards.
+
+    Returns ``REPLICATE`` when a single shard covers the full tensor, ``0`` when
+    the shards partition rows (row-parallel-loaded / column-parallel weights),
+    or ``1`` when they partition columns (row-parallel weights, 2-D-col tiles).
+    Raises when the layout shards both dims at once (a later 2-D slice).
+    """
+    shards = tensor.shards
+    if not shards or len(shards) == 1:
+        return REPLICATE
```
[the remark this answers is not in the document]
kwen2501: Fixed in `92710f2` with an explicit full-coverage check before returning `REPLICATE`, plus negative-path tests.

modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_m2n_bind.py
```diff
@@ -0,0 +1,325 @@
...
+        self.nccl.ncclCommDestroy(comm)
+
+    def comm_get_async_error(self, comm: int) -> int:
+        """Block until the comm leaves ncclInProgress, then return its async state."""
+        state = ctypes.c_int(ncclInProgress)
+        while state.value == ncclInProgress:
+            rc = self.nccl.ncclCommGetAsyncError(comm, ctypes.byref(state))
+            if rc != ncclSuccess:
+                raise RuntimeError(f"ncclCommGetAsyncError rc={rc}")
+        return state.value
```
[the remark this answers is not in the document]
kwen2501: Obsolete after `e11c3c4`: the ctypes binding is gone. Finalize polling in `_M2nRuntime._wait_for_finalize` is already bounded with sleep + deadline.

modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_m2n_bind.py
```diff
@@ -0,0 +1,325 @@
...
+
+def make_tensor_desc(data_ptr, local_shape, ndims, dtype, mesh: ncclMesh_t) -> ncclDistTensor_t:
+    t = ncclDistTensor_t()
+    t.dataPtr = ctypes.c_void_p(data_ptr if data_ptr else None)
+    for i in range(NCCL_RESHARD_MAX_TENSOR_DIMS):
+        t.localShape[i] = local_shape[i] if i < len(local_shape) else 0
+    t.ndims = ndims
+    t.dtype = dtype
+    t.mesh = ctypes.pointer(mesh)
+    return t
```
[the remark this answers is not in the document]
kwen2501: Obsolete after the binding removal. `_validate_local_tiles` now rejects ranks > 3.

modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_m2n_bind.py
```diff
@@ -0,0 +1,325 @@
...
+        rc = m2n.nccl.ncclGetUniqueId(ctypes.byref(uid))
+        if rc != ncclSuccess:
+            raise RuntimeError(f"ncclGetUniqueId rc={rc}")
+
+    # Broadcast the 128 uniqueId bytes from rank 0 over torch.distributed.  Use a
+    # CPU tensor so the control-plane PG can be gloo -- keeping torch from
+    # standing up its own NCCL communicator alongside the reshard comm.
+    buf = torch.frombuffer(bytearray(uid.internal), dtype=torch.uint8).clone()
+    dist.broadcast(buf, src=0)
+    ctypes.memmove(ctypes.byref(uid), buf.numpy().tobytes(), NCCL_UNIQUE_ID_BYTES)
```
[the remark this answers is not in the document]
kwen2501: Obsolete after the binding removal. GPU e2e drivers now init an explicit gloo PG.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,269 @@
...
+        max_cta: int | None = None,
+    ) -> None:
+        self._m2n = m2n
+        self._comm = comm
+        self._rank = rank
+        self._tp_src = tp_src
+        self._tp_dst = tp_dst
+        self._device_id = device_id
+        self._stream = stream
+        self._is_src = rank < tp_src
```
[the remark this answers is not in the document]
kwen2501: Same as above: lane-local communicator ranks are `[0, nranks)` with src/dst split at `tp_src` by contract.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,269 @@
...
+
+        The window is collectively registered, so every rank must size it to the
+        same value -- callers pass a world-consistent worst case.
+        """
+        if self._window is not None and self._window_bytes >= nbytes:
+            return
+        self._release_window()
+        self._window_buf = self._m2n.mem_alloc(nbytes)
+        self._window = self._m2n.window_register(self._comm, self._window_buf, nbytes)
+        self._window_bytes = nbytes
```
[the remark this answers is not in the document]
kwen2501: Obsolete: symmetric-window management moved into nccl-extensions; MX no longer registers windows itself.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,269 @@
...
+            dst_mesh = binding.make_mesh(
+                dst_mesh_dc.dims, dst_mesh_dc.start_rank, dst_mesh_dc.placement
+            )
+
+            src_local = tile_shape(p.global_shape, src_mesh_dc)
+            dst_local = tile_shape(p.global_shape, dst_mesh_dc)
+
+            # Stage the owned tile into the window base before the collective.
+            if self._is_src:
+                self._m2n.memcpy_dtod(self._window_buf, p.local_ptr, p.local_nbytes)
```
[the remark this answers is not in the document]
kwen2501: Addressed: `_validate_local_tiles` checks local tensor shape against the mesh-derived tile before submit.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,269 @@
...
+            src_ptr = self._window_buf if self._is_src else 0
+            dst_ptr = self._window_buf if not self._is_src else 0
+            src_t = binding.make_tensor_desc(
+                src_ptr, src_local, p.ndims, p.dtype_nccl, src_mesh
+            )
+            dst_t = binding.make_tensor_desc(
+                dst_ptr, dst_local, p.ndims, p.dtype_nccl, dst_mesh
+            )
+
+            self._m2n.reshard(self._comm, self._window, src_t, dst_t, self._stream)
```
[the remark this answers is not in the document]
kwen2501: Addressed in `2a167e3` / runtime lanes with stream-ordered staging and sync.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,269 @@
...
+
+            # Copy the resharded tile out of the window into the live param.
+            if not self._is_src:
+                self._m2n.device_synchronize()
+                self._check_async_error(p.name)
+                self._m2n.memcpy_dtod(p.local_ptr, self._window_buf, p.local_nbytes)
```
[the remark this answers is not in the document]
kwen2501: Addressed in `d2f8e56`: full-model staging before commit.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_executor.py
```diff
@@ -0,0 +1,269 @@
...
+                dtype_nccl=torch_dtype_to_nccl(param.dtype),
+                local_ptr=param.data_ptr(),
+                local_nbytes=param.numel() * elem,
+            )
+        )
+
+        src_mesh, dst_mesh = build_tp_meshes(shard_dim, tp_src, tp_dst)
+        src_bytes = prod(tile_shape(global_shape, src_mesh)) * elem
+        dst_bytes = prod(tile_shape(global_shape, dst_mesh)) * elem
+        window_bytes = max(window_bytes, src_bytes, dst_bytes)
```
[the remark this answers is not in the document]
kwen2501: Fixed in `92710f2`: dtype comes from `TrainerTable` and the local parameter must match.

modelexpress_client/python/tests/gpu/run_reshard_e2e.py
```diff
@@ -0,0 +1,112 @@
...
+    return list(torch.chunk(g, tp_dst, dim=shard_dim))[dst_index]
+
+
+def main() -> int:
+    rank = int(os.environ["RANK"])
+    world = int(os.environ["WORLD_SIZE"])
+    local = int(os.environ.get("LOCAL_RANK", rank))
+    tp_src = int(os.environ.get("TP_SRC", "1"))
+    tp_dst = int(os.environ.get("TP_DST", "1"))
+    assert tp_src + tp_dst == world, f"tp_src+tp_dst ({tp_src}+{tp_dst}) != world {world}"
```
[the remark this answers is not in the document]
kwen2501: Fixed in `92710f2`: replaced `assert` with `SystemExit` checks for world size and row divisibility.

modelexpress_client/python/tests/test_nccl_m2n_reshard.py
```diff
@@ -0,0 +1,120 @@
...
+    assert torch.equal(torch.cat(dst_shards, dim=0), g)
+
+
+def test_golden_reshard_dim1_2to4():
+    torch.manual_seed(1)
+    g = torch.randn(16, 8)
+    dst_shards = _reference_reshard(g, shard_dim=1, tp_src=2, tp_dst=4)
+    _, dst_mesh = build_tp_meshes(shard_dim=1, tp_src=2, tp_dst=4)
+    assert tuple(dst_shards[0].shape) == tile_shape((16, 8), dst_mesh)
+    assert torch.equal(torch.cat(dst_shards, dim=1), g)
```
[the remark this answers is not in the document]
kwen2501: Fixed in `92710f2`: golden tests now assert each dst tile equals the matching global slice rather than an identity round-trip.

kwen2501:
Addressed the remaining open review threads in `92710f2` (on top of the earlier hardening / staging / stream / runtime commits).

**Still open from the suite-level notes** (not blocking the inline threads): wiring `tests/gpu/run_reshard_e2e.py` into GPU-gated CI so the byte-exact path is exercised automatically. Happy to follow up if that should land in this PR.

pkousha:
I pushed my fixes regardless of whether this PR lands the data plane or #659. Let this be an example of what I had in mind for NCCL M2N support

PR #498, opened 2026-07-09
fix: MTP loading

emricksini-h:
## Description

We encountered an error when deploying a native multi-token-prediction (MTP) model such as Qwen3.5 MTP or Holo3-35B with `--load-format=mx` enabled. The base model loads, then the engine fails while loading the MTP drafter:

```
(Worker_TP1_EP1) E  [metadata_stream.cpp:83] Socket Bind failed ... Address already in use [98]
(Worker_TP1_EP1) WARNING NIXL registration failed ...
...
RuntimeError: [Worker 1] No loading strategy succeeded for model '.../model_streamer/7ef1add9'
(EngineCore) ERROR EngineCore failed to start.
```

Root cause seems to be that MX assumes the loader runs once per worker (the main model). MTP breaks that assumption: vLLM invokes the loader a second time in the same worker process for the draft head. That second pass tries to open a NIXL metadata listener on a port the target already holds, so RDMA fails on bind, and the remaining strategies then fail too.

## Proposed fix

This PR proposes the following changes to fix the error:
- Keep the draft out of P2P. `_is_speculative_draft()` detects the second pass through `model_config.runner_type == "draft"` (vLLM builds the draft ModelConfig with `runner="draft"`) and sets `ctx.p2p_enabled = False`. With that flag off, the draft skips NIXL registration, metadata publication, and the RDMA strategy, so it no longer collides on the target's port and does not touch the per-device tensor/manager registries the target populated.
- Stream only the draft's shards. With RDMA disabled, the draft loads through the ModelStreamer path, which was re-reading the entire checkpoint from S3 (about 38.5 GB, 45 s) to extract a roughly 0.85 GB head. `build_model_streamer_weight_iter` now reads `model.safetensors.index.json` and streams only the shards whose tensors start with `mtp.`.

## Summary by CodeRabbit

* **New Features**
  * Added support for multi-token prediction models with a separate draft loading path.
  * Improved shard streaming by loading only the draft-related weights when available.

* **Bug Fixes**
  * Prevented connection and metadata conflicts during draft loading by disabling peer-to-peer handling for that pass.
  * Added safer fallback behavior when draft-specific shard metadata is unavailable.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/vllm/loader.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/loader.py
@@ -34,7 +34,7 @@ from ... import configure_vllm_logging
 from ...load_strategy import LoadContext, LoadStrategyChain
 from ...nixl_transfer import NixlTransferManager
 from ...vmm.runtime import log_arena_post_load, maybe_enter_vmm_arena
-from .adapter import build_vllm_load_context
+from .adapter import _is_speculative_draft, build_vllm_load_context
 from .artifacts import install_vllm_cache_artifacts, schedule_vllm_cache_artifact_publish
 
 from vllm.config import ModelConfig, VllmConfig
@@ -50,19 +50,6 @@ logger = logging.getLogger(__name__)
 # Global storage for tensor metadata, keyed by device_id (local CUDA ordinal).
 _tensor_registry: dict[int, dict[str, torch.Tensor]] = {}
 _nixl_managers: dict[int, NixlTransferManager] = {}
-
-
-def _is_speculative_draft(vllm_config: VllmConfig, model_config: ModelConfig) -> bool:
-    """True when this load is the speculative draft model, not the target.
-
-    vLLM builds the draft ModelConfig with runner="draft"; the target resolves
-    to "generate". This is more reliable than comparing against
-    speculative_config.draft_model_config, which ngram/custom_class alias to
-    the target's own config.
-    """
-    if getattr(vllm_config, "speculative_config", None) is None:
-        return False
-    return getattr(model_config, "runner_type", None) == "draft"
 
 
 class MxModelLoader(BaseModelLoader):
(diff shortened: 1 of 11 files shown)


modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -82,15 +82,17 @@ def load_model(
...
+        ctx.p2p_enabled = not _is_speculative_draft(vllm_config, model_config)
         self._ctx = ctx
 
         logger.info(
             f"[Worker {ctx.global_rank}] MxModelLoader starting "
-            f"(model={ctx.identity.model_name})"
+            f"(model={ctx.identity.model_name}, p2p_enabled={ctx.p2p_enabled})"
         )
 
         with maybe_enter_vmm_arena(ctx):
```
nv-hwoo, line 93: We should also ensure we don't run vmm arena during the second draft load so that it doesn't affect the already loaded model:
```python
def maybe_enter_vmm_arena(ctx):
    if not envs.MX_VMM_ARENA or not ctx.p2p_enabled:
        yield
        return
```

zhengluo-nv:
Thanks for your contribution. Taking a look...

nv-hwoo (changes requested):
Hi @emricksini-h thank you for the PR. I was able to reproduce the bug with Qwen3.5 2B and verified the current PR fixes the issue. One quick suggestion in the current PR is to disable VMM ARENA mode when going into the MTP model loading path since it conflicts with the already live model tensors. Other than that I think the PR looks good.

nv-hwoo:
/ok to test [392af73](https://github.com/ai-dynamo/modelexpress/pull/498/commits/392af73adcf8cd30b58d2d4717f010ecd1aa7a6f)

nv-hwoo:
/ok to test [27eb673](https://github.com/ai-dynamo/modelexpress/pull/498/commits/27eb673c6e54df89255c98322e872227bacb9051)

nv-hwoo (approved):

hzxuzhonghu:
@emricksini-h Does it support p2p for draft model loading? I mean in theory?

PR #501, opened 2026-07-13
Kimik26 humming

gaogaoSpark:
**vllm**: 0.23.0
**humming**: 0.1.10(humming_kernels-0.1.10-py3-none-any.whl)
**modelexpress**: 0.4.1 (modelexpress-0.4.1-cp312-cp312-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl)

1. **RDMA target rejected or rebuilt the wrong tensor layout.**
   Some source tensors were published as dense BF16
   `UnquantizedLinearMethod` weights, while the target dummy-load path created
   packed Humming `int32` weights for the same module. Before this fix the
   target could not safely decide whether to rebuild, keep, or reject the local
   tensor.

2. **Target forward path could still use Humming kernels after receiving dense BF16 weights.**
   Even after the target tensor shape/dtype was rebuilt to match the source,
   the target module could keep its local Humming `quant_method`. vLLM then
   interpreted a dense BF16 tensor as a packed Humming tensor and failed during
   profile/warmup.

3. **RDMA target could publish a polluted manifest after a successful load.**
   The target creates post-load derived runtime tensors such as MLA attention
   tensors (`W_UV`, `W_UK_T`) before RDMA receive. Those tensors are local
   target artifacts and should not become part of the RDMA source manifest.
   Before the final cleanup the target manifest grew from the source's
   `1554` tensors to `2091` tensors.

## Summary by CodeRabbit

- **New Features**
  - Added provider-aware model caching, including migration of legacy cache records.
  - `model clear` and `model clear-all` now remove server-side registry entries when possible.
  - Added multi-architecture Python wheel builds and pure-Python wheel support.
  - Improved RDMA transfers with richer tensor metadata, quantization support, validation, and safer retries.
  - Added readiness checks and clearer Kubernetes deployment examples.

- **Bug Fixes**
  - Stale registry entries no longer prevent missing models from being re-downloaded.
  - Improved NGC downloads across authenticated and presigned artifact sources.

- **Documentation**
  - Standardized vLLM usage on `modelexpress`, while retaining `mx` compatibility.
  - Documented `MX_SERVER_ADDRESS` and deprecated `MODEL_EXPRESS_URL`.

- **Chores**
  - Bumped the release version to 0.4.1.


zhengluo-nv:
https://github.com/vllm-project/vllm/pull/48507/changes
We have an upstream fix to vLLM. Meanwhile, I will take a look of this PR and try to fix it for existing vLLM versions

zhengluo-nv:
Fix is in https://github.com/ai-dynamo/modelexpress/pull/514

gaogaoSpark:
> Fix is in #514

Great, I’ll close this PR.

PR #491, opened 2026-07-07
fix: filter stale READY sources from ListSources

AllenXu93:
## Summary

Adds a server-side freshness filter for `ListSources(status_filter=READY)` so stale READY source records are not returned while waiting for the background reaper.

Partially addresses #484.

## Validation

Not run locally in this follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source discovery so only currently fresh “Ready” workers are returned.
  * Expired Ready entries are now filtered out immediately, reducing stale results while waiting for background cleanup.
  * Clarified architecture documentation to reflect the updated heartbeat freshness behavior during source listing.

PR #493, opened 2026-07-07
fix: try next RDMA source after transfer failure

AllenXu93:
## Summary

Changes `RdmaStrategy` so a failed RDMA source can roll back/reinitialize and try the next ranked source candidate before falling through to later load strategies. The retry budget is configurable via `MX_P2P_MAX_SOURCE_RETRIES` and defaults to 3.

Partially addresses #484.

## Validation

Not run locally in this follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a new setting to control how many RDMA source candidates are retried during P2P loading.

* **Bug Fixes**
  * RDMA loading now retries additional source candidates after transfer failures instead of stopping immediately.
  * Target state is safely reset between retry attempts when needed, improving recovery after failed transfers.

* **Documentation**
  * Updated architecture and deployment docs to reflect the new retry behavior and configuration option.

* **Tests**
  * Expanded test coverage for environment parsing, retry limits, failure recovery, and retry-related metrics.

PR #502, opened 2026-07-13
fix: handle stale P2P sources during rollout

### LINKED ISSUE #484: P2P source selection can choose stale or self sources during rollout/restart
## Problem

P2P source discovery can return sources that are not actually usable during pod rollout or worker restart.

Two failure modes have been observed:

1. During rolling update/delete+recreate, a new pod may select a READY metadata record left behind by an old pod before the reaper marks it STALE.
2. During in-place worker restart, such as OOM/CrashLoop, the new process can reuse the same pod IP and rank port while the previous process READY metadata is still present, making it look like the worker selected itself.

## Current behavior

`ListSources(status_filter=READY)` currently trusts the stored status field. It does not prove the source endpoint is still reachable or that the metadata belongs to the current source runtime generation. The reaper eventually marks old records STALE, but that cleanup is asynchronous.

## Proposed fix plan

- Add a server-side freshness filter for `ListSources(READY)` so expired READY records are not returned while waiting for the next reaper pass.
- Add source runtime identity/generation metadata, for example process UUID plus pod/container identity where available, and exclude the caller own runtime from candidate selection.
- Add a lightweight source probe or identity validation before transfer so the client can confirm the endpoint matches the selected metadata record.
- Make RDMA candidate handling continue to the next source on connection/manifest/transfer failures before falling back to storage loading.

## Notes

The freshness filter reduces the stale metadata window, but it does not fully solve freshly crashed sources or self-selection after same-pod restarts. Those require runtime identity and candidate retry.

zhengluo-nv:
## Summary

Fix P2P source selection during pod rollout and worker restart by validating the
selected runtime generation before preparing target tensors, then retrying an
alternate READY source when the selected source is stale or fails during
transfer.

Fixes #484.

## Root cause

ModelExpress discovery records identify a worker by `worker_id`, but the P2P
manifest request previously validated only `mx_source_id`. After an in-place
worker restart, the pod IP and worker ports can be reused while the previous
process's READY record is still fresh. A target could therefore send the old
selection to the new process and mistake that endpoint for the selected
generation.

Rolling updates also expose a smaller window where a READY candidate is already
terminating. The existing retry loop covered metadata misses before target
preparation, but a connection or transfer failure could immediately fall
through to disk even when another compatible source was available.

The heartbeat freshness filtering added by #491 limits how long these records
remain discoverable. It cannot distinguish two process generations that reuse
the same endpoint inside that freshness window.

## Changes

- Add optional `worker_id` fields to the tensor-manifest request and response.
  Optional presence preserves rolling-upgrade compatibility with older clients
  and sources.
- Publish the process generation into each worker gRPC server.
- Fetch and validate the selected generation's tensor manifest before target
  preparation. Reuse that prefetched manifest during the transfer.
- Reject stale selections with `FAILED_PRECONDITION` when the endpoint serves a
  different worker generation.
- Retry up to the existing three-candidate budget for manifest/connection and
  transfer failures.
- Roll back NIXL state and reinitialize the model before retrying when a failed
  transfer may have mutated target tensors.
- Bound the lightweight tensor-manifest RPC to five seconds so an unreachable
  candidate does not consume the kernel's approximately 20-second TCP connect
  timeout.
- Update selection metrics and architecture documentation for the retry path.

## Compatibility

Generation validation is negotiated by optional proto presence:

- New target + new source: validates `worker_id` in both directions.
- New target + old source: the old source ignores the unknown request field and
  returns no response generation; the target accepts the legacy response.
- Old target + new source: the request omits `worker_id`, so the new source
  preserves the previous behavior.

### Unit tests

```text
uv run --extra dev pytest \
  tests/test_artifact_transfer.py tests/test_source_selection.py -q
64 passed
```

The full local client run reached 496 passed and 22 skipped. Seven
`test_pool_registration.py` setup cases could not run on the macOS test host
because `cuda.bindings` is not installed.

Coverage includes:

- source rejects a stale `worker_id` with `FAILED_PRECONDITION`
- generation mismatch retries the next candidate before target preparation
- transfer failure retries after rollback/reinitialization
- final mutated failure still propagates mutation state to the outer strategy
- reinitialization failure stops retry and preserves mutation state
- tensor-manifest default deadline is five seconds
- optional-field compatibility with a legacy source

### Eight-replica rolling update

Using vLLM 0.23.0, Qwen2.5-0.5B-Instruct, TP=1, Kubernetes metadata, and UCX
RDMA on B200:

- seeded one disk source, then scaled to eight replicas
- all seven scale-out targets transferred 267 tensors / 0.99 GB over RDMA at
  164-173 Gbps
- rolled all eight replicas with `maxUnavailable: 0`, `maxSurge: 1`
- all eight replacements transferred over RDMA at 161-181 Gbps
- two replacements selected terminating endpoints, retried one or two later
  candidates, and completed RDMA without disk fallback
- inference succeeded on all eight replacements
- metadata converged to eight new READY and eight replaced STALE generations

### Live worker-generation mismatch

- started two compatible sources
- restarted one EngineCore in place, producing old worker `f7bab72f` and new
  worker `0ee4aa4d` on the same `10.0.7.106:6555` endpoint
- launched six targets while both records were still READY
- two targets selected `f7bab72f` and received:

  ```text
  FAILED_PRECONDITION: worker_id mismatch: expected 0ee4aa4d, got f7bab72f
  ```

- both immediately tried candidate 1 and completed the 0.99 GB RDMA transfer at
  approximately 181 Gbps
- inference succeeded on all six targets; disk fallbacks: 0
- after cleanup, metadata naturally converged to 0 READY / 9 STALE records

No metadata CRs were manually modified or deleted during these tests.

## Summary by CodeRabbit

- **New Features**
  - Improved peer-to-peer model loading with automatic retries across available sources.
  - Reinitializes the target model when a failed transfer may have changed its state.
  - Added worker-generation validation to prevent loading metadata from stale processes.
  - Reuses fetched tensor metadata to streamline transfers.
  - Supports older workers during rolling upgrades.

- **Bug Fixes**
  - Improved handling of manifest mismatches and transfer failures.
  - Added clearer fallback behavior when a source is unavailable.

- **Documentation**
  - Updated P2P loading, source selection, transfer safety, and coordination guidance.


modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -255,6 +283,32 @@ def _fetch_worker_metadata(
...
+
+        manifest_start = time.perf_counter()
+        tensor_protos, manifest_bytes = fetch_tensor_manifest(
+            endpoint=source_worker.worker_grpc_endpoint,
+            mx_source_id=mx_source_id,
+            worker_id=worker_id,
+        )
+        # Store the validated descriptors on the metadata object so the
+        # transfer can reuse them without a second manifest RPC.
+        source_worker.tensor_source.tensors.extend(tensor_protos)
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in c7feca3. The prefetched manifest now clears the existing tensor descriptors before storing the validated descriptors, so stale or duplicate entries cannot survive. The generation-mismatch regression now starts with pre-populated stale metadata and verifies that only the fetched manifest reaches the transfer; the source-selection suite passes 34/34.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -135,16 +134,41 @@ def load(self, result: LoadResult, ctx: LoadContext) -> LoadResult:
...
+                    raise
+
+                logger.warning(
+                    f"[Worker {ctx.global_rank}] RDMA source worker {worker_id} "
+                    f"failed: {e}. Trying next candidate."
+                )
+                self.rollback(ctx)
+                if e.mutated:
+                    try:
+                        result = ctx.adapter.reinit_for_retry(result)
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in c7feca3. rollback() now propagates NIXL shutdown failures and clears the manager reference only after successful cleanup. load() converts cleanup failure to StrategyFailed(mutated=True) and does not call reinit_for_retry. Added test_load_cleanup_failure_aborts_before_reinit; the source-selection suite passes 34/34.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -169,8 +195,8 @@ def load(self, result: LoadResult, ctx: LoadContext) -> LoadResult:
             f"[Worker {ctx.global_rank}] Tried {tried} of {len(candidates)} source workers "
             f"(max retries={MAX_SOURCE_RETRIES}), falling through"
         )
-        # Only pre-target metadata/discovery misses reach here. Failures after
-        # target preparation are raised from _load_as_target() as mutated=True.
+        # Reaching here means every attempted transfer either failed cleanly or
+        # was followed by a successful reinitialization, so the model is clean.
         raise StrategyFailed("No RDMA source succeeded", mutated=False)
```
yixinh-nv: **Possible bug: `mutated=False` here can discard a freshly-reinitialized model and fail the whole load.** When an earlier candidate fails after target prep, L173's `reinit_for_retry` nulls the passed-in `LoadResult` in place and returns the fresh model only into `load()`'s local `result`. If a later candidate then misses and we fall through to this raise, the chain still holds the nulled object and — seeing `mutated=False` — skips its own reinit, so the next strategy gets `model=None` → "No loading strategy succeeded". Suggest tracking whether a reinit ran and propagating that into the final `mutated=` (as #493 did with `needs_outer_reinit`) so the chain rebuilds instead.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -293,6 +319,10 @@ def _fetch_worker_metadata(
                 f"and no P2P endpoint, skipping"
             )
             return None
+        if worker.worker_grpc_endpoint:
+            # Validate the selected endpoint before target preparation. Keep
+            # the fetched descriptors for _receive_from_peer() to reuse.
+            self._prefetch_tensor_manifest(ctx, worker, mx_source_id, worker_id)
```
yixinh-nv, line 333: The prefetch runs even when `worker_id` is empty (k8s-service backend), where it can't validate anything and just re-fetches the manifest over a load-balanced channel that may hit a different pod than `get_metadata` did — mixing that pod's descriptors with the original's `agent_name`/endpoint. Since `get_metadata` already fetched the manifest there, consider guarding this on a non-empty `worker_id` so it's skipped.

zhengluo-nv: Addressed

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
zhengluo-nv: Addressed.

yixinh-nv (approved):

Issue #495, opened 2026-07-08
Regression in main: publish_metadata_and_ready drops tensors field, server stores 0 tensors per worker


OrZipori:
# Regression in main: `publish_metadata_and_ready` drops `tensors` field, server stores 0 tensors per worker

## What happened

After deploying a vLLM source worker with `--load-format modelexpress` against a Redis-backed MX server, the server was logging:

```
Published metadata for 'MyModel' (source_id=..., worker_id=...): rank 0 (0 tensors)
Published metadata for 'MyModel' (source_id=..., worker_id=...): rank 1 (0 tensors)
```

...while the worker itself was logging right before that:

```
[Worker 0] Publishing 2585 tensors for model 'MyModel'
```

All target workers then fell back to disk loading because `rdma_strategy.py` skips any worker with no tensors and no `worker_grpc_endpoint`:

```python
if not worker_tensor_descriptors(worker) and not worker.worker_grpc_endpoint:
    continue
```

## Root cause

This is a regression introduced in commit `545bdf8` ("feat: add source-typed artifact metadata payloads #411"). That commit migrated `WorkerMetadata` construction in two places from the legacy `tensors` repeated field (field 3) to the new `tensor_source` oneof field (field 20), but it just swapped one for the other instead of writing both.

**`modelexpress/metadata/publish.py`** (non-P2P branch, the `else` in `publish_metadata_and_ready`):

```python
# before 545bdf8 (v0.4.0 tag, works fine):
worker = p2p_pb2.WorkerMetadata(
    worker_rank=worker_rank,
    nixl_metadata=nixl_manager.nixl_metadata,
    tensors=tensor_protos,
)

# after 545bdf8 (current main, broken):
worker = p2p_pb2.WorkerMetadata(
    worker_rank=worker_rank,
    nixl_metadata=nixl_manager.nixl_metadata,
    tensor_source=tensor_source_metadata(tensor_protos),  # tensors= dropped
)
```

Same thing in **`modelexpress/trtllm_live_transfer.py`** (both `publish_model_params` and `publish_from_worker`).

The problem is the Redis backend log line in `redis.rs` reads directly from the legacy `tensors` field on the stored `WorkerRecord`:

```rust
info!("... rank {} ({} tensors)", worker_record.worker_rank, worker_record.tensors.len());
```

[...]

PR #504, opened 2026-07-14
fix: dropped tensors field regression

tanushriya910:
### Summary
Fixes a regression where the full tensor-manifest publish path (MX_P2P_METADATA=0) dropped the legacy tensors field from WorkerMetadata, so servers predating the tensor_source field stored 0 tensors per worker and targets silently fell back to disk loading.

### Issue
https://github.com/ai-dynamo/modelexpress/issues/495

### Root cause
Commit 545bdf8 migrated WorkerMetadata construction from the legacy tensors field to the newer tensor_source oneof, but swapped one for the other instead of writing both. A server that only reads tensors (any build predating tensor_source) sees an empty list, logs rank N (0 tensors), and targets skip RDMA.

### Fix
Dual-write both fields at all three publish sites, the same pattern the server already uses on its WorkerRecord → WorkerMetadata round-trip:
- metadata/publish.py (full tensor-manifest branch)
- trtllm_live_transfer.py (publish_model_params, publish_from_worker)

### Test coverage

- Unit test (test_full_manifest_publish_dual_writes_tensor_fields): asserts the published WorkerMetadata populates both tensors and tensor_source. Deterministic — red without the fix, green with it.
- CI, same-version: new test-p2p matrix entry runs the full source→target RDMA path with MX_P2P_METADATA=0 against this commit's server, guarding the full-manifest path end-to-end.
- CI, mixed-version: same path with the last released server (0.4.0), which predates tensor_source. This reproduces the regression (server logged 0 tensors pre-fix) and confirms the fix (267 tensors, target completes RDMA cross-version). A live backward-compatibility guard.

## Summary by CodeRabbit

* **New Features**
  * Added configurable tensor-manifest metadata publishing for distributed model workflows.
  * Added support for full tensor-manifest publishing scenarios, including mixed-version server testing.
* **Bug Fixes**
  * Improved compatibility with older servers by preserving both current and legacy tensor metadata formats.
  * Improved handling of metadata sources when server responses use older formats.
* **Tests**
  * Expanded validation for full-manifest publishing and consistent tensor information across supported modes.

### DIFF
--- a/.github/actions/run-mx-p2p-test/action.yml
+++ b/.github/actions/run-mx-p2p-test/action.yml
@@ -61,12 +61,11 @@ inputs:
   mx_p2p_metadata:
     description: |
       Value for the ${MX_P2P_METADATA} substitution in manifests that
-      template it (currently vllm/manifest-azure.yaml). "1" (default) keeps
-      the P2P-metadata path (source serves the tensor manifest on-demand via
-      a WorkerGrpcServer). "0" exercises the non-P2P embedded-metadata path
-      (tensors carried in the WorkerMetadata proto) — the path where the
-      MX-396 dropped-tensors regression surfaces. Applied to both source and
-      target manifest applies.
+      template it (currently vllm/manifest-azure.yaml). "1" (default) serves
+      the tensor manifest on-demand via a WorkerGrpcServer. "0" exercises the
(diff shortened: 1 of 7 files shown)


zhengluo-nv (approved):

tanushriya910:
Closes #495

PR #505, opened 2026-07-14
fix(vllm): support mixed Humming tensor layouts over RDMA

zhengluo-nv:
## Summary

- carry tensor layout and quantization metadata through RDMA manifests
- capture Humming's packed runtime weights and rebuild compatible target tensors before registration
- align mixed unquantized layers on the target, then refresh derived attention tensors after RDMA receive
- filter target-only post-load tensors so a successful target does not publish a polluted source manifest

## Why

With Kimi K2.6 using Humming, some source modules publish dense BF16 weights while the target dummy-load path creates packed Humming `int32` weights for the same module. The old RDMA path only compared names, sizes, and dtypes after target setup, so it could not safely rebuild the target layout or align the module's quantization method. Target-side post-load tensors could also leak into the next published manifest.

This is a clean replacement for #501, rebased onto current `main` and limited to its unique Humming/RDMA fix. The original fix author is preserved on the commit.

## Validation

- `pytest -q tests/test_vllm_adapter.py tests/test_tensor_utils.py tests/test_transfer_safety.py tests/test_vllm_loader.py` (144 passed, 8 skipped)
- focused non-CUDA pool-registration tests (CUDA-driver-only cases require `cuda.bindings`)
- Ruff on all changed Python files
- `cargo check --workspace --all-targets`


modelexpress_client/python/modelexpress/engines/vllm/adapter.py
```diff
@@ -92,14 +112,44 @@ def before_rdma_receive(self, result: LoadResult) -> LoadResult:
         result = self._finalize_model_specific_weights(result)
         return self._process_weights_after_loading(result)
 
+    def after_rdma_receive(self, result: LoadResult) -> LoadResult:
+        return self._refresh_attention_weights_after_rdma(result)
```
yafshar: This runs on every vLLM RDMA receive, not just Humming. That means paths
that already worked (e.g. DeepSeek MLA, FP8) will re-run
`process_weights_after_loading()` on attention weights refreshed by
`_refresh_vllm_attention_runtime_tensors()`.

Is `process_weights_after_loading()` guaranteed to be idempotent for
`Attention`, `MLAAttention`, and `MMEncoderAttention` across supported
vLLM versions? If not, should this be gated to Humming (or cases where
derived tensors are excluded from the manifest)?

zhengluo-nv: This PR is copied from https://github.com/ai-dynamo/modelexpress/pull/501 and it's WIP. We should not merge this

PR #507, opened 2026-07-14
feat: set ownerReference on ModelMetadata CRs for automatic GC on pod deletion

omerap12:
## Overview

This PR associates Kubernetes-backed `ModelMetadata` records with the Pod that
published them. When a client provides a complete same-namespace Pod identity,
the Kubernetes metadata backend adds a Pod `ownerReference`, allowing Kubernetes
to garbage-collect the metadata immediately when that Pod is deleted.

The change applies to every engine and metadata type that uses the shared
`PublishMetadata` path, including weight and artifact metadata.

### What changed

- Added optional `pod_name`, `pod_uid`, and `pod_namespace` fields to
  `PublishMetadataRequest`.
- Updated the Python client to populate those fields from `POD_NAME`, `POD_UID`,
  and `POD_NAMESPACE`.
- Updated the Kubernetes backend to:
  - create Pod-owned `ModelMetadata` CRs for complete same-namespace identities;
  - patch existing CRs when Pod identity becomes available on a later publish;
  - reject partial or cross-namespace identities for ownership without rejecting
    metadata publication.
- Preserved the existing `ModelMetadata` -> tensor `ConfigMap` ownership chain,
  so deleting a Pod collects both the CR and its tensor metadata.
- Added Downward API wiring to the checked-in vLLM, SGLang, Dynamo, and CI
  manifests.
- Updated the deployment, metadata, and Kubernetes P2P example documentation.

### Compatibility

The protobuf fields are additive. MX 0.4.0 and other older clients omit them and
continue publishing metadata without a Pod owner reference. Missing, partial,
or cross-namespace identity also preserves this fallback behavior. Redis,
in-memory, and non-Kubernetes environments ignore the Pod fields and continue
using the existing heartbeat/reaper lifecycle.

### Local and CI-equivalent checks

- Rust tests: 184 passed
- Workspace Clippy: passed
- Python client, vLLM, and SGLang tests: 100 passed, 8 skipped
- Go binding tests: passed
- YAML validation: passed
- Post-`origin/main` merge focused client/SGLang tests: 44 passed

### Kubernetes E2E

Tested in namespace `zheng` on a B200 node with the Kubernetes metadata backend,
an isolated server built from this PR, the shared model PVC, and
`Qwen/Qwen2.5-0.5B`. Source and target ran on the same node; this validates the
complete Kubernetes lifecycle and NIXL/UCX P2P path, but is not a cross-node RDMA
benchmark.

| Engine | Runtime | Source behavior | Target P2P result | Inference |
|---|---|---|---|---|
| vLLM | 0.23.0 | Cold-loaded from disk and published Ready metadata | 267 tensors / 0.99 GB in 0.062s (128 Gbps) | `/v1/models` succeeded; completion returned `Paris` |
| SGLang | 0.5.13.post1 | Cold-loaded from disk and published Ready metadata | 170 tensors / 0.99 GB in 0.321s (24.6 Gbps) | `/v1/models` succeeded; completion returned `Paris` |

For both engines:

- source and target published separate Ready `ModelMetadata` CRs;
- each CR referenced the correct publishing Pod name and UID;
- deleting the client Deployments reduced the matching CR count to zero through
  Kubernetes owner garbage collection;
- no `ModelMetadata` CRs were manually deleted.

The live E2E run exercised weight metadata. Artifact metadata uses the same
`PublishMetadata` ownership path, but artifact transfer was not run as a separate
E2E case in this validation.


modelexpress_server/src/p2p/backend/kubernetes.rs
```diff
@@ -220,6 +223,20 @@ impl MetadataBackend for KubernetesBackend {
...
+                                kind: "Pod".to_string(),
+                                name: pod_name.to_string(),
+                                uid: pod_uid.to_string(),
+                                controller: Some(false),
+                                block_owner_deletion: Some(false),
+                            },
+                        ])
+                    } else {
+                        None
+                    },
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed. `publish_metadata` now computes the Pod owner reference on every publish, fetches the current CR, and merge-patches `metadata.ownerReferences` whenever a complete same-namespace Pod identity is available and differs from the stored value. This covers CRs initially created without ownership as well as create races, while missing, partial, or cross-namespace identity still preserves the no-owner fallback for backward compatibility.

zhengluo-nv:
/ok to test [df84eee](https://github.com/ai-dynamo/modelexpress/pull/507/commits/df84eee4a411e237a72a3d91b481c94d02d932ec)

zhengluo-nv (approved):

zhengluo-nv (commented):
Do not merge, waiting for final review

modelexpress_client/python/modelexpress/envs.py
```diff
@@ -219,6 +221,8 @@ def _env_float(name: str, default: float) -> float:
     "VLLM_ATTENTION_BACKEND": lambda: os.environ.get("VLLM_ATTENTION_BACKEND", "auto"),
     "HOSTNAME": lambda: os.environ.get("HOSTNAME", ""),
     "POD_NAMESPACE": lambda: os.environ.get("POD_NAMESPACE", ""),
+    "POD_NAME": lambda: os.environ.get("POD_NAME", ""),
+    "POD_UID": lambda: os.environ.get("POD_UID", "")
```
omerap12, line 225: nit:
```suggestion
    "POD_UID": lambda: os.environ.get("POD_UID", ""),
```

omerap12 (commented):
/lgtm

@zhengluo-nv Thanks for helping get this across the finish line! 🙂

modelexpress_server/src/p2p/backend/kubernetes.rs
```diff
@@ -202,6 +229,8 @@ impl MetadataBackend for KubernetesBackend {
 
         // First, ensure the CR exists
         let existing = api.get_opt(&cr_name).await?;
+        let pod_owner_references =
+            Self::pod_owner_references(pod_name, pod_uid, pod_namespace, &self.namespace);
```
omerap12, line 233: Maybe worth revisiting this architecture in the future? Currently, when the pod and ModelMetadata CR are in different namespaces, we lose automatic GC and fall back to the reaper. I could be wrong, but I imagine many customers will have pods outside the ModelExpress namespace. Would it make sense to create the ModelMetadata CR (and the corresponding ConfigMap) in the same namespace as the pod? We'd need broader RBAC, but since the ModelMetadata CRD belongs to ModelExpress, that seems reasonable.

zhengluo-nv: This PR preserves the existing single metadata-namespace contract and intentionally falls back to the reaper for cross-namespace Pods. Will create a new task to support worker pods and modelexpress are in different namespaces.

modelexpress_client/python/modelexpress/envs.py
zhengluo-nv: Will fix in the new PR

PR #510, opened 2026-07-15
feat: bandwidth-aware load_aware P2P source selection

yixinh-nv:
## Summary

Adds a third P2P source-selection policy, **`load_aware`**, that biases weight-pull
source selection away from sources whose RDMA path is already busy, so scaling up
or rolling a live inference fleet does not steal bandwidth from sources that are
actively serving. It builds on the `random` / `rendezvous_hash` selector interface
from #461: same `ScoredSelector` seam, one new scalar on the wire, no new RPCs, and
it collapses to `rendezvous_hash` whenever the load signal is absent — never worse
than the deterministic baseline.

The MX servers stay **stateless**: the load signal is *published by the source about
itself* and passed through by the server — the server never counts or accumulates
anything — so replicas can sit behind a load balancer and scale arbitrarily, and any
two replicas rank an identical `ListSources` identically.

### `source_load`: a source-published load scalar (proto + server passthrough)
- New `float source_load` on `SourceInstanceRef` (=6) and `UpdateStatusRequest` (=5)
  in `p2p.proto` (Python + Go bindings regenerated).
- The server plumbs it as an opaque passthrough: `WorkerRecord.source_load`, refreshed
  on `update_status` and surfaced on `list_sources`; memory / Redis (`#[serde(default)]`)
  / Kubernetes-CRD (`sourceLoad`) backends all carry it. The server never derives it.

### `load_aware` selector (client)
- `score = unit_hash(target, candidate) − w_load · source_load`, `source_load` clamped
  to `[0,1]`, `w_load = MX_P2P_LOAD_WEIGHT` (default 1.0, clamped `≥ 0`).
- Absent/zero load ⇒ every penalty is 0 ⇒ **exact `rendezvous_hash` ordering** (safe
  default; also covers old servers that predate the field).

### Where the load number comes from (two providers, blended)
`source_load` is produced *on the source* by `make_source_load_provider(device_id)`,
reporting `max()` of two independent signals so selection reacts to a NIC that is
physically hot **and** to an imminent serving spike a counter has not seen yet:

1. **RDMA NIC utilization** (`nic_metrics.py`, runtime-agnostic, default). Reads the
   GPU-affine RDMA NIC's InfiniBand port byte counters
   (`/sys/class/infiniband/<dev>/ports/<p>/counters/{port_xmit_data,port_rcv_data}`,
   ×4 for the 4-octet spec unit) and divides the delta since the last sample by the
   elapsed time and the link capacity (parsed from `.../rate`). The NIC is the exact
   rail transfers pin to — resolved via the same `ucx_utils.probe_nic_pin_for_device`
   PCIe-affinity assignment the NIXL transfer path uses — so it reflects the contention
   a puller would actually hit. Degrades safely to *unknown* — the field is left unset (no
   device / unreadable sysfs / unparseable rate) — so IB/RoCE report a real number and other
   fabrics get the selector's neutral prior, never a spurious "idle". **Note:** on SR-IOV VF pods (the common `rdma/ib` device-plugin setup) the container's VF sysfs omits the `counters/` stats (they live on the host PF). The sampler then falls back to the same statistics over netlink (`RDMA_NLDEV_CMD_STAT_GET`), which works unprivileged, but mlx5's default counter set carries no byte totals — those are optional counters an admin must enable with `CAP_NET_ADMIN` — so on VF pods this provider reports unknown and provider #2 is the effective signal. Sampled off the existing publisher heartbeat (no extra sleeps).
2. **Inference-runtime serving load** (`runtime_load.py`, opt-in via
   `MX_P2P_RUNTIME_METRICS_URL`). Scrapes the co-located vLLM/SGLang Prometheus
   `/metrics` for a KV-cache-utilization gauge (`vllm:kv_cache_usage_perc`, or the
   legacy `vllm:gpu_cache_usage_perc`; `sglang:token_usage`, or Dynamo's runtime-agnostic re-export `dynamo_component_gpu_cache_usage_percent` as a fallback for engines behind a Dynamo runtime, e.g. TRT-LLM) as a `[0,1]` serving-load
   proxy — *predictive*, since a full KV cache foretells RDMA the NIC counter has not
   measured yet, and it works on fabrics whose sysfs counters cannot be read.

Neither provider touches the server, proto, or selector; the only wire contract is the
normalized `source_load` scalar, and any provider error leaves it unset (unknown), which the
selector treats as a neutral prior rather than as an idle source.

### Observability
- `mx_p2p_source_load{scheme, source_worker_id}` gauge recorded per candidate at
  selection (opt-in `MX_METRICS_ENABLED=1`), alongside the existing
  `mx_p2p_source_selections_total` / `mx_p2p_transfer_seconds`.

## Compatibility
- Additive proto field; old clients/servers ignore it and `load_aware` degrades to
  `rendezvous_hash`. Default policy remains `random`. No new RPCs, no publish-cadence
  change, no server-side state.

### Unit tests
- Python: `load_aware` collapse-to-rendezvous at zero load, steering away from the
  busiest source, `w_load` monotonicity, negative-weight clamp, `[0,1]` clamping,
  missing-field tolerance; NIC sampler (injected reader/clock/link-rate); runtime
  provider (vLLM new/legacy gauge names, SGLang, unreachable ⇒ unknown); publisher forwards a
  provider value and leaves `source_load` unset on error. Rust: `Option<f32>` passthrough
  across memory/Redis/k8s backends with non-zero round-trip assertions.

### NIC sampler calibration against `ib_write_bw` (ConnectX-7, 400 Gb/s NDR)

The NIC provider reads the port's hardware counters, so its accuracy was checked
against an independent, software-side measurement: `ib_write_bw` between two nscale
B200 nodes (`mlx5_0`, 400 Gb/s NDR InfiniBand, PF counters via `hostNetwork` pods),
software rate-limited to a fixed 100 Gb/s for 100 s with 1 MiB messages, while
`NicUtilizationSampler` sampled the same port every 30 s. Idle noise floor beforehand
was 0.00 MB/s on every 400G rail of both nodes.

| | Gb/s | fraction of link |
|---|---|---|
| requested (SW rate limit) | 100.00 | 0.2500 |
| `ib_write_bw` achieved (payload, software-side) | 100.03 | 0.2501 |
| sampler published, fully-covered 30 s windows | 101.02 | **0.2525** |

They agree within 1% (+0.99%); the difference is wire header and ack overhead that
payload-side accounting does not count. Partial windows matched the client's wall-clock
overlap to within 1 Gb/s, the t=0 sample was unset (baseline), and receive traffic was
~0.02 Gb/s (acks). Because the two measurements are independent, their agreement
validates the counter source, the 4-octet word scaling, the window arithmetic, the
`rate` parse, and device selection together. Verified on ConnectX-7 InfiniBand only;
EFA reports unknown by design.

### B200 E2E
Ran on B200, InfiniBand RDMA with a vLLM **disaggregated**
prefill/decode fleet, TP=1, kubernetes metadata backend, `MX_METRICS_ENABLED=1`, and
`source_load = max(NIC util, vLLM KV-cache usage)`. Fleet driven by AIPerf (closed-loop
concurrency 48, 512-in / 128-out, streaming chat) against the dynamo frontend, capturing
TTFT / request-latency / MX metrics at **baseline**, during a **Case-1 scale-up**
(decode 4→8 under load, new replicas pull weights over MX P2P), and during a **Case-2
rolling update** (decode replicas replaced one at a time under load).

- **Mechanism plumbing, end to end.** New replicas load weights via `--load-format
  modelexpress` and register as P2P sources; the NIC/KV `source_load` provider, the
  selector, and the `mx_p2p_source_load` telemetry all run on real hardware.
- **This validation surfaced a Kubernetes-backend bug, now fixed in this PR.** The
  deployed `ModelMetadata` CRD is a structural schema with no `sourceLoad` field, so the
  Kubernetes metadata backend **pruned `source_load` on write** -- every candidate read
  back `0`, silently collapsing `load_aware` to `rendezvous_hash` on that backend. This
  PR adds `sourceLoad` to the CRD (`crd-modelmetadata.yaml`, `examples/crds.yaml`) so the
  value survives the CR round trip. (The Rust serde round-trip tests passed because they
  exercise struct (de)serialization, not the live API server's schema pruning.) **Validated live:** on a B200 re-run with the fixed CRD, busy sources published a non-zero `source_load` (~0.3-0.6) that propagated through the CR to `ListSources`, and `random` routed ~17-33% of scale-up pulls onto them -- confirming the fix restores the signal on the k8s backend (pre-fix every candidate read `0`).
- **No regression.** Because `source_load` was inert on the k8s backend (above) -- and is
  in any case uniform under Dynamo round-robin serving -- `load_aware` rode its
  `rendezvous_hash` fallback and matched the deterministic baseline exactly. On
  **Qwen2.5-14B** a scale-up under load held TTFT flat (p99 ~= 380 -> 375 ms, throughput
  up ~5%). This confirms the designed safety property (never worse than baseline); it
  does **not** demonstrate load-based steering, which is shown by the simulation below.

### `load_aware` vs `random`: head-to-head
The same B200 fleet was run end to end back-to-back under **`load_aware`** and
**`random`** (Qwen2.5-32B, TP=1, disaggregated prefill/decode, AIPerf closed-loop
concurrency 64), comparing baseline serving, a Case-1 decode scale-up (4→8 under load),
and a Case-2 rolling update.

**Live (disaggregated): no regression, parity.** Serving is identical within
run-to-run noise:

| metric (baseline) | `load_aware` | `random` |
|---|---|---|
| TTFT avg / p99 (ms) | 334 / 843 | 362 / 935 |
| request latency avg (ms) | 2234 | 2236 |
| token throughput (tok/s) | 16 817 | 16 828 |

During the scale-up **every candidate's `source_load` read back `0.0` at selection**,
for two reasons: the CRD pruning above (fixed in this PR), and -- structurally --
disaggregated decode pullers pull from same-role *decode* sources that are KV-idle, while
the busy 99.9%-KV *prefill* workers are a different role and are filtered out of the
candidate set. With a uniform-zero signal `load_aware` collapses to `rendezvous_hash` by
construction, so it spreads and completes exactly like `random` -- confirming the safety
property (never worse than the baseline), not a win.

**Where the sources are actually busy — simulation over the real selector.**
A committed discrete-event harness drives the
**real `load_aware` / `rendezvous_hash` / `random` selector code** over a fan-out where a
few sources are persistently busy (serving + streaming KV over RDMA) and the rest idle:

| M sources / busy / N pullers | policy | pulls onto a busy source | fan-out makespan | busy-source disruption |
|---|---|---|---|---|
| 8 / 1 / 40 | **`load_aware`** | **0.0%** | **8.0** | **0.0** |
| 8 / 1 / 40 | `random` | 12.5% | 11.2 | 9.0 |
| 8 / 2 / 40 | **`load_aware`** | **0.0%** | **9.3** | **0.0** |
| 8 / 2 / 40 | `random` | 25.8% | 12.7 | 18.6 |

> `rendezvous_hash` is intentionally absent from this table. The script that
> produced it was scratch-only and is no longer recoverable, and a reconstruction
> reproduced the busy-pull column but not the makespan/disruption scale, so its
> numbers are not comparable with the two columns above. The live 400-trial check
> below covers `rendezvous_hash` on the same question with the real selector.

`load_aware` routes **0%** of pulls onto a serving source vs `random`'s **12–26%**,
cutting fan-out makespan **~27–28%** and eliminating the bandwidth stolen from
actively-serving sources (the busy-source disruption proxy), while staying at parity
when load is uniform. The harness is an offline, GPU-independent driver of the real
selector code; the steering direction it measures is guarded in CI by the
`test_load_aware_steers_away_from_busy_source` unit test.

### Real-server selection check (live k8s backend, default weight)
Complementing the offline simulation, source workers were registered on a live MX
server (kubernetes metadata backend) with a controlled `source_load` set via
`update_status`, then the **real `load_aware` / `random` selectors** were driven over
400 puller trials at the **default weight** (`MX_P2P_LOAD_WEIGHT=1.0`). The `source_load`
survived the full `update_status -> ModelMetadata CR -> ListSources` round trip
(re-confirming the CRD fix above), and `load_aware` routed **~0% of pulls onto busy
sources vs `random`'s ~25%** (2 busy + 6 idle sources), steering harder as the busy
source's load rises:

| busy `source_load` | `load_aware` busy-hit | `rendezvous_hash` busy-hit | `random` busy-hit |
|---|---|---|---|
| 0.3 | 2.0% | 27.2% | 22.0% |
| 0.6 | **0.0%** | 25.5% | 31.8% |
| 0.9 | **0.0%** | 23.0% | 25.5% |

The `rendezvous_hash` column comes from a re-run of this same harness that also
re-measured `load_aware` and `random` and reproduced their published values within
tolerance (2.8/0.0/0.0% and 27.5/27.2/25.5%), so the three columns are comparable.
It sits at the ~25% base rate throughout, i.e. it is load-blind exactly as designed:
the avoidance in the `load_aware` column comes from the load signal, not merely from
replacing random shuffling with deterministic hashing.

This exercises the real server, CRD backend, and selector end to end (selection +
plumbing) at the default weight, without relying on serving-KV saturation.

### Transfer-time under RDMA contention

The selection checks above show `load_aware` *steers away* from a loaded source; this
measures the **payoff in wall-clock weight-transfer time**, using the real
`vllm serve --load-format modelexpress` path on nscale B200 (Qwen2.5-32B, TP=1). One
source served an idle baseline, then a **12-way concurrent fan-out**, 
so the source's single NIC -- the shared
bottleneck of any fan-out -- is genuinely contended by MX's own P2P traffic:

| source NIC | per-pull weight transfer (`mx_p2p_transfer_seconds`) | achieved BW |
|---|---|---|
| idle | **2.3 s** | 227 Gbps |
| contended (12-way fan-out) | **22.8 s** | 22 Gbps |

Pulling from the contended source is **~9.9× slower** (all 12 contended samples kept
after a bandwidth-drop filter that discards any pull that did not actually share the
NIC). The same stack was then used to measure routing under that contention, with
**one contended and one idle source** (so a topology-blind policy sits at the ~50%
base rate, not the ~25% of the 2-busy/6-idle table above): `load_aware` sent
**4.8%** of pulls to the hot source vs `random`'s **45%**. Composing that routing
with the measured penalty gives the routing-weighted expected transfer time:

| policy | E[weight-transfer time] |
|---|---|
| `random` | 11.6 s |
| `load_aware` | **3.3 s** |

### Live Mooncake production-trace benchmark (B200) — the load signal, end to end

The runs above set `source_load` explicitly to isolate the selector. This one lets it
arise **organically from real production traffic**: the Mooncake arXiv trace is replayed
(original request shapes) against **2 of an 8-source fleet**, every source publishes its
own vLLM KV-cache utilization as `source_load` via `MX_P2P_RUNTIME_METRICS_URL`, and
targets then pull the 15.2 GB model over RDMA under each policy (5 waves x 4 targets =
20 pulls per policy, fresh fleet per policy).

The signal separated cleanly every round -- **0.65-0.91 on the trace-loaded sources vs
0.0 on every idle one** -- so this exercises the whole chain: real inference load ->
KV gauge -> published `source_load` -> `ListSources` -> selector.

Which sources the pulls actually landed on (loaded = `source_load` >= 0.05 at wave time):

| policy | pulls onto a LOADED source |
|---|---|
| `load_aware` | **0/18 (0%)** |
| `rendezvous_hash` | 5/19 (26.3%) |
| `random` | 3/19 (15.8%) |

With 2 loaded sources out of 8, a load-blind policy sits at a ~25% base rate, which is
where both baselines land; `load_aware` avoided them entirely. (An earlier 3-pull-per-policy
attempt could not distinguish the policies at all -- every policy scored 0/3, which is
unremarkable at p=0.42 -- hence 20 pulls per policy here.)

Transfer time over the same waves:

| comparison | worst transfer | mean transfer |
|---|---|---|
| `load_aware` vs `random` | **54.1% lower** (0.98 s vs 2.13 s) | **33.2% lower** |
| `load_aware` vs `rendezvous_hash` | -4.1% (parity) | -4.9% (parity) |

Parity with `rendezvous_hash` is the expected result, not a shortfall: it already spreads
deterministically, so with only 2 of 8 sources loaded there is little tail left to
reclaim (its worst pull was 0.94 s). The win shows up against `random`, which piled up
to 5-deep on one source and produced a 2.13 s / 57 Gbps straggler. In other words the
policy behaves as designed under live production load -- it avoids busy sources outright,
beats the naive baseline on the tail, and never regresses against the deterministic one.

### How the advantage scales with the loaded fraction

Every measurement above uses a fleet with 1-2 loaded sources. This sweeps the loaded
fraction on the same live harness (400 trials, `MX_P2P_LOAD_WEIGHT=1.0`, busy
`source_load` 0.9, `source_load` set via `update_status`) and adds one larger fan-out:

| fleet (busy / idle) | loaded fraction | `load_aware` busy-hit | `rendezvous_hash` busy-hit | `random` busy-hit |
|---|---|---|---|---|
| 2 / 6 | 25% | **0.0%** | 25.0% | 25.5% |
| 4 / 4 | 50% | **0.0%** | 47.0% | 47.5% |
| 6 / 2 | 75% | **0.0%** | 80.0% | 76.0% |
| 4 / 12 (larger pool) | 25% | **0.0%** | 23.8% | 25.2% |

The intuition that the advantage should shrink as the loaded fraction rises does **not**
hold at the selection level, and it is worth stating plainly: `load_aware` keeps the
busy-hit rate at zero at every fraction tested, because it only has to find *some* idle
source, not a large pool of them. Meanwhile the load-blind policies track the base rate,
so the gap actually *widens* with load (0% vs 25% -> 0% vs ~80%).

Where the headroom argument does bite is transfer time rather than selection: at 75%
loaded, every pull is funnelled onto the 2 remaining idle sources, so the avoided
contention is partly traded for self-contention on those sources. That is the same
effect visible in the Mooncake run below, where an 8-source fleet with only 2 loaded
left `load_aware` at parity with `rendezvous_hash` on transfer time while still winning
decisively on which sources were chosen. Selection avoidance and transfer-time gain are
therefore not the same curve, and only the former is fleet-size-independent.


docs/ARCHITECTURE.md
```diff
@@ -614,12 +614,15 @@ Strategies handle the loading path and NIXL tensor registration. `LoadContext.ac
...
+
+The selector controls ordering only; `RdmaStrategy` enforces the fixed three-candidate retry budget. Selectors read the few fields they need (`worker_rank`, `worker_id`, `identity.model_name`, and `source_load` for `load_aware`) directly off the live `LoadContext` and candidate refs, so there is no parallel context type. Topology-aware policies remain deferred (they still need node/rack topology signals); the same scoring interface accepts them as drop-ins. Selection decisions are emitted as structured logs, and an opt-in Prometheus collector (`metrics.py`, `MX_METRICS_ENABLED=1`) re-emits them as `mx_p2p_*` metrics (including `mx_p2p_source_load`) for benchmarking different schemes.
```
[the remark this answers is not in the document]
yixinh-nv: Fixed in 34c41cc: the `update_status` signature row now includes `source_load`, and the L631 wording no longer calls `load_aware` deferred (it ships in this PR). The remaining 'deferred' at L625 refers to *topology-aware* policies, which genuinely are still deferred.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

modelexpress_client/python/modelexpress/envs.py
```diff
@@ -195,6 +199,8 @@ def _env_float(name: str, default: float) -> float:
     # ── P2P source selection ───────────────────────────────────────────────
     # Raw (None when unset); source_selection applies its DEFAULT_SELECTOR fallback.
     "MX_P2P_SOURCE_SELECTOR": lambda: os.environ.get("MX_P2P_SOURCE_SELECTOR"),
+    "MX_P2P_LOAD_WEIGHT": lambda: _env_float("MX_P2P_LOAD_WEIGHT", 1.0),
+    "MX_P2P_RUNTIME_METRICS_URL": lambda: os.environ.get("MX_P2P_RUNTIME_METRICS_URL"),
```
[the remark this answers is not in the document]
yixinh-nv: Fixed in 34c41cc: `MX_P2P_LOAD_WEIGHT` is now clamped to `>= 0` in envs, with a test (`test_load_aware_negative_weight_clamped_to_zero`) asserting a negative value collapses to rendezvous ordering rather than rewarding load.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

modelexpress_client/python/modelexpress/runtime_load.py
```diff
@@ -0,0 +1,90 @@
...
+
+logger = logging.getLogger("modelexpress.runtime_load")
+
+# Metrics already expressed as a [0, 1] utilization ratio, in priority order.
+# vLLM: fraction of KV-cache blocks in use. SGLang: token/KV usage fraction.
+_RATIO_METRICS = (
+    "vllm:gpu_cache_usage_perc",
+    "sglang:token_usage",
+    "sglang:cache_usage",
+)
```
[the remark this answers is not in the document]
yixinh-nv: Fixed in 65f5731: added `vllm:kv_cache_usage_perc` first in `_RATIO_METRICS`, kept `vllm:gpu_cache_usage_perc` as legacy fallback, plus a test (`test_provider_reads_vllm_017_kv_cache_usage`).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

modelexpress_client/python/tests/test_runtime_load.py
```diff
@@ -0,0 +1,79 @@
...
+def test_make_source_load_provider_blends_when_url_set(monkeypatch):
+    # With the runtime URL set, the provider is the max of NIC and runtime; both
+    # degrade to 0 locally (no IB device, unreachable URL), so it must not crash
+    # and must return a valid [0,1] float.
+    from modelexpress.nic_metrics import make_source_load_provider
+
+    monkeypatch.setenv("MX_P2P_RUNTIME_METRICS_URL", "http://127.0.0.1:1/metrics")
+    provider = make_source_load_provider(0)
+    val = provider()
+    assert isinstance(val, float) and 0.0 <= val <= 1.0
```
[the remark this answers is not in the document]
yixinh-nv: Fixed in 34c41cc: the blend test now patches the `_http_get` fetch seam to return None, so it does no real socket I/O.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

yixinh-nv:
Re: the Codecov patch report — the `codecov/patch` check is passing (96.25%, no repo threshold, informational). The three uncovered lines are trivial `source_load` passthrough and aren't cheaply unit-testable:

- `memory.rs`: the `map_or_else` default branch for a worker entry with no ranks — a defensive/unreachable state not producible through the backend API.
- `kubernetes.rs` / `redis.rs`: `source_load` read/set on the live k8s-API / redis paths, which need a real API client; the existing serde round-trip tests already cover (de)serialization, matching the pre-existing coverage pattern for those backends.

No behavioral risk. Happy to add fake-client tests if you'd like 100% patch coverage here.

modelexpress_client/python/modelexpress/nic_metrics.py
```diff
@@ -0,0 +1,226 @@
...
+def _parse_rate_bytes_per_sec(rate_str: str) -> Optional[float]:
+    """Parse an IB ``rate`` file (e.g. ``"400 Gb/sec (4X HDR)"``) to bytes/sec."""
+    m = re.search(r"([0-9.]+)\s*Gb/sec", rate_str)
+    if not m:
+        return None
+    gbps = float(m.group(1))
+    return gbps * 1e9 / 8.0  # Gb/s -> bytes/s (full-duplex, per direction)
+
+
+class NicUtilizationSampler:
```
zhengluo-nv, line 258: Can vllm/sglang provide metrics that we can use to estimate the load?

yixinh-nv: Yes — `runtime_load.py` already scrapes `vllm:kv_cache_usage_perc` / `sglang:token_usage` / `dynamo_component_gpu_cache_usage_percent` behind `MX_P2P_RUNTIME_METRICS_URL` and blends it with the NIC signal as a max, so I only fixed the stale docstring here that described it as future work.

modelexpress_common/proto/p2p.proto
```diff
@@ -637,6 +650,11 @@ message UpdateStatusRequest {
 
   // worker_id returned from PublishMetadata
   string worker_id = 4;
+
+  // Source-published busyness in [0, 1], refreshed on each heartbeat so the
+  // server's SourceInstanceRef.source_load tracks live load. 0 means
```
nicolasnoble: The comment here says 0 means idle/unknown. Plain proto3 float has no presence bit, and the selector computes `unit_hash - w*load`, so 0.0 scores highest. `nic_metrics.py` documents the provider reporting 0.0 on VF pods under the common rdma/ib device-plugin setup, and older clients send the proto3 default, so I think in a mixed deployment sources without a load reading end up ahead of the ones that have it. `optional` or a sentinel would separate the two cases.

modelexpress_server/src/p2p/service.rs
```diff
@@ -396,7 +397,13 @@ impl P2pService for P2pServiceImpl {
 
         match self
             .state
-            .update_worker_status(&req.mx_source_id, &req.worker_id, req.worker_rank, status)
+            .update_worker_status(
+                &req.mx_source_id,
+                &req.worker_id,
+                req.worker_rank,
+                status,
+                req.source_load,
```
nicolasnoble: `status` on the line above is a validated local, `req.source_load` goes to the backend as received. This is a public gRPC surface and the value is publisher-asserted, so a NaN looks reachable from a non-Python client. `serde_json` writes a non-finite float as null and `serde(default)` doesn't rescue an explicit null on read, so I think one bad heartbeat is enough to make `status:{rank}` unreadable: `overlay_status` falls back to the record heartbeats never rewrite, `updated_at` stays at registration time, and the reaper collects a worker that is still heartbeating. The shipped python client clamps, so it isn't reachable today.

modelexpress_client/python/modelexpress/metrics.py
```diff
@@ -281,6 +282,15 @@ def _build_families(self) -> None:
...
             registry=registry,
         )
+        # Same label, same cardinality argument as selections above: the peer id
+        # is per-process, so it rides the same opt-in switch rather than being
+        # the one mx_p2p_* family that grows unbounded by default.
+        self.source_load = _make_gauge(
+            registry,
+            "mx_p2p_source_load",
+            "Source-published load in [0,1] observed on candidates at selection.",
+            ["scheme", "source_worker_id"] if self._source_id_label else ["scheme"],
```
nicolasnoble, line 365: Minor. With `MX_METRICS_SOURCE_ID_LABEL` off, `scheme` is the only label, so I think the per-candidate loop overwrites one series and what remains is the last element of `ordered`, which under `load_aware` is the busiest candidate.

helm/crds/modelexpress-crds.yaml
```diff
@@ -88,6 +88,10 @@ spec:
                     accelerator:
                       type: string
                       description: Runtime accelerator family for compatibility filtering
+                    sourceLoad:
```
nicolasnoble, line 91: Minor. The new status fields go into all three CRD copies without `x-kubernetes-preserve-unknown-fields`, and helm doesn't upgrade files under `crds/`, so a cluster upgraded without a manual apply should have the field dropped by the api server.

nicolasnoble (commented):
This conflicts with #512, and one can't be merged after the other.

modelexpress_common/proto/p2p.proto
yixinh-nv: Agreed and fixed in 31b4eb9f: `source_load` is now `optional` at both proto sites (same field numbers, wire-compatible), carried as `Option<f32>` through every backend and the CRD, providers return `None` instead of 0.0 when they have no reading, the publisher leaves the field unset, and the selector scores unset as a neutral prior (`UNKNOWN_LOAD_PRIOR = 0.5`) so a source with no reading neither outranks nor trails one that reports real load — happy to move that prior if you'd rather it sit elsewhere.

modelexpress_server/src/p2p/service.rs
yixinh-nv: Right — the handler now sanitizes `source_load` alongside the existing `status` validation: non-finite becomes `None` (no reading) rather than reaching the backend, and finite values are clamped to [0,1]; I chose degrade-to-unknown over rejecting the RPC because a rejected heartbeat would trip the client's new re-registration path over a telemetry glitch (31b4eb9f).

modelexpress_client/python/modelexpress/metrics.py
yixinh-nv: Confirmed and fixed (31b4eb9f): with the label off, `observe_candidate_loads` now records only the chosen candidate (`ordered[0]`) instead of looping over all of them, skips candidates with no reading rather than writing 0.0, and a test pins that the single series holds the chosen 0.0, not the busiest 0.9.

helm/crds/modelexpress-crds.yaml
yixinh-nv: True, and main's #696 already documents the reapply procedure for existing clusters, so rather than duplicate it the `MX_P2P_SOURCE_SELECTOR` row now says `load_aware` on the kubernetes backend needs the CRD from this release, that Helm does not update `crds/` on upgrade, and links to that procedure — otherwise the API server prunes the field and the policy degrades silently to rendezvous (31b4eb9f).

modelexpress_server/src/p2p/k8s_types.rs
```diff
@@ -127,6 +127,14 @@ pub struct WorkerStatus {
     #[serde(default)]
     pub accelerator: String,
 
+    /// Source-published busyness in [0, 1] (default: RDMA NIC utilization).
+    #[serde(
+        default,
+        rename = "sourceLoad",
+        skip_serializing_if = "Option::is_none"
```
nicolasnoble: `skip_serializing_if` here drops a None from the patch body, and both status writes go out as `Patch::Merge` at lines 362 and 763, where an omitted key means leave unchanged. So I think this field can go None to Some but not back: once a real reading is persisted, nothing on this backend clears it.

`reaper.rs:75` makes that reachable. It passes None when marking a rank Stale, so a worker that heartbeated a load and then went quiet keeps advertising its last reading on the CRD after it is marked Stale. Redis does not have this, since the Lua `HSET` rewrites the whole field.

Dropping skip_serializing_if so None serializes as null would need `nullable: true` alongside it. All three CRD copies have `sourceLoad` as `type: number` with no nullable, so the api server would reject the patch as it stands.

yixinh-nv: Confirmed on all three links — dropped `skip_serializing_if` so `None` reaches the merge patch as an explicit `null` (delete), added `nullable: true` to `sourceLoad` in all three CRD copies so the API server accepts it, and pinned it with a serde test for `None` → `null` out and `null` → `None` back (506f0034).

nicolasnoble (approved):

PR #513, opened 2026-07-16
fix(server): touch downloaded cache hits

andrewpaprotsky:
## Summary

- update `last_used_at` when `EnsureModelDownloaded` returns an existing downloaded model
- add a regression test proving a valid cache hit calls `touch_model`

## Root cause

The distributed registry refactor in `b0c94ed` removed the status pre-check that previously touched cache entries. The replacement `AlreadyExists(DOWNLOADED)` path returned success without an equivalent touch, leaving LRU timestamps at their original download time even when clients reused the model.

## Impact

Cache hits now follow the documented LRU semantics without changing the download, retry, or stale-record recovery paths. Touch failures remain non-fatal, matching the existing `get_status` behavior.

## Validation

- `cargo test -p modelexpress-server --lib` (183 passed)
- `cargo clippy -p modelexpress-server --lib --tests -- -D warnings`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model caching behavior by refreshing a downloaded model’s recent-use timestamp when it is accessed.
  * Helps keep frequently used models available in the cache longer.
  * Added validation to ensure cached model access updates usage tracking correctly.

### DIFF
--- a/modelexpress_server/src/services.rs
+++ b/modelexpress_server/src/services.rs
@@ -744,6 +744,12 @@ impl ModelDownloadTracker {
                         self.delete_status(model_name).await;
                         continue;
                     }
+                    if existing == ModelStatus::DOWNLOADED {
+                        // Returning an existing downloaded model is a cache hit for LRU purposes.
+                        if let Err(e) = self.registry.touch_model(model_name).await {
+                            error!("Failed to touch cached model {model_name}: {e}");
+                        }
+                    }
                     break (existing, false);
                 }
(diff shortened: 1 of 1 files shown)


fujin (approved):

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/513/commits/e9607c6d4a27364b7bbd5b9e2895ea89005be5e5

AndyDai-nv (approved):

PR #516, opened 2026-07-16
feat: extend MX bootstrap to support NCCL bootstrap

pkousha:
## Summary

This draft adds the minimal ModelExpress control-plane support needed to bootstrap an NCCL M2N communicator without Torch/Gloo rendezvous.

- Extend the existing `WeightSyncService`; no new service or proto.
- Add typed `PublishNcclBootstrap` and `GetNcclBootstrap` RPCs.
- Store one complete bootstrap record by a never-reused `attempt_id`.
- Keep `cohort_id` as the stable logical communication-group identity, including across future membership reconfiguration.
- Use the generated `WeightSyncServiceStub` directly from the NCCL bootstrap helper.
- Initialize with `ncclCommInitRankConfig(blocking = 0)`, poll `ncclCommGetAsyncError` to completion or deadline, and best-effort abort locally on failure.
- Remove the previous dedicated service, Redis/CAS backend, abort tombstones, bootstrap client/types package, session/provider abstractions, and file-based coordinator.

The external NeMo RL/Dynamo coordinator remains responsible for membership, roles, NCCL rank assignment, result collection, release, and whole-attempt termination. MX only brokers the NCCL bootstrap record.

See [`design.md`](design.md) and the [bootstrap flow diagram](docs/images/nccl-m2n-bootstrap-flow.svg).

## Flow

1. The coordinator assigns `cohort_id`, a new UUIDv4 `attempt_id`, the ordered roster, roles, NCCL ranks, source/destination sizes, and `roster_digest`.
2. Assigned NCCL rank zero generates one `ncclUniqueId` and calls `PublishNcclBootstrap`.
3. Every participant, including rank zero, polls `GetNcclBootstrap(attempt_id)` with its remaining local deadline.
4. Each participant validates cohort, attempt, sizes, digest, UID, and local rank before NCCL initialization.
5. Each participant calls `ncclCommInitRankConfig` with `blocking = 0` and polls `ncclCommGetAsyncError`.
6. After fetching the record, clients stop reading MX. The coordinator collects local results and releases all ranks or terminates the attempt.

## API and storage

The bootstrap record contains:

- `cohort_id`
- `attempt_id`
- 128-byte `nccl_unique_id`
- source, destination, and total world sizes
- 32-byte `roster_digest`

`WeightSyncState` stores `attempt_id -> NcclBootstrapRecord` in its existing process-local `RwLock`. The server validates canonical UUIDv4 identity, nonempty cohort, UID/digest lengths, positive source/destination sizes, checked size addition, and total-world consistency.

There is deliberately no CAS, conflict detection, fingerprint, TTL, deletion, expiry, abort state, revision, or publisher field. Correctness relies on the trusted coordinator assigning one publisher, never reusing an `attempt_id`, and using a new attempt for every retry or reconfiguration. A publish timeout fails the attempt; the client does not regenerate or automatically retry under the same ID.

## Cohorts and future elastic membership

`cohort_id` identifies a stable logical group, such as one pipeline-parallel source/destination group. `attempt_id` identifies one concrete bootstrap or future reconfiguration operation over a roster snapshot.

This v1 treats membership as fixed for one communicator because current NCCL communicator APIs do. That is not an MX restriction. If NCCL adds supported elastic join or reconfiguration APIs, the coordinator can keep the same `cohort_id`, issue a new `attempt_id`, and publish the updated roster digest, ranks, sizes, and NCCL-provided payload through typed reconfiguration messages. CAS is still unnecessary while the coordinator serializes operations; generation fencing becomes necessary only if MX must arbitrate competing reconfigurations.

## Deployment and failure semantics

Every publish/get for an attempt must reach the same `WeightSyncServiceImpl` state object. The supported v1 deployment therefore uses one MX server instance. Independent MX replicas have independent maps; ordinary load balancing can route the publisher and readers to different state and cause reader timeouts.

- Restart before all ranks fetch loses the record and fails the attempt.
- After all ranks fetch, NCCL initialization can continue because clients no longer read MX.
- Records accumulate until restart because v1 has no cleanup.
- Multi-replica/HA support requires backend-neutral shared ephemeral storage.
- `ncclCommAbort` is synchronous and can hang; a process supervisor is required for a hard whole-attempt termination deadline.

## Verification

The Python bootstrap suite uses focused fakes plus the real generated `WeightSyncServiceStub` over an in-process gRPC server.

The four-rank control-plane test starts three nonzero ranks polling before publication, then verifies a two-source/two-destination world where:

- NCCL rank zero generates and publishes exactly one 128-byte UID;
- all four ranks fetch the same record and initialize world size four with ranks 0–3;
- every rank transitions through simulated nonblocking `ncclInProgress` to `ncclSuccess`; and
- no rank aborts on the successful path.

Additional coverage includes publication rejection and RPC failure, missing/malformed records, metadata and rank mismatches before NCCL init, remaining-deadline propagation, immediate/asynchronous NCCL failure, null communicator, timeout/abort, abort-failure preservation, and verification that MX is not read after NCCL initialization begins.

Executed locally:

```text
uv run --no-sync pytest -q tests/test_m2n_bootstrap.py
30 passed

uv run --no-sync pytest -q \
  tests/test_m2n_bootstrap.py \
  tests/test_rl_weight_sync.py \
  tests/test_m2n.py \
  tests/test_e2e_nccl_m2n.py
89 passed
```

Rust unit tests were added for server validation, publish/get round trip, unknown attempts, cohort/attempt isolation, same-cohort new attempts, and identical republish. They were not executed in this shell because `cargo` is unavailable.

The four-rank test simulates NCCL calls; it verifies the MX/gRPC bootstrap behavior, not a real NCCL communicator. Native-shim compilation in this shell is blocked by an incomplete CUDA development-header bundle (`nv/target` is missing).

## Dependency and remaining integration work

PR #497 has not landed. This draft intentionally excludes its eight data-plane files so this remains a thin bootstrap PR. After #497 lands:

- rebase this PR onto #497;
- update #497's existing `_nccl_m2n_bind.M2N` class with device selection, UID-byte generation, nonblocking init, async-error polling, and local abort wrappers;
- replace all native return-code `assert` statements and handle ignored cleanup return codes explicitly;
- replace `bootstrap_comm_from_torch` in the GPU harness with the generated WeightSync stub and coordinator assignment; and
- rerun real two-source/two-destination GPU/M2N validation.

The selected NCCL M2N revision must support nonblocking communicators, including asynchronous completion of its internal `ncclDevCommCreate`. Multi-parameter staging-window completion ordering and a canonical collective manifest remain #497/data-plane prerequisites rather than responsibilities of this bootstrap control-plane change.

## Summary by CodeRabbit

* **New Features**
  * Added NCCL M2N communicator bootstrapping through WeightSyncService.
  * Added support for publishing and retrieving attempt-specific NCCL bootstrap records.
  * Added nonblocking communicator initialization, asynchronous status polling, timeout handling, and failure recovery.
  * Added validation for bootstrap identity, world sizes, roster information, and NCCL unique IDs.
  * Added optional Linux NCCL support for Python installations.

* **Tests**
  * Added coverage for successful bootstrap, validation failures, RPC errors, timeouts, and communicator abort handling.


modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_m2n_bind.py
```diff
@@ -0,0 +1,334 @@
...
+            ctypes.POINTER(ncclDistTensor_t),  # dst
+            ctypes.c_void_p,  # cudaStream_t
+        ]
+        self.m2n.ncclReshardWithWindow.restype = ctypes.c_int
+
+    # -- thin wrappers --
+    def init(self, max_cta: int | None = None):
+        cfg = make_config(max_cta)
+        rc = self.m2n.ncclM2nInit(ctypes.byref(cfg))
+        assert rc == ncclSuccess, f"ncclM2nInit rc={rc}"
```
nicolasnoble: These four are still `assert rc == ...` - `ncclM2nInit` here, plus `ncclMemAlloc` at :194, `cudaMemcpy` at :255, `cudaDeviceSynchronize` at :259 - while the rest of the file moved to explicit `if rc != ...: raise`. Under `python -O` the asserts get stripped, so an optimized build silently accepts a failed alloc or copy and reshards garbage into the live weights with the rc unreadable. Worth converting these to match the others.

modelexpress_client/python/modelexpress/weight_transfer/transport/nccl_m2n_bootstrap.py
```diff
@@ -0,0 +1,247 @@
...
+        raise
+
+
+def _abort_with_note(m2n: _M2nBinding, comm: int, error: BaseException) -> None:
+    """Abort best-effort without replacing the original bootstrap failure."""
+
+    if not comm:
+        return
+    try:
+        m2n.comm_abort(comm)
```
nicolasnoble: `comm_abort` here has no timeout, and this is the cleanup path for a failed nonblocking communicator - exactly the case `ncclCommAbort` can hang on. If it wedges, the abort blocks in error handling and swallows the original failure instead of surfacing it. The gRPC abort side already caps at `min(5.0, timeout)`; worth bounding this one too, or at least noting the process-supervisor dependency at the call site.

modelexpress_client/python/tests/test_m2n_test_coordinator.py
```diff
@@ -0,0 +1,67 @@
...
+    (attempt_directory / "ready-0").write_text("rank-0", encoding="utf-8")
+    time.sleep(0.1)
+    assert process.poll() is None
+    assert not (attempt_directory / "decision").exists()
+
+    (attempt_directory / "ready-1").write_text("rank-1", encoding="utf-8")
+    stdout, stderr = process.communicate(timeout=5)
+    assert process.returncode == 0, (stdout, stderr)
+    decision = attempt_directory / "decision"
+    assert decision.read_text(encoding="utf-8") == "release"
```
nicolasnoble: This exercises only the release path. The abort branches in `gpu/m2n_test_coordinator.py` - failure marker at :73-83, timeout at :97-105, both writing an `abort:` decision and calling `_abort_mx` - are never hit, so the abort-wins side of the release/abort contention (including the `except FileExistsError` reconciliation) is untested. Worth a case that drops a `failed-*` marker and asserts the abort decision.

modelexpress_client/python/modelexpress/weight_transfer/transport/_nccl_bootstrap_ext.cpp
kwen2501, line 1: Should we use nccl4py instead of dlopen in this file?

kwen2501:
> Before this PR, is there any rendezvous infra in ModelExpress?

Yes — there are **two** existing services in ModelExpress that could serve as rendezvous for `ncclUniqueId` without any changes:

---

### 1. `WeightSyncService` — `PublishTrainerTable` / `GetTrainerTable` (cleanest fit)

This is literally a generic key-value store over `bytes`:

- `PublishTrainerTable(model_key: string, table_payload: bytes)` — rank 0 calls this with the 128-byte UID
- `GetTrainerTable(model_key: string) → (found: bool, table_payload: bytes)` — other ranks poll until `found=true`

No schema validation on `table_payload`. Server-side it's just `HashMap.insert` / `HashMap.get`. Zero modification needed. The only downside is it's in-memory only (not Redis), but bootstrap blobs are ephemeral anyway.

Defined in `modelexpress_common/proto/weight_sync.proto`, implemented in `modelexpress_server/src/weight_sync/service.rs`.

---

### 2. `P2P Service` — `PublishMetadata` / `GetMetadata` (also works, more ceremony)

The existing NIXL bootstrap already uses this exact pattern. `WorkerMetadata.nixl_metadata` is a raw `bytes` field — rank 0 today publishes its NIXL agent info (rkeys, endpoint) as an opaque blob; other ranks call `ListSources` + `GetMetadata` to retrieve it. An `ncclUniqueId` is also 128 bytes and would slot in identically, with Redis persistence as a bonus.

---

### What `WeightSyncService` also has: a true all-ranks barrier

`RegisterM2nWorker` / `GetM2nPlan` (added in PR #497) accumulates registrations from all N workers before firing — only when `registrations.len() == total_workers` does it unblock `GetM2nPlan`. That's a real collective rendezvous that already exists, though it's currently coupled to RDMA routing plan output, not blob distribution.

---

### Bottom line

PR #516 adds a third rendezvous service (`M2nBootstrapService`) with new fencing concepts (`attempt_id`, abort tombstones, `roster_digest`) that don't exist in the two prior services. For the bare goal of "rank 0 publishes `ncclUniqueId`, others fetch it," `PublishTrainerTable`/`GetTrainerTable` would work today with zero server-side changes. The new service is justified primarily by the error-fencing semantics (safe retry via `publication_fingerprint`, abort propagation), not by lack of a key-value primitive.

pkousha:
@nicolasnoble  yes the CAS can be dropped because the attempt_id is never reused. Any restart/rebuild will issue a new attempt_id The goal of the new bootstrap was to capture more resiliency. As a result of removing CAS, failure/abort ownership should move entirely to the coordinator which I assume safe as the coordinator (nemo RL) is responsible for initiating communicator setup. Failure/abort ownership should move entirely to the coordinator.

pkousha:
I was unaware of WeightSyncService. Given the coordinator guarantees and dropping CAS/fencing, I can simplify the design as follows:
```
Coordinator defines trainer + generator membership,
NCCL ranks, roles, world size, and attempt_id
        ↓
Source NCCL rank 0 generates ncclUniqueId
        ↓
Source rank 0 publishes bootstrap record through WeightSyncService,
keyed exclusively by attempt_id
        ↓
All participants fetch and validate the record
        ↓
Each participant calls ncclCommInitRankConfig
        ↓
Coordinator collects initialization results
        ↓
Coordinator releases everyone or terminates the attempt
```
I would add typed RPCs to WeightSyncService, rather than overload the TrainerTable payload:
```
PublishNcclBootstrap(...)
GetNcclBootstrap(...)
```
The bootstrap record would contain the UID, attempt ID, world size, and enough coordinator-provided topology metadata for client-side validation.
This solution has the following caveats -- mostly resiliency:

1. Existing RegisterM2nWorker coordinates only inference-side NIXL planning. It neither registers trainer ranks nor assigns NCCL ranks, and NCCL M2N does not require its RDMA plan.
2. The coordinator must define membership, roles, NCCL ranks, world size, and a never-reused attempt_id. Without CAS, correctness depends on one publisher and a new attempt for every retry. I think this is OK overall with our vision just need to be communicated. 
3. Current WeightSync storage is a process-local HashMap: it is lost on restart, unsafe across independent MX replicas, and has no TTL or bootstrap-record deletion.
4. WeightSync has no abort or all-rank initialization barrier. The coordinator must collect results, release successful ranks, and terminate all participants on failure; clients still need deadlines.
5. Production HA would require a backend-neutral ephemeral KV implementation; the existing P2P metadata backend is not currently a generic KV interface.

The nonblocking NCCL initialization, ncclCommGetAsyncError polling, deadline handling, and coordinator release barrier would remain unchanged.

kwen2501:
Screenshot 2026-07-23 at 2 58 39 PM

If these two protocols look too specific, another possibility is to create a new .proto file that defines NCCL's format -- we just need to define `service` (publish, get) and `message` (i.e. the unique id). Shouldn't be too hard following the examples.

kwen2501:
> (In WeightSyncService)
> ```
> PublishNcclBootstrap(...)
> GetNcclBootstrap(...)
> ```

That works too! Thanks @pkousha

PR #517, opened 2026-07-16
Feat: Add delta weight sync RFC

athreesh:
## Summary

- Adds `docs/DELTA_WEIGHT_SYNC_RFC.md` to frame delta-compressed weight updates as a first-class ModelExpress source type for async RL workflows.
- Calls out requirements for checkpoint anchors, delta lineage, pre-pause fetch/process/stage, pause-bound apply/resume, target demand, fallback, correctness checks, and observability.
- Adds the no-hard-wrapped-Markdown guidance to `CLAUDE.md` and carries forward the scoped `AGENTS.md` user-facing guide content on the new branch.

## Why

Async RL rollout workers need a way to stay close to trainer policy updates without repeatedly moving full checkpoints or stretching the serving pause window.

The RFC keeps the first review at the MX product boundary: what MX should own for versioning, readiness, tensor identity, target demand, fallback, and observability before API or implementation design.

## Validation

- Docs-only change.
- `git diff --check`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,181 @@
+---
+name: modelexpress-user-guide
+description: Help users understand, deploy, operate, and debug public ModelExpress. Use for questions about MX concepts, quick starts, Kubernetes, Helm, metadata backends, engine integrations, ModelStreamer, P2P transfer, and troubleshooting.
+license: Apache-2.0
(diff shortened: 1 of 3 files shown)


athreesh:
Closing this draft PR because the delta weight sync RFC should be tracked as an issue instead: https://github.com/ai-dynamo/modelexpress/issues/524

PR #523, opened 2026-07-20
feat(client): support heterogeneous NIXL weight sources (CUDA/XPU)

yafshar:
## What

Makes accelerator compatibility source-type aware so base model weights can
transfer over NIXL/RDMA between CUDA and XPU workers, while keeping generated
artifacts restricted to the same accelerator family and blocking quantized
weights cross-vendor.

- Permit `cuda -> xpu` and `xpu -> cuda` source selection for
  `MX_SOURCE_TYPE_WEIGHTS`. The receive path validates tensor name, size, and
  dtype before issuing transfers.
- Restrict cross-vendor weight transfer to explicitly unquantized identities.
  A plain identity requires `quantization` empty/`none` AND a `dtype` on an
  allowlist of plain dtypes (`float32`/`float16`/`bfloat16` and aliases). This
  is an allowlist, not a denylist: `int8`, `int4`, `qint8`, `auto`, and any
  unknown future quantized dtype fail closed even when `quantization` is empty,
  so a hardware-specific layout is never silently admitted cross-vendor.
- Quantized layouts (fp8 scale packing, fp4/nvfp4 swizzle, hidden quant-config
  tensors) are kernel- and hardware-specific; copying the post-processed bytes
  cross-vendor can make the transfer succeed while the receiver's kernel
  misreads the layout, silently corrupting inference. Same-family quantized
  transfers stay allowed, since source and target run identical post-processing.
- Require an exact source/target tensor-name match for cross-family transfers.
  A local-only/source-only name diff or a zero-match transfer fails closed with
  `ManifestMismatchError` instead of transferring a subset, which would leave
  vendor-specific hidden or derived tensors at dummy values while RDMA reports
  success. Same-family transfers keep tolerating subset transfers.
- Fail closed on undeclared quantization: a caller that does not pass both
  `quantization` and `dtype` never gets a cross-vendor weight transfer.
- Defer an empty (unknown) accelerator ref on the pre-fetch check so the
  k8s-service backend (which publishes a synthetic ref and only learns the real
  accelerator from `GetTensorManifest`) is not stranded; the authoritative
  post-fetch check fails closed on a still-unknown accelerator for a quantized
  identity.
- Keep generated-artifact sources, including CUDA graph, `torch.compile`,
  Triton, DeepGEMM, TileLang, CuTe, and FlashInfer caches, restricted to the
  same accelerator family because they are architecture-specific.
- Fail closed when the source type is unspecified: cross-vendor selection is
  only enabled when the caller explicitly passes a hetero-eligible source type.
- Preserve existing behavior for exact same-family matches and empty/legacy
  accelerator values.
- Leave LoRA cross-vendor transfer disabled until there is a live LoRA transfer
  path to validate.

No server-side change is required. The Rust server continues to pass through the
runtime accelerator string and applies no compatibility logic.

## Why

Enables RL rollout and disaggregated inference deployments where weight sources
and targets run on different accelerator vendors, for example NVIDIA trainers
serving Intel XPU rollout workers.

Quantized weights are excluded because ModelExpress transfers post-processed
in-memory tensor bytes, not raw checkpoint files. For quantized models the
in-memory layout depends on the accelerator family, GPU architecture, selected
kernel, and framework quantization backend, so the transfer API and CRC can pass
while inference is silently wrong. In a mixed CUDA/XPU fleet serving quantized
weights, a target that finds only cross-family sources skips P2P RDMA and falls
through to the next load strategy (GDS or disk).

## Changes

- `metadata/payload.py`: source-type-aware
  `accelerators_compatible(..., *, mx_source_type=None, quantization=_UNSPECIFIED,
  dtype=_UNSPECIFIED, defer_unknown=True)`. Gates cross-vendor weights on an
  explicitly unquantized identity via `_is_unquantized`, which now checks the
  dtype against an `_UNQUANTIZED_DTYPES` allowlist (with `_normalize_dtype`
  stripping a leading `torch.` so `str(tensor.dtype)` and the identity form
  normalize alike). Fails closed on undeclared quantization/dtype, and defers
  vs rejects unknown accelerators by the pre-/post-fetch phase.
- `nixl_transfer.py`: add `require_exact_match` to
  `NixlTransferManager.receive_from_source()`. When set, a tensor-name mismatch
  (local-only or source-only) or a zero-match transfer raises
  `ManifestMismatchError` rather than warning and transferring a subset.
- `load_strategy/rdma_strategy.py`: pass `ctx.identity.mx_source_type`,
  `quantization`, and `dtype` into source pre-filtering and post-`GetMetadata`
  compatibility checks; the post-fetch check uses `defer_unknown=False`. Set
  `require_exact_match=True` only when the target and source accelerator
  families differ.
- `metadata/artifact_transfer.py`: pass artifact source type explicitly so
  artifact discovery remains same-family only.
- Tests: policy-matrix coverage (including the dtype allowlist and aliases)
  plus source-selection, artifact-discovery, receive-path manifest-validation,
  cross-family exact-name-match (partial and zero-match), loader
  accelerator-skip, quantized pre-fetch/post-fetch, and k8s-service
  synthetic-ref coverage.
- `docs/ARCHITECTURE.md`: document the CUDA/XPU weight-transfer and quantization
  policy.
- `docs/DEPLOYMENT.md`: add the Cross-Vendor (CUDA/XPU) Compatibility section,
  including the dtype allowlist and the cross-family exact-name-match rule.

## Testing

- Python client test suite:
  - NVIDIA node
  - XPU node
- Physical two-node smoke validation:
  - NVIDIA L40S <-> Intel Arc Pro B60
  - RoCEv2 via `rc_mlx5/mlx5_0:1`
  - UCX 1.22.0 + NIXL 1.3.0
  - torch 2.13.0+cu130 on NVIDIA
  - torch 2.13.0+xpu on XPU
  - central metadata mode: `MX_P2P_METADATA=0`
- Byte-verified transfers:
  - `cuda -> xpu`
  - `xpu -> cuda`
  - `cuda -> cuda` same-family sanity
  - `xpu -> xpu` same-family sanity
- Mixed-dtype smoke manifest verified end to end: `float32`, `float16`,
  `bfloat16`, `int32`, `int8`, `uint8`. Note: the smoke exercises the raw
  `receive_from_source` path directly. `int32`/`int8`/`uint8` move correctly as
  bytes, but the accelerator-compatibility gate deliberately rejects them for
  cross-vendor weight selection (not on the plain-dtype allowlist); only the
  float dtypes are admitted through the policy.

## Out of Scope / Follow-Ups

- Full-model refit through the vLLM/SGLang adapter beyond the smoke manifest.
- Hardware validation of specific quantized CUDA/XPU quantization/layout pairs;
  a proven pair would be enabled by an allowlist extension beyond the
  family-level `HETEROGENEOUS_NIXL_WEIGHT_PAIRS`, not by converting weights
  during transfer.
- Pre-existing: `test_pool_registration.py` CUDA-driver fixtures error rather
  than skip on non-CUDA hosts; unrelated to this change.
- LoRA cross-vendor transfer.

## Summary by CodeRabbit

* **New Features**
  * Added accelerator compatibility rules for cross-family weight transfers between supported CUDA and XPU combinations.
  * Artifact and RDMA source selection now applies source-type-aware accelerator matching.
  * Preserved backward compatibility for missing accelerator metadata while rejecting unsupported combinations.

* **Bug Fixes**
  * Prevented incompatible artifact sources from being selected.
  * Ensured tensor size mismatches are detected regardless of differing device identifiers.

* **Documentation**
  * Documented accelerator compatibility and fail-closed behavior for unsupported source types.


zhengluo-nv:
/ok to test [8029e75](https://github.com/ai-dynamo/modelexpress/pull/523/commits/8029e75690219eb1e790845a61ab7903cacf9bfa)

zhengluo-nv:
/ok to test [cc6f3f2](https://github.com/ai-dynamo/modelexpress/pull/523/commits/cc6f3f21cb30788d8613db315130cae929cc17a3)

modelexpress_client/python/modelexpress/metadata/payload.py
```diff
@@ -29,14 +29,147 @@ def worker_tensor_count(worker: p2p_pb2.WorkerMetadata) -> int:
...
+
     if not target or not source:
         return True
-    return target == source
+    if target == source:
+        return True
+    if mx_source_type in HETEROGENEOUS_NIXL_SOURCE_TYPES:
+        if frozenset((target, source)) not in HETEROGENEOUS_NIXL_WEIGHT_PAIRS:
+            return False
+        return _is_unquantized(quantization, dtype)
```
zhengluo-nv, line 191: Could we require exact source/target tensor catalogs before allowing a heterogeneous transfer here? `receive_from_source()` currently validates size and dtype only for names present on both sides; local-only/source-only names merely produce a warning, and even zero matches return successfully. The caller does not reject that result.

Once this path admits CUDA↔XPU sources, vendor-specific hidden or derived tensors can differ, leaving parts—or all—of the target at dummy values while RDMA is reported as successful. For cross-family transfers, please fail on any tensor-name mismatch and add partial- and zero-match regression tests.

modelexpress_client/python/modelexpress/metadata/payload.py
```diff
@@ -29,14 +29,147 @@ def worker_tensor_count(worker: p2p_pb2.WorkerMetadata) -> int:
...
+    cross-family quantized transfer.
     """
+    if quantization is _UNSPECIFIED or dtype is _UNSPECIFIED:
+        return False
+    if str(quantization).strip().lower() not in _UNQUANTIZED_QUANTIZATION_VALUES:
+        return False
+    dtype_norm = str(dtype).strip().lower()
+    if not dtype_norm:
+        return False
+    return not any(token in dtype_norm for token in _QUANTIZED_DTYPE_TOKENS)
```
zhengluo-nv: This does not appear fully fail-closed. Because it uses a denylist, `quantization=""` with `int8`, `int4`, `qint8`, `auto`, or an unknown future quantized dtype is considered unquantized and enables CUDA↔XPU transfer.

Could we instead allowlist the normalized plain identity dtypes we explicitly support—such as `bfloat16`, `float16`, and `float32`, plus intentional aliases—and reject everything else? That would match the documented “explicitly unquantized” contract and avoid silently admitting a hardware-specific layout.

modelexpress_client/python/modelexpress/metadata/payload.py
yafshar: Fixed in 645b223. Cross-family transfers now require an exact tensor-name match. Added `require_exact_match` to `NixlTransferManager.receive_from_source():` on a heterogeneous transfer, any local-only/source-only name diff or a zero-match transfer raises `ManifestMismatchError` instead of warning-and-transferring-a-subset. `RdmaStrategy` sets `require_exact_match=True` only when the target and source accelerator families differ, so same-family transfers keep tolerating subset transfers as before. Regression tests added in `test_pool_registration.py:` `test_hetero_name_mismatch_raises` (local-only), `test_hetero_source_only_name_mismatch_raises` (source-only), `test_hetero_zero_match_raises` (zero match), and `test_hetero_matched_name_passes_guard` (identical name sets clear the guard and proceed to the size check).

modelexpress_client/python/modelexpress/metadata/payload.py
yafshar: Fixed in 645b223. Replaced the `_QUANTIZED_DTYPE_TOKENS` denylist with an `_UNQUANTIZED_DTYPES` allowlist (`float32/float16/bfloat16` plus aliases: `float`, `f32`, `half`, `f16`, `bf16`). `_is_unquantized()` now requires the `dtype` to normalize onto that allowlist, so `int8`, `int4`, `qint8`, `auto`, and any unknown future quantized `dtype` fail closed even with `quantization=""`. Added `_normalize_dtype()` to strip a leading `torch`. and lowercase, so `str(tensor.dtype)` and the identity form match. Tests in `test_payload.py:` `test_weights_cross_family_non_allowlisted_dtype_rejected` (`int8/int4/qint8/uint4/auto/e4m3/future`) and `test_weights_cross_family_plain_dtype_aliases_allowed` (aliases + torch.-prefixed).

zhengluo-nv:
/ok to test [67d5bc5](https://github.com/ai-dynamo/modelexpress/pull/523/commits/67d5bc59b02f70eda46205db604b74d40d1b911c)

zhengluo-nv (approved):

PR #528, opened 2026-07-22
perf(reshard-refit): bound descriptors for strided TP pulls

KavinKrishnan:
## Summary

- replace descriptor-heavy exact strided copies with one full-source staging pull when a copy exceeds `MX_RESHARD_MAX_SEGMENTS_PER_COPY` (default 64)
- reconstruct only complete, gap-free dim-0 source partitions, then replay loader-captured views locally into existing receive buffers
- preserve the exact known-correct descriptor plan when a source layout cannot use bounded staging
- report exact/actual descriptor counts, descriptors saved, extra wire bytes, full-pull sources, and unbounded sources

## Why

Row-parallel TP placement can produce one NIXL descriptor per row. At model scale, descriptor preparation and dispatch dominate even when the requested bytes are small. This policy makes the tradeoff explicit: bounded descriptors in exchange for extra wire bytes and persistent staging memory.

## Measured Topology A evidence

Real NeMo-RL/Megatron Qwen3-30B-A3B BF16, 16 trainer GPUs -> 16 vLLM GPUs (4 x TP4), AWS EFA:

- exact descriptors avoided per rank: **12,576,768**
- descriptors remaining per rank: **117,123**
- useful bytes/rank: **14.58 GB**
- extra wire bytes/rank: **15.20 GB** (51% redundant)
- `wire_exact` warm median: **0.937 s**
- `wire_full` warm median: **0.668 s**
- total wire warm median: **1.684 s**
- local reslice warm median: **0.072 s**
- `fallback=0`; 48 incompatible sources retained exact descriptors

This row is `PERF_PASS_CORRECTNESS_PENDING`: it validates the descriptor/byte/time tradeoff, not final parameter-equality or FP8 A/B correctness.

## Safety and memory tradeoff

- bounded staging is used only for complete dim-0 partitions with no gaps or overlaps
- non-dim0, incomplete, overlapping, or otherwise incompatible layouts retain exact descriptors; this optimization never converts them into a correctness fallback
- each bounded source adds one complete persistent GPU staging tensor in addition to existing receive buffers, so HBM scales with the number and size of bounded sources
- the threshold is configurable and requires topology/model-specific tuning
- unsupported loader operations remain fail-closed as defined by merged #496

## Test plan

- [x] one signed commit and six changed files on current `main`
- [x] 28 focused reshard/planner/transport tests after rebase
- [x] full Python client suite: 645 passed, 1 skipped
- [x] Ruff check and format
- [x] bit-exact in-memory reconstruction after full staging + local captured-view replay
- [x] incompatible non-dim0 source retains exact descriptors
- [x] descriptor-savings and extra-wire-byte metrics

The managed full-suite process hit the known native dependency teardown SIGSEGV after all tests completed; no assertion failed.

Depends on merged #496.

---

## Correctness status, updated 2026-07-27

Two topologies have since been measured behind a byte-level verification gate
that compares each installed shard against the publisher's digest. The results
change what this PR may claim, so they are stated here rather than left in a
side document.

| Topology | Status | What may be claimed |
| :-- | :-- | :-- |
| A (tensor parallel 2, expert parallel 4, data parallel 8 to tensor parallel 4) | `PERF_PASS_CORRECTNESS_PARTIAL_SCOPE` | Timings, and byte-correctness of the full-pull set only |
| B (pipeline parallel 2, expert parallel 8) | `PERF_PASS_CORRECTNESS_PENDING` | Timings, and byte-correctness of the full-pull set only |

### Topology A: the gate ran and passed, over a partial scope

`checked 6144, skipped_no_digest 0, mismatches 0, divergent_replicas 0`. The
6,144 checks are 48 layers times 128 experts, which is exactly the set this PR
routes through bounded full-pull staging. So the policy this PR introduces is
byte-verified on Topology A.

The scope limit is that attention, normalization and gate/up projections take
the exact-fetch path and carry no digest, so they are unverified rather than
verified-clean. That is a gap in gate coverage, not a known defect.

### Topology B: the mismatches were root-caused, and this path is not the cause

An earlier revision of this description reported nine tensors installing wrong
bytes on Topology B and asked reviewers to treat this path as suspect. That has
since been root-caused to two defects, neither of which is in this PR. The
earlier reading is withdrawn.

The first is a publisher-side naming defect. Megatron's `named_parameters()`
returns pipeline-**local** layer indices, so at `pp_size=2` both stages published
`decoder.layers.0..23`. The shard table is keyed by name with a
first-writer-wins tiebreak, so one stage's 24 layers displaced the other's: 24 of
48 layers were never refit, and where stage 1's offer won, destination layer N
was filled with global layer N+24. The nine reported tensors were the remainder
of that tiebreak, not a transport or geometry fault. Topology A was never
affected, because at `pp_size=1` local index equals global index.

The second is a gate defect. `_prepare()` runs once, so the digests being
compared against described the weights as of the first refit. Any parameter that
training legitimately updated was then reported as corruption. The diagnostic is
unusually clean: across two runs the expected digest was byte-identical while
the delivered one changed, and address re-checking reported no source address
moving across roughly 4.4 million comparisons.

With the naming fix in place, coverage is 100.0% and `params_installed` is
435/435 on all 16 ranks, with 0 real mismatches.

The scope limit that remains is the same one Topology A carries: only full pulls
are digest-checked, so attention other than `o_proj`, normalization and gate/up
projections are unverified rather than verified-clean. Status is therefore
`PENDING` rather than `PASS` on both topologies.

### Topology A and B transport figures

Both rows below were captured with the read batching from #546 and #547 enabled.
The Topology B column is the post-naming-fix run; an earlier revision of this
description quoted a run that refit 51% of the model, and those figures should
not be cited.

| | Topology A | Topology B |
| :-- | --: | --: |
| wire median | 1.512 s | 1.273 s |
| per-rank throughput | 157.6 Gbps | 255.3 Gbps |
| bytes on the wire per rank | 29.78 GB | 40.61 GB |
| useful bytes per rank | 14.58 GB | 30.54 GB |
| extra wire from pulling full tensors | 51% | 24.8% |
| segments per rank | 117,123 | 19,011 |
| unbounded sources | 48 | 0 |
| measured coverage | 100% | 100% |

The geometry, not the fabric, is what separates them: B moves roughly twice the
useful payload of A while issuing about a sixth as many descriptors, and pays a
proportionally smaller redundancy penalty for the strided pulls this PR bounds.

### The promotion threshold is a cliff, and that is a follow-up not a change here

A sweep of `MX_RESHARD_MAX_SEGMENTS_PER_COPY` over 64, 256, 1024 and 4096
produced two distinct plans, not a curve. The first three are identical at
117,123 descriptors and 14.50 GB of extra wire; 4096 promotes nothing and yields
12,693,891 descriptors, which is the explosion this PR exists to prevent. Every
descriptor-heavy copy in this topology needs between 1,025 and 4,096 segments,
so the threshold either promotes all of them or none.

A cost model between two points, one of them catastrophic, re-derives the point
already chosen. The redundancy is reachable only by changing what a promotion
fetches - a bounded full pull covering the union of a source's needed segments
rather than the whole containing tensor, or a per-copy rather than per-source
promotion decision. Both are follow-ups. Reopening this PR for either resets a
review that is otherwise complete.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -41,6 +41,18 @@ from modelexpress.refit.reshard.types import CaptureResult, UnsupportedReshard
 logger = logging.getLogger("modelexpress.refit.reshard.receiver")
 
 
+def _replay_ops(tensor: torch.Tensor, op_chain: tuple) -> torch.Tensor:
+    """Replay a captured loader view chain on a staged full-source tensor."""
+    value = tensor
+    for op_name, args, frozen_kwargs in op_chain:
+        kwargs = dict(frozen_kwargs)
+        if op_name == "__getitem__":
+            value = value.__getitem__(*args)
+        else:
+            value = getattr(value, op_name)(*args, **kwargs)
+    return value
+
+
 class ReshardReceiver:
     """Pull-mode slice-resharding weight receiver (engine-agnostic).
 
@@ -109,6 +121,8 @@ class ReshardReceiver:
             str, torch.Tensor
         ] = {}  # dtype-convert param -> bf16 staging (RDMA target)
         self._staging_ptr: dict[str, int] = {}
+        self._full_staging: dict[str, torch.Tensor] = {}
+        self._full_staging_ptr: dict[str, int] = {}
         self._param_ptr: dict[
             str, int
         ] = {}  # segment param_name -> receive-buffer data_ptr
@@ -227,6 +241,31 @@ class ReshardReceiver:
             )
             self._staging_ptr = {n: t.data_ptr() for n, t in self._staging.items()}
 
+        # Descriptor-heavy strided copies pull each complete source into one
+        # persistent contiguous staging tensor, then replay captured loader views
+        # locally. Each source shard contributes one bounded descriptor.
+        self._full_staging = {}
+        self._full_staging_ptr = {}
+        if plan.full_pulls:
+            with classic_cuda_alloc():
+                self._full_staging = {
+                    full_pull.src_name: torch.empty(
+                        full_pull.global_shape,
+                        dtype=full_pull.dtype,
+                        device=self._device,
+                    )
+                    for full_pull in plan.full_pulls
+                }
+            self._manager.register_tensors(
+                {
+                    f"__full__{name}": tensor
+                    for name, tensor in self._full_staging.items()
+                }
+            )
+            self._full_staging_ptr = {
+                name: tensor.data_ptr() for name, tensor in self._full_staging.items()
+            }
+
         # Receive buffers: one per captured param at its CAPTURED (load-time)
         # shape/dtype, classic cudaMalloc, registered once. The live params are
         # NOT RDMA targets; _install() writes the buffers into the live params.
@@ -251,10 +290,16 @@ class ReshardReceiver:
                 self._param_ptr[name] = self._recv_buffers[name].data_ptr()
 
         logger.info(
-            "[reshard] prepared: %d segments, %d convert(s), %.1f MB/pull, %d fallback",
-            len(plan.segments),
+            "[reshard] prepared: %d descriptor(s), %d full-pull source(s), "
+            "%d convert(s), %.1f MB/pull, %d descriptor(s) saved, "
+            "%.1f MB extra wire, %d unbounded source(s), %d fallback",
+            plan.descriptor_count(),
+            len(plan.full_pulls),
             len(plan.converts),
             plan.bytes_planned() / 1e6,
+            plan.descriptor_savings(),
+            plan.extra_wire_bytes() / 1e6,
+            len(plan.unbounded_sources),
             len(plan.fallback),
         )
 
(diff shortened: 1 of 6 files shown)


KavinKrishnan:
### Correction to the description: the Topology B figures and the correctness claim have both changed

Flagging this as a comment because the earlier description asked reviewers to treat this path as a suspect in a correctness failure, and that was wrong.

**The nine mismatching tensors are not this PR.** They root-cause to a publisher-side naming defect: Megatron's `named_parameters()` returns pipeline-local layer indices, so at `pp_size=2` both stages published `decoder.layers.0..23` and the shard table's first-writer-wins tiebreak let one stage's 24 layers displace the other's. A second, independent defect had the verify gate comparing against digests captured at `_prepare()`, so parameters that training legitimately updated were reported as corruption. With the naming fix, coverage is 100.0% and `params_installed` is 435/435 on all 16 ranks with 0 real mismatches.

**The Topology B transport numbers previously quoted were measured on a refit that covered 51% of the model** and should not be cited. That run moved half the bytes and issued half the descriptors, which made it look like it beat Topology A on every axis. Corrected against the post-fix run:

| | previously quoted (invalid) | corrected |
| :-- | --: | --: |
| wire median | 0.706 s | 1.273 s |
| per-rank throughput | 233.6 Gbps | 255.3 Gbps |
| bytes on the wire per rank | 20.62 GB | 40.61 GB |
| useful bytes per rank | 15.58 GB | 30.54 GB |
| segments per rank | 9,507 | 19,011 |
| measured coverage | 51% | 100% |

The conclusion this PR rests on is unchanged and in fact reads better: the descriptor bound still holds, and the redundancy penalty for pulling full tensors on Topology B is 24.8% rather than the 51% seen on Topology A.

Status moves from `PERF_PASS_CORRECTNESS_FAIL` to `PERF_PASS_CORRECTNESS_PENDING`. It is `PENDING` and not `PASS` because only full pulls are digest-checked on either topology, so attention other than `o_proj`, normalization and gate/up projections are unverified rather than verified-clean.

KavinKrishnan:
/ok to test [6f19286](https://github.com/ai-dynamo/modelexpress/pull/528/commits/6f1928638e52a908a81fabdc96873998745e25fd)

KavinKrishnan:
/ok to test [6f19286](https://github.com/ai-dynamo/modelexpress/pull/528/commits/6f1928638e52a908a81fabdc96873998745e25fd)

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -227,6 +241,31 @@ def _prepare(self, timeout: float) -> None:
...
+        # locally. Each source shard contributes one bounded descriptor.
+        self._full_staging = {}
+        self._full_staging_ptr = {}
+        if plan.full_pulls:
+            with classic_cuda_alloc():
+                self._full_staging = {
+                    full_pull.src_name: torch.empty(
+                        full_pull.global_shape,
+                        dtype=full_pull.dtype,
+                        device=self._device,
```
tanushriya910, line 255: thinking about memory implications here, can probably be optimized by not reserving the memory for all full tensors at once but doing in batches. But this can be a follow up optimization

tanushriya910 (approved):

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: agreed. This currently trades persistent HBM for bounded descriptor count: each promoted source is allocated and registered once, then reused across refits.

follow up PR on batching is needed

PR #536, opened 2026-07-23
fix(refit): keep rendezvous sources ready and fresh

KavinKrishnan:
## Summary

ModelExpress receivers can only pull weights from trainer sources that are both
visible and current in the rendezvous catalog. The existing reshard path violated
both conditions: trainers published sources with an `UNKNOWN` status while
receivers searched only for `READY` sources, and a one-time publication could
expire while a cold receiver was still starting.

This change publishes registered trainer sources as `READY`, keeps them current
through ModelExpress's shared publisher heartbeat, and marks them `STALE` during
graceful shutdown. It fixes source discovery and lifetime management; it does not
change how tensor bytes are planned or transferred.

## Where this fits

The rendezvous catalog is the control-plane handoff between trainer registration
and receiver-side transfer planning:

```mermaid
flowchart LR
    A[Trainer registers NIXL buffers] --> B[Publish source as READY]
    B --> C[ModelExpress rendezvous catalog]
    B -. shared READY status heartbeat .-> C
    C --> D[Receiver discovers READY sources]
    D --> E[Receiver builds transfer plan]
    E --> F[NIXL moves tensor bytes]
```

NVIDIA Inference Xfer Library (NIXL) registration establishes which memory the
trainer can serve. This PR begins after that registration and ends when discovery
returns the source metadata. The transfer planner and NIXL data path consume the
result but are otherwise unchanged.

This is a follow-up to #496, which introduced the no-gather reshard receiver.
The bounded-handshake work in #548 and the Megatron publish path in #552 build on
this lifecycle contract.

### 1. A completed publication was invisible

The publication request omitted its status. Protocol Buffers therefore supplied
the default value, `UNKNOWN`, while receiver discovery explicitly filtered for
`READY`. A trainer could register valid source buffers and publish them
successfully, yet the receiver would never see them.

The publication now sets `SOURCE_STATUS_READY` after registration is complete.
The status therefore means that the advertised source metadata and registered
buffers are ready for discovery.

### 2. A live trainer could disappear during receiver startup

Marking the initial record `READY` fixes visibility but not freshness. The server
reaper expires records that stop refreshing. In the observed failure, 16 trainers
published once, then a cold receiver took long enough to initialize that every
record became `STALE`. READY-filtered discovery returned no sources and waited
until its 1,800-second timeout.

The rendezvous now uses the existing `PublisherThread` to send
`UpdateStatus(READY)` on the standard `MX_HEARTBEAT_INTERVAL_SECS` cadence. This
uses the same lifecycle and centralized environment setting as other ModelExpress
sources instead of maintaining a second thread and reshard-specific variable.

Validated on the 16-trainer Topology A run: discovery returned all 16 sources,
the workload completed all 11 refit steps, and the run attributed 95.9% of
receiver time to measured stages.

## Lifecycle contract

`MxReshardRendezvous.publish()` runs after the NIXL agent and source buffers are
registered. It publishes the full record as `READY` and starts the shared
`PublisherThread`. A later call stops the previous publisher before replacing
the record, which prevents the old publisher from marking the replacement stale.

`close()` delegates heartbeat shutdown and best-effort `STALE` publication to the
same shared publisher lifecycle. Framework adapters should call it during
graceful trainer shutdown. If the process exits abruptly, the heartbeat stops and
the existing server reaper eventually expires the record; this PR does not
provide immediate failure detection.

The heartbeat period comes from centralized `modelexpress.envs` configuration and
must be positive. A non-positive value fails before publication rather than
creating a source that appears healthy but cannot remain discoverable.

## Fail-closed behavior clarified by this PR

Some existing names and documentation implied that unsupported tensor geometry
would use another transfer route. That route does not exist. The historical
`fallback` field records sources the planner cannot serve, and the receiver rejects
the complete update before issuing transfer reads.

The geometry and transfer-plan edits in this PR clarify that contract and correct
tests that exercised a partially valid plan a real receiver would reject. They do
not add a fallback or change supported placement behavior.

## Scope and remaining limits

This PR solves publication visibility and freshness. It does not:

- reconstruct metadata after the server loses an existing source record;
- rebuild an already cached transfer plan when trainer topology or buffer
  addresses change;
- recover a transfer already in progress when a trainer fails;
- make unsupported tensor geometry transferable;
- change staging-memory ownership or the engine's post-load weight transforms;
- replace framework teardown with a general re-registration orchestrator.

Those are separate receiver, transport, or framework-lifecycle concerns.

## Validation

- [x] Publications use `READY` and are visible to READY-filtered discovery.
- [x] The shared publisher sends `READY` status heartbeats without re-publishing
      the full record.
- [x] Repeated publication replaces the prior publisher without leaving an old
      heartbeat thread.
- [x] `close()` uses the shared publisher to stop heartbeats and request `STALE`.
- [x] A non-positive heartbeat period fails before publication.
- [x] Unsupported-source tests confirm that no transport read is issued.
- [x] 62 focused rendezvous, publisher, environment, geometry, planner, transfer,
      and transport tests pass after merging current `main`.
- [x] Full Python client suite: 665 passed, 1 skipped.
- [x] Ruff and whitespace checks pass.

The managed full-suite process hit the known native-dependency teardown
segmentation fault after all tests completed; no test assertion failed.

## Reviewer guide

The PR changes 12 files (+354/-65), but the behavioral change is concentrated in
two places:

1. `modelexpress/refit/reshard/rendezvous.py` — review the `READY` publication,
   reuse of `PublisherThread`, repeated-publication ordering, and `close()` first.
2. `tests/test_reshard_refit_rendezvous.py` — review visibility, shared heartbeat,
   shutdown, and invalid-period coverage.

The README and architecture document record the shared heartbeat setting and
lifecycle. The remaining reshard files and tests align terminology with the
existing fail-closed receiver contract.


modelexpress_client/python/modelexpress/refit/reshard/types.py
```diff
@@ -45,9 +45,9 @@ class RecordedCopy:
 class CaptureResult:
     """Output of a bake: the recorded copies plus what fell back.
 
-    ``unsupported`` = source names whose loader used an unsupported op (full-pull
-    those). ``unattributed`` = copy_ calls fired with no active loader stamp (the
-    destination param can't be attributed -> also full-pull)."""
+    ``unsupported`` = source names whose loader used an unsupported op.
+    ``unattributed`` = copy_ calls fired with no active loader stamp. Either
+    condition makes the update fail closed in the current receiver."""
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `d690dfa`. The opening sentence now reads "the recorded copies plus what could not be attributed," and the closing sentence says explicitly that there is no fallback path serving those tensors by another route.

Worth keeping the wording tight here for the reason you implied: `fallback` retains a historical field name from when a per-source fallback genuinely existed, so anything in a docstring that reinforces that reading sends a reviewer looking for a recovery path that was removed.

modelexpress_client/python/tests/test_reshard_refit_transfer.py
```diff
@@ -133,7 +133,7 @@ def test_strided_source_reconstructs_exactly():
     )
     sources = {"row": SourceInfo(tuple(row_src.shape), torch.float32, EL, [shard])}
 
-    # Only the 'row' copy is planned here (others have no source -> fallback).
+    # Only the 'row' copy is planned here; missing sources are marked unsupported.
```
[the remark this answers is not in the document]
KavinKrishnan: Good catch on the first half, and I've taken it. Pushed `d690dfa`. I've implemented the second half differently, and want to explain why rather than quietly diverge.

**The row-only capture: agreed, done.** `test_strided_source_reconstructs_exactly` captured the whole manifest while publishing only `row`, so five sources landed in `plan.fallback` and the test executed the plan's valid segments anyway. That is worse than untidy — a plan carrying unsupported sources is one `ReshardReceiver._prepare` refuses outright, so the test was asserting reconstruction against a plan that can never reach a receiver. It now captures a row-only manifest and asserts `plan.fallback == []`, which is what makes the strided reconstruction it exists to test actually meaningful.

**On "verify that a fallback plan is rejected": rejection is not this layer's job.** `execute_transfer` deliberately does not reject. It passes `fallback` through in its returned stats so the *receiver* can, and the refusal lives in the `if plan.fallback:` guard in `ReshardReceiver._prepare`, which raises `UnsupportedReshard` before any transport call. Asserting rejection at `execute_transfer` would pin behaviour that does not exist and that I do not think should — the planner and executor stay mechanism, the receiver holds policy.

So I've split your ask across the two layers where each half is true:

- The unsupported-source test now asserts the fail-closed *consequence* at this layer: the unsupported source gets no planned segment, and executing the plan issues zero reads against that session. That is the "performs zero transport reads" part, verified with a recording transport.
- A new test pins the contract that makes receiver-side rejection possible at all: `execute_transfer` surfaces `fallback` in its stats. A receiver that cannot see the list cannot fail closed, so this is the invariant worth protecting.

One gap I'd rather name than paper over: the receiver's raise itself is still untested, because reaching `_prepare` requires the discovery and P2P metadata handshake stack. That is the test that would most directly cover the contract, and it needs a receiver-level harness that does not exist yet. Tracked, not done here.

KavinKrishnan:
/ok to test [d690dfa](https://github.com/ai-dynamo/modelexpress/pull/536/commits/d690dfae1e0cc959f165fa6c5699be7cfccdb96c)

KavinKrishnan:
/ok to test [d690dfa](https://github.com/ai-dynamo/modelexpress/pull/536/commits/d690dfae1e0cc959f165fa6c5699be7cfccdb96c)

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -289,13 +299,103 @@ def _identity(self, role: str) -> "p2p_pb2.SourceIdentity":
...
+        )
+        try:
+            period = float(raw)
+        except ValueError as exc:
+            raise ValueError(f"{_HEARTBEAT_ENV} must be a number, got {raw!r}") from exc
+        if period <= 0:
+            raise ValueError(f"{_HEARTBEAT_ENV} must be positive, got {period}")
+        return period
+
+    def _start_rendezvous_heartbeat(self, worker, period: float) -> None:
```
tanushriya910: wondering if this logic is already there in the current MX implementation too and if we can reuse it here?

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -289,13 +299,103 @@ def _identity(self, role: str) -> "p2p_pb2.SourceIdentity":
...
         self._mx_source_id = self.client.publish_metadata(
             self._identity(self.role), worker, self.worker_id
         )
+        self._start_rendezvous_heartbeat(worker, heartbeat_period)
         return self._mx_source_id
 
+    @staticmethod
+    def _heartbeat_period() -> float:
+        raw = os.environ.get(
+            _HEARTBEAT_ENV,
```
tanushriya910: should we follow the same env var naming as MX for the heartbeat period?

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -289,13 +299,103 @@ def _identity(self, role: str) -> "p2p_pb2.SourceIdentity":
...
+        )
         self._mx_source_id = self.client.publish_metadata(
             self._identity(self.role), worker, self.worker_id
         )
+        self._start_rendezvous_heartbeat(worker, heartbeat_period)
         return self._mx_source_id
 
+    @staticmethod
+    def _heartbeat_period() -> float:
+        raw = os.environ.get(
```
nv-hwoo: we should use our centralized env.py

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -289,13 +299,103 @@ def _identity(self, role: str) -> "p2p_pb2.SourceIdentity":
...
+        heartbeat_period = self._heartbeat_period()
+        worker = p2p_pb2.WorkerMetadata(
+            worker_rank=self.rank,
+            nixl_metadata=blob,
+            status=p2p_pb2.SOURCE_STATUS_READY,
+        )
         self._mx_source_id = self.client.publish_metadata(
             self._identity(self.role), worker, self.worker_id
         )
+        self._start_rendezvous_heartbeat(worker, heartbeat_period)
```
nv-hwoo: can we use our existing publisher thread?

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
nv-hwoo: haha you beat me to it :D

KavinKrishnan: Yes—this was duplicating ModelExpress lifecycle code. Fixed in `1677edd`: `MxReshardRendezvous` now uses the existing `PublisherThread` for READY heartbeats and graceful STALE shutdown. The bespoke reshard heartbeat thread and full-record refresh loop are removed.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
KavinKrishnan: Agreed. Fixed in `1677edd`: reshard rendezvous now uses the existing `MX_HEARTBEAT_INTERVAL_SECS` setting (30 seconds by default), matching the rest of ModelExpress. The separate `MX_RESHARD_HEARTBEAT_S` variable has been removed from code and documentation.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
KavinKrishnan: Fixed in `1677edd`. The rendezvous reads `envs.MX_HEARTBEAT_INTERVAL_SECS`; direct `os.environ` parsing and the local default were removed. The non-positive guard remains at the publication boundary so an invalid cadence cannot create a READY source without a viable heartbeat.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
KavinKrishnan: Yes. Fixed in `1677edd`: the custom daemon, event, and lock are replaced by `metadata.publisher.PublisherThread`. Repeated publication stops the prior publisher before replacing the record, and `close()` delegates READY/STALE lifecycle handling to the shared implementation.

KavinKrishnan:
/ok to test [1677edd](https://github.com/ai-dynamo/modelexpress/pull/536/commits/1677eddb589bdb75dbd5b990dea51b2772cf634a)

nv-hwoo (approved):

PR #538, opened 2026-07-23
test(client): gate cuda allocation tests on cuda bindings

yafshar:
## What

Guard the `fake_driver` fixture in `test_pool_registration.py` with
`pytest.importorskip("cuda.bindings.driver")` so the CUDA allocation-discovery
tests skip on hosts without the `cuda` Python bindings instead of erroring at
fixture setup.

## Why

On a non-CUDA host, such as an XPU-only node, the fixture's unconditional
`from cuda.bindings import driver` raised `ModuleNotFoundError` during setup.
That caused the seven CUDA allocation-discovery tests in
`TestFindCudaAllocations` and `TestRawDescriptorMemType` to report as ERRORs
rather than skips.

These tests exercise CUDA-driver allocation discovery and legitimately cannot
run where the CUDA Python bindings are absent.

## Changes

- `modelexpress_client/python/tests/test_pool_registration.py`: replace the
  fixture's unconditional `from cuda.bindings import driver` with
  `driver = pytest.importorskip("cuda.bindings.driver")`.
- Document the skip behavior in the fixture docstring.

## Testing

- CUDA bindings present: `tests/test_pool_registration.py` -> 21 passed.
- CUDA bindings absent, simulated by blocking the `cuda` import: 14 passed,
  7 skipped.
- XPU node run confirms the seven CUDA setup errors are gone and now report as
  skips.

## Summary by CodeRabbit

* **Tests**
  * Improved test compatibility on hosts without CUDA bindings.
  * Allocation-discovery tests now skip automatically when the required CUDA components are unavailable.
  * Preserved existing validation of memory-address range calls.


zhengluo-nv:
/ok to test [2719e0c](https://github.com/ai-dynamo/modelexpress/pull/538/commits/2719e0c5dad9a1e64aa3f1d826360c1f8ed3c38f)

zhengluo-nv (approved):

PR #547, opened 2026-07-27
perf(reshard-refit): issue the three refit read phases as one batch

KavinKrishnan:
## Summary

A reshard refit reads three independent groups of bytes:

- exact segments directly into receive buffers;
- descriptor-heavy tensors into full-pull staging;
- dtype-mismatched tensors into conversion staging.

The receiver previously issued and drained each group separately. This PR builds
all three descriptor lists first and issues them as one transport batch. Local
re-slicing and dtype conversion still run only after every read completes.

## Why this is needed

The three sequential drains introduced barriers that the data flow does not
require. A full-pull read could not begin until every exact read completed, and
a conversion read could not begin until every full pull completed, even though
the groups write to different destinations.

```mermaid
flowchart LR
    subgraph Before
        E1[Read exact] --> F1[Read full pulls] --> C1[Read conversions]
        C1 --> R1[Re-slice] --> X1[Cast] --> I1[Install]
    end
    subgraph After
        B2[Read exact + full + conversion] --> R2[Re-slice]
        R2 --> X2[Cast] --> I2[Install]
    end
```

`exact_descriptors()` is extracted from `execute_transfer()` so the receiver can
build the exact-read descriptors without performing that read immediately.

## Correctness argument

Fusion is safe because:

1. Exact reads, full pulls, and conversion reads target disjoint buffers.
2. Full-pull re-slicing starts after the fused batch completes.
3. Dtype conversion starts after the fused batch completes.
4. Installation remains after all reads and transformations.

The CPU reference-transport tests execute real byte copies through both the fused
and phased paths. They compare final buffers against each other and against
independently constructed expected values. Mutation checks confirmed that
removing a read group fails the tests, while reordering independent descriptors
does not.

## Diagnostic control

Fusion is enabled by default. Set `MX_RESHARD_FUSED_WIRE=0` to restore the three
phased drains. Phased mode remains useful for attributing wire time to exact,
full-pull, and conversion traffic separately.

The flag is registered in the centralized environment configuration and
documented in the client README variable table alongside the other reshard
variables. Common case-insensitive boolean spellings are accepted. Invalid values
emit a warning and use the default instead of silently changing an A/B run.

## Performance evidence

Status: `PERF_PASS_CORRECTNESS_PENDING`.

A preliminary Topology A capture used real Qwen3-30B-A3B BF16 tensors, 16
Megatron trainer ranks (TP=2, PP=1, EP=4, ordinary DP=8) and vLLM generation at
TP=4 over AWS Elastic Fabric Adapter with the NIXL `LIBFABRIC` backend. Each
receiver rank planned 29.78 GB across 117,123 segments.

The phased wire median was 1.474 seconds. The fused wire median was 1.388
seconds, a 5.8% reduction. Median per-rank throughput increased from 161.6 to
171.7 Gbps.

These numbers are directional, not a publishable benchmark row:

- only 12 of 16 inference ranks completed the fused capture;
- observed fused rank times ranged from 1.06 to 1.83 seconds, wider than the
  median improvement;
- the capture does not provide the required three cold starts plus ten measured
  warm versions and fleet-critical maxima;

The supported conclusion is that fusion removes unnecessary barriers without
changing the transfer plan or installed bytes. It does not establish a complete
fleet-level speedup claim.

## Scope

This PR changes batching and accounting around existing transfer descriptors. It
does not change:

- source selection or discovery;
- tensor coverage or transfer-plan geometry;
- the number of useful or extra bytes;
- dtype conversion semantics;
- engine-specific installation behavior.

## Validation

- [x] Rebased onto current `main`; the obsolete #528 commit was dropped because
      its patch is already upstream.
- [x] 45 complete `test_reshard_refit_*` tests pass after the rebase.
- [x] 21 focused environment, fused-wire, and transfer tests pass.
- [x] Fused and phased paths produce identical buffers and accounting.
- [x] Expected tensor values are checked independently of path parity.
- [x] Installation is asserted to occur after all wire batches in both modes.
- [x] Empty full-pull and conversion groups issue no empty reads in either mode.
- [x] Boolean flag parsing covers mixed-case true/false, yes/no, and on/off.

Full GitHub CI is running again on the rebased branch.

## Reviewer guide

The PR changes seven files (+440/-59). Suggested order:

1. `modelexpress/refit/reshard/receiver.py` — verify that all descriptor lists
   are built before the fused read and that re-slice, cast, and install remain
   after it.
2. `tests/test_reshard_refit_fused_wire.py` — review buffer parity, independent
   expected-value checks, ordering, accounting, and empty-group coverage.
3. `modelexpress/refit/reshard/transfer_plan.py` — confirm
   `exact_descriptors()` is a pure extraction and `execute_transfer()` preserves
   its previous behavior.
4. `modelexpress/envs.py`, `tests/test_envs.py`, and
   `modelexpress_client/python/README.md` — verify default-enabled flag parsing,
   warning behavior for invalid values, and the documented default.
5. `tests/test_reshard_refit_transfer.py` — confirm descriptor extraction returns
   exactly the descriptors `execute_transfer()` reads.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -24,6 +24,7 @@ transport, buffers and the router dtype-cast are shared here.
 from __future__ import annotations
 
 import logging
+import os
 
 import torch
 
@@ -31,7 +32,11 @@ from modelexpress.client import MxClient
 from modelexpress.nixl_transfer import NixlTransferManager
 from modelexpress.refit.reshard.cuda_pool import classic_cuda_alloc
 from modelexpress.refit.reshard.rendezvous import gather_sources
-from modelexpress.refit.reshard.transfer_plan import execute_transfer, plan_transfer
+from modelexpress.refit.reshard.transfer_plan import (
+    exact_descriptors,
+    execute_transfer,
+    plan_transfer,
+)
 from modelexpress.refit.reshard.transport import (
     NixlReshardTransport,
     ReadDescriptor,
@@ -39,6 +44,15 @@ from modelexpress.refit.reshard.transport import (
 from modelexpress.refit.reshard.types import CaptureResult, UnsupportedReshard
 
 logger = logging.getLogger("modelexpress.refit.reshard.receiver")
+
+
+def _fused_wire_enabled() -> bool:
+    """Whether to issue the exact, full-pull and convert reads as one batch.
+
+    Read at call time so an A/B can toggle it without re-importing. Set
+    ``MX_RESHARD_FUSED_WIRE=0`` to drain each phase in turn.
+    """
+    return os.environ.get("MX_RESHARD_FUSED_WIRE", "1") not in ("0", "false")
 
 
 def _replay_ops(tensor: torch.Tensor, op_chain: tuple) -> torch.Tensor:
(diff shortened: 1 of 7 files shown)


modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -41,6 +46,15 @@
...
 
 
+def _fused_wire_enabled() -> bool:
+    """Whether to issue the exact, full-pull and convert reads as one batch.
+
+    Read at call time so an A/B can toggle it without re-importing. Set
+    ``MX_RESHARD_FUSED_WIRE=0`` to drain each phase in turn.
+    """
+    return os.environ.get("MX_RESHARD_FUSED_WIRE", "1") not in ("0", "false")
+
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `878614c`. The flag now uses the shared environment-variable parser, so values such as `False`, `no`, and `off` work as expected. Invalid values produce a warning and use the default. Tests cover valid true, valid false, and invalid values.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -41,6 +46,15 @@
 logger = logging.getLogger("modelexpress.refit.reshard.receiver")
 
 
+def _fused_wire_enabled() -> bool:
+    """Whether to issue the exact, full-pull and convert reads as one batch.
+
+    Read at call time so an A/B can toggle it without re-importing. Set
+    ``MX_RESHARD_FUSED_WIRE=0`` to drain each phase in turn.
```
tanushriya910, line 53: this looks like a good change to issue in one batch. not sure it would give much perf gain though. wondering if we need to introduce an extra env variable would doing it serially ever provide a benefit over batch?

tanushriya910 (commented):
I do see a correctness argument in the overview but not correctness tests proving it.

> Topology B had not passed its correctness gate for this capture.

does this mean it resulted in incorrect results?

KavinKrishnan:
This was outdated wording in the PR description. Topology B correctness is fixed; the reported failure came from the verification instrumentation, not corrupted weights.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: The measured gain is modest: median wire time improved from 1.474 s to 1.388 s (5.8%). The wider rank-to-rank variation means we should not claim a broad performance win yet.

Serial execution is not expected to be faster. The switch remains useful for two reasons: it lets us measure the three transfer groups separately, and it provides a quick rollback if batching causes problems on a transport we have not tested.

A better long-term solution is to keep batching and add per-group timing. I suggest handling that separately rather than expanding this PR.

tanushriya910 (approved):

KavinKrishnan:
/ok to test [2715229](https://github.com/ai-dynamo/modelexpress/pull/547/commits/2715229720e95859c722b015c340236066a18b9c)

PR #548, opened 2026-07-27
fix(reshard-refit): bound the peer handshake and reject empty publishers

KavinKrishnan:
## Summary

This PR bounds and retries the receiver's peer-metadata handshake, narrows it to
the trainers a receiver actually reads from, and prevents READY publishers with
empty shard tables from satisfying the discovery quorum. Together these turn an
ambiguous refit hang into either successful progress or a bounded error naming
the peers that did not answer.

The PR targets current `main`; #536 is already merged and its former base commits
have been removed.

## Why this is needed

Before weight reads begin, each receiver fetches NIXL metadata from its trainer
peers. The previous loop attempted peers once, in catalog order, and allowed one
peer to consume the full refit timeout. A dead or temporarily busy publisher
could therefore block all healthy peers and produce only a generic timeout.

It also dialed every discovered trainer, including ones it would never read a
byte from. The handshake exists to resolve a peer's remote memory registrations
so reads can address them, so a peer that is never read from costs a dial and
buys nothing. Unnarrowed, the dial count is one per receiver-trainer pair and
grows as the product of both sides.

Discovery had a related problem: a READY source with an empty shard table counted
toward the trainer quorum even though it had no registered tensors to serve. The
receiver could stop discovery early and then stall while handshaking a source
that could never contribute bytes.

## Handshake behavior

`handshake_with_peers()` applies one budget across all peers and retries:

- `MX_RESHARD_HANDSHAKE_TIMEOUT_S` controls the total handshake budget;
- `MX_RESHARD_HANDSHAKE_ATTEMPT_S` caps one dial and is clamped to the remaining
  total budget;
- failed peers move to the back of the queue so healthy peers can progress;
- a full pass without progress waits `MX_RESHARD_HANDSHAKE_BACKOFF_S` before
  retrying;
- timeout errors identify every outstanding peer, endpoint, attempt count, and
  last error.

A peer that succeeds is removed from the queue and is not dialed again.
Endpoints split on the final colon, preserving IPv6 literal handling, and are
parsed inside the retry so an unparseable catalog entry is reported as that
peer's failure rather than raised past every other peer.

All three bounds are registered in `modelexpress/envs.py` and documented in the
client README variable table, so they parse and warn like every other variable
and are read live rather than frozen at import. A value that is not a finite
positive number — zero, negative, unparseable, `inf`, or `nan` — warns and falls
back to the documented default, because each of those disables the bound rather
than tightening it.

## Which peers get dialed

The handshake now runs after `plan_transfer()` rather than immediately after
discovery, and dials only the trainers the plan reads from.

Discovery still has to collect every trainer's shard table: which trainers hold
the bytes a given receiver is missing is only known once the slice arithmetic is
done. The handshake has no such constraint — it only has to precede the first
read. By the time the plan exists, `TransferPlan.sessions()` names the sessions
actually read from, and `handshake_endpoints_for_plan()` narrows the endpoint map
to their agents. On a topology where each receiver pulls from a fraction of the
trainers, that fraction is what gets dialed.

Two smaller consequences of the reorder. An unsupported plan is now rejected
before the handshake budget is spent, which is the ordering a review comment
flagged on the earlier revision. And a trainer that the plan reads from but which
published no endpoint now fails closed with that trainer named, rather than
deferring the failure to `prep_xfer_dlist`, which cannot say which peer's
metadata was missing.

`TransferPlan` grows an `_all_segments()` helper for this. Segments live in three
places — read straight into live params, into dtype-conversion staging, and into
full-pull staging — and any question about the plan as a whole has to ask all
three. `bytes_planned()`, `descriptor_count()`, and the new `sessions()` now
share one traversal instead of each re-enumerating the phases.

**This is scaling headroom, not a measured speedup.** On Topology A the handshake
is a 3.2 s median across 16 peers, about 4% of an 87 s cold start, and `_prepare`
runs once per run because the plan is cached and reused every step. Capture and
plan are 51 s of that same 87 s and remain where cold-start effort belongs.

## Discovery behavior

`discover_trainers()` now fetches metadata while polling and counts a READY rank
only when its decoded shard table is non-empty. Empty publishers are excluded
from the returned payloads, and collection stops once the requested number of
non-empty ranks is in hand: a source left READY by an earlier run is both an
extra peer to handshake and a second set of shards competing to describe the
same tensor names.

Metadata is inspected for every visible READY source, whatever the count.
Gating that inspection on the READY count already meeting the quorum defeated
the purpose of the change, because below the quorum the timeout then reported
zero non-empty and zero empty tables - the one number that distinguishes "a rank
has not come up yet" from "a rank came up with nothing to serve". Timeout
messages report READY, non-empty, and empty counts separately.

This moves metadata reads into the polling loop. That adds repeated catalog reads
while waiting, but it is necessary because the source listing does not contain
the shard table needed to decide whether a publisher can serve tensors.

## Scope and limits

This PR does not terminate a trainer process that outlives its driver, add run or
epoch identity, or recover a transfer already in progress. The shared publisher
heartbeat from #536 and the existing server reaper handle normal freshness and
stale expiry; an orphan process that remains alive can still renew its own
source.

The change limits and identifies handshake failures, and stops the receiver
dialing peers it has no reason to talk to. It does not make an incorrect or
abandoned publisher safe to use.

## Validation

- [x] Rebased onto current `main`; merged #536 base commits were dropped.
- [x] 71 `test_reshard_refit_*` / `test_reshard_*` tests pass, 22 of them in the
      handshake file.
- [x] Tests cover total and per-attempt deadlines, retry ordering, backoff,
      progress logging, IPv6 endpoints, malformed endpoints, fallback on
      non-finite and non-positive bounds, over-quorum collection, and partial
      READY reporting.
- [x] Peer-narrowing tests cover selection from all three plan phases, address
      preservation, the fail-closed missing-endpoint case, and an empty plan
      dialing nobody.
- [x] The merged #536 shared-publisher test still expects one metadata publish;
      READY heartbeats remain status updates rather than full re-publications.
- [x] Core Ruff checks and whitespace validation pass. The 60 failures in
      `test_vmm_*` are pre-existing on this branch point and require CUDA VMM;
      they touch none of these files.

## Reviewer guide

1. `refit/reshard/receiver.py` — review `handshake_with_peers()` and its deadline
   arithmetic first. Confirm each attempt is bounded by both the attempt ceiling
   and remaining total budget, and that everything that can fail per peer sits
   inside the retry.
2. `refit/reshard/receiver.py`, `_prepare()` — check the new ordering: discover,
   capture, plan, reject unsupported, then narrow and handshake. `session_to_agent`
   comes from `gather_sources()` and the handshake must still precede the first
   read in `execute_transfer()`.
3. `refit/reshard/receiver.py`, `handshake_endpoints_for_plan()` — the narrowing
   itself, in particular that it fails closed on a planned trainer with no
   endpoint rather than silently dropping it.
4. `refit/reshard/transfer_plan.py` — `_all_segments()` and `sessions()`.
   Worth confirming the three phases are all walked, since missing one would drop
   a peer the plan genuinely reads from.
5. `tests/test_reshard_refit_handshake.py` — per-attempt capping, failed peer
   deferral, no-progress backoff, completed-peer removal, the malformed endpoint
   case, and the narrowing tests at the end of the file.
6. `refit/reshard/rendezvous.py` — verify only non-empty publishers count toward
   quorum, that collection stops at the requested count, and that timeout
   accounting reflects every visible source.
7. `tests/test_reshard_refit_rendezvous.py` — verify heartbeat behavior from #536
   and the new empty-table, over-quorum, and partial-set cases are all preserved.
8. `modelexpress/envs.py` and `modelexpress_client/python/README.md` — the three
   handshake bounds and their documented defaults.

Catching broad `Exception` around one dial is deliberate: failures remain
retryable inside the bounded budget, while the final error retains peer-specific
diagnostics.

### DIFF
--- a/modelexpress_client/python/modelexpress/envs.py
+++ b/modelexpress_client/python/modelexpress/envs.py
@@ -61,6 +61,9 @@ if TYPE_CHECKING:
     MX_RESHARD_FUSED_WIRE: bool
     MX_RESHARD_REQUIRE_FULL_COVERAGE: bool
     MX_RESHARD_COVERAGE_FLOOR: float
+    MX_RESHARD_HANDSHAKE_TIMEOUT_S: float
+    MX_RESHARD_HANDSHAKE_ATTEMPT_S: float
+    MX_RESHARD_HANDSHAKE_BACKOFF_S: float
     # Kubernetes service backend
     MX_K8S_SERVICE_PATTERN: str
     MX_K8S_SOURCE_RETRIES: str
@@ -205,6 +208,12 @@ environment_variables: dict[str, Callable[[], Any]] = {
     .lower()
     in _TRUTHY,
     "MX_RESHARD_COVERAGE_FLOOR": lambda: _env_float("MX_RESHARD_COVERAGE_FLOOR", 0.995),
+    # Whole-handshake budget across every peer and every retry, the per-dial
+    # ceiling, and the pause after a full pass over the pending peers makes no
+    # progress. See modelexpress.refit.reshard.receiver.handshake_with_peers.
+    "MX_RESHARD_HANDSHAKE_TIMEOUT_S": lambda: _env_float("MX_RESHARD_HANDSHAKE_TIMEOUT_S", 900.0),
+    "MX_RESHARD_HANDSHAKE_ATTEMPT_S": lambda: _env_float("MX_RESHARD_HANDSHAKE_ATTEMPT_S", 20.0),
+    "MX_RESHARD_HANDSHAKE_BACKOFF_S": lambda: _env_float("MX_RESHARD_HANDSHAKE_BACKOFF_S", 2.0),
     # ── Kubernetes service backend ─────────────────────────────────────────
     "MX_K8S_SERVICE_PATTERN": lambda: os.environ.get("MX_K8S_SERVICE_PATTERN", "mx-sources"),
     "MX_K8S_SOURCE_RETRIES": lambda: os.environ.get("MX_K8S_SOURCE_RETRIES", ""),
--- a/modelexpress_client/python/modelexpress/envs.py
+++ b/modelexpress_client/python/modelexpress/envs.py
@@ -169,6 +169,27 @@ def _env_bool(name: str, default: bool) -> bool:
     return default
 
 
+def _env_positive_float(name: str, default: float) -> float:
+    """As :func:`_env_float`, for a variable where zero and negatives are meaningless.
+
+    Timeouts, per-attempt ceilings and backoff pauses are all bounds, and a bound of
+    zero or less is not a smaller bound but an absent one: it turns the loop it
+    governs into an unbounded one, which is the failure the variable exists to
+    prevent. Falling back to the documented default keeps the bound in place, and
+    the warning says which value was ignored.
+    """
+    value = _env_float(name, default)
+    if value <= 0.0:
+        logger.warning(
+            "Invalid %s=%r; a bound must be positive, using default %s",
+            name,
+            os.environ.get(name),
+            default,
+        )
+        return default
+    return value
+
+
 # One entry per variable. The lambda owns the default and parsing; callers that
 # need a site-specific default receive the raw value (``None`` when unset) and
 # apply their own fallback.
@@ -208,12 +229,25 @@ environment_variables: dict[str, Callable[[], Any]] = {
     .lower()
     in _TRUTHY,
     "MX_RESHARD_COVERAGE_FLOOR": lambda: _env_float("MX_RESHARD_COVERAGE_FLOOR", 0.995),
-    # Whole-handshake budget across every peer and every retry, the per-dial
-    # ceiling, and the pause after a full pass over the pending peers makes no
-    # progress. See modelexpress.refit.reshard.receiver.handshake_with_peers.
-    "MX_RESHARD_HANDSHAKE_TIMEOUT_S": lambda: _env_float("MX_RESHARD_HANDSHAKE_TIMEOUT_S", 900.0),
-    "MX_RESHARD_HANDSHAKE_ATTEMPT_S": lambda: _env_float("MX_RESHARD_HANDSHAKE_ATTEMPT_S", 20.0),
-    "MX_RESHARD_HANDSHAKE_BACKOFF_S": lambda: _env_float("MX_RESHARD_HANDSHAKE_BACKOFF_S", 2.0),
+    # Bounds for the reshard peer handshake; see
+    # modelexpress.refit.reshard.receiver.handshake_with_peers.
+    #
+    # TIMEOUT_S is the budget across every peer and every retry. A refit timeout is
+    # the wrong bound, since it lets one unreachable peer consume the whole refit,
+    # but it must stay generous: a peer registering tens of GB with the fabric
+    # provider blocks its listen thread for minutes, and waiting is the correct
+    # response to that. ATTEMPT_S caps a single dial, so an unreachable peer frees
+    # the budget for a different one instead of blocking on it. BACKOFF_S is the
+    # pause after a full pass over the pending peers makes no progress.
+    "MX_RESHARD_HANDSHAKE_TIMEOUT_S": lambda: _env_positive_float(
+        "MX_RESHARD_HANDSHAKE_TIMEOUT_S", 900.0
+    ),
+    "MX_RESHARD_HANDSHAKE_ATTEMPT_S": lambda: _env_positive_float(
+        "MX_RESHARD_HANDSHAKE_ATTEMPT_S", 20.0
+    ),
+    "MX_RESHARD_HANDSHAKE_BACKOFF_S": lambda: _env_positive_float(
+        "MX_RESHARD_HANDSHAKE_BACKOFF_S", 2.0
+    ),
     # ── Kubernetes service backend ─────────────────────────────────────────
     "MX_K8S_SERVICE_PATTERN": lambda: os.environ.get("MX_K8S_SERVICE_PATTERN", "mx-sources"),
     "MX_K8S_SOURCE_RETRIES": lambda: os.environ.get("MX_K8S_SOURCE_RETRIES", ""),
--- a/modelexpress_client/python/modelexpress/envs.py
+++ b/modelexpress_client/python/modelexpress/envs.py
@@ -30,6 +30,7 @@ Not covered here (intentional exceptions):
 from __future__ import annotations
 
 import logging
+import math
 import os
 from typing import TYPE_CHECKING, Any, Callable, Optional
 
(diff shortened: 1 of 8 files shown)


modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -40,6 +43,135 @@
...
+                f"{succeeded} of {total} peer(s) answered. Outstanding: "
+                f"{outstanding}. These publishers are advertised in the MX catalog "
+                f"but did not answer - either the process is gone while something "
+                f"still heartbeats its source, or its NIXL listen thread is not "
+                f"accepting."
+            )
+
+        agent_name, endpoint = pending.popleft()
+        host, port_str = endpoint.rsplit(":", 1)
+        this_timeout = max(1.0, min(attempt_timeout, remaining))
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `47d5f4c`. Endpoint parsing now happens inside the retry block, so one malformed endpoint is reported as that peer's failure without preventing connections to healthy peers. The regression test covers this case.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -352,21 +361,36 @@ def discover_trainers(
...
+                    meta = self.client.get_metadata(inst.mx_source_id, inst.worker_id)
+                    if not meta.found:
+                        continue
+                    payload = unwrap_rendezvous_blob(meta.worker.nixl_metadata)
+                    if not payload[3]:
+                        empty += 1
+                        continue
+                    payloads.append(payload)
+                if len(payloads) >= expected_trainers:
+                    break
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `47d5f4c`. Discovery now stops after collecting the requested number of non-empty trainer sources. This avoids connecting to stale extra sources or accepting conflicting descriptions of the same tensors. The over-quorum case is covered by a test.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -352,21 +361,36 @@ def discover_trainers(
...
             if time.monotonic() >= deadline:
                 raise TimeoutError(
-                    f"timed out after {timeout}s waiting for {expected_trainers} trainer ranks "
-                    f"(saw {len(instances)})"
+                    f"timed out after {timeout}s waiting for {expected_trainers} "
+                    f"trainer ranks (saw {len(instances)} READY source(s), "
+                    f"{len(payloads)} with a non-empty shard table, {empty} empty)"
                 )
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `47d5f4c`. We now inspect every visible READY source before checking whether the requested count was reached. Timeout messages therefore distinguish a missing trainer from a trainer that is READY but has no tensors. A regression test covers the partial set case.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -41,6 +44,152 @@
 logger = logging.getLogger("modelexpress.refit.reshard.receiver")
 
 
+def _handshake_seconds(name: str) -> float:
```
nv-hwoo: let's make this a `_env_positive_float()` method in `envs.py` as this has nothing to do with handshake envs specifically.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -41,6 +44,152 @@
...
+    A non-positive value turns the corresponding bound off, which silently
+    reintroduces the failure mode it exists to prevent, so it is rejected.
+    """
+    value = float(getattr(envs, name))
+    if not value > 0.0:
+        raise ValueError(f"{name} must be a positive number of seconds, got {value}")
+    return value
+
+
+def handshake_with_peers(
```
tanushriya910, line 91: thanks for adding this improvement

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -352,21 +366,41 @@ def discover_trainers(
...
+            # Every visible READY source is inspected, whatever the count: with
+            # fewer sources than expected the shard-table state is exactly what a
+            # timeout here has to report, and reaching the quorum on instances
+            # alone says nothing about whether any of them published bytes.
+            for inst in instances:
+                meta = self.client.get_metadata(inst.mx_source_id, inst.worker_id)
+                if not meta.found:
+                    continue
+                payload = unwrap_rendezvous_blob(meta.worker.nixl_metadata)
+                if not payload[3]:
```
tanushriya910: can we do named tuple instead payload[3] index for better readability

nv-hwoo (commented):
One small comment and one question: Am I understanding correctly that each receiver (engine) rank is trying to handshake with every trainer ranks? If the answer is yes, is it possible to delay the handshake to after the transfer plan construction so that we can use that information to fetch only from the trainers that it reads from, avoiding O(M*N) fetches?

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: Fixed in `00c55ad`. The shared `_env_positive_float()` helper now lives in `envs.py`, and the handshake settings use it directly. Invalid or non-positive values produce a warning and use the documented default. Tests cover valid, non-positive, and invalid values.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
KavinKrishnan: Fixed in `00c55ad`. The decoded value is now a `RendezvousPayload` named tuple, so code can use clear fields such as `payload.tensors`. Existing tuple unpacking and indexing still work. A test covers both access styles.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: Thanks. This logging was added after a timeout gave no indication of which peer had stalled. The new messages identify the peer and retry attempt.

KavinKrishnan:
@coderabbitai review

modelexpress_client/python/modelexpress/envs.py
```diff
@@ -148,6 +151,27 @@ def _env_float(name: str, default: float) -> float:
...
+    value = _env_float(name, default)
+    if value <= 0.0:
+        logger.warning(
+            "Invalid %s=%r; a bound must be positive, using default %s",
+            name,
+            os.environ.get(name),
+            default,
+        )
+        return default
+    return value
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `15b8654`. The validation now rejects non-finite values as well as zero and negative values. Otherwise `inf` or `nan` could prevent the deadline from expiring. Tests cover `inf`, `-inf`, and `nan`.

KavinKrishnan:
@nv-hwoo Yes, you read it correctly — every receiver rank was dialing every trainer rank. Good catch, and your suggested fix is the right one. Done in a646037.

The handshake now runs after `plan_transfer()` instead of right after discovery, and dials only the trainers the plan reads from. `TransferPlan.sessions()` reports the sessions the plan actually pulls from, and `handshake_endpoints_for_plan()` narrows the endpoint map to those agents.

To be clear about what could and could not move: discovery still has to collect every trainer's shard table, because which trainers hold the bytes a given receiver is missing is only known after the slice arithmetic. The handshake had no such constraint — it only needs to happen before the first read — so it was simply in the wrong place.

Two things fell out of the reorder. An unsupported plan is now rejected before any handshake budget is spent, which is separately what CodeRabbit flagged. And if the plan reads from a trainer that published no endpoint, we now fail with that trainer named, instead of letting `prep_xfer_dlist` fail later without being able to say which peer's metadata was missing.

One caveat I want to state rather than let the change imply otherwise: on Topology A the handshake is a 3.2 s median across 16 peers, about 4% of an 87 s cold start, and `_prepare` runs once per run because the plan is cached and reused every step. Capture and plan are 51 s of that 87 s. So this is scaling headroom for larger rank counts and the removal of work that was never needed — not a win you will see in the current numbers.

Tests cover selection across all three plan phases (direct, dtype-conversion staging, full-pull staging — missing one would silently drop a peer), address preservation, the fail-closed missing-endpoint case, and an empty plan dialing nobody.

While in the same function I also fixed the other CodeRabbit finding about the dial timeout: `max(1.0, min(attempt_timeout, remaining))` rounded a sub-second remainder up to a full second, so the last dial could overrun the total budget. It now uses the remainder directly, with a test for the sub-second case.

nv-hwoo (approved):

PR #551, opened 2026-07-27
feat(reshard-refit): measure refit coverage and gate on it

KavinKrishnan:
## Summary

This PR adds a machine-readable `refit-coverage-v1` record that measures what a
receiver intends to install against the engine's complete parameter layout. It
also adds an opt-in coverage floor for benchmark and full-refit runs.

The PR now targets current `main`; #528 is already merged and its obsolete branch
commit has been removed.

## Why this is needed

Digest checks can validate bytes only for tensors the receiver requests. If a
publisher omits half the model, those tensors never enter the plan and therefore
never reach a digest check. The transfer can report zero mismatches while leaving
half the engine at its previous values.

That failure occurred when pipeline-parallel trainers published local rather
than global layer names. The resulting refit covered 51% of the model and looked
faster because it moved half the bytes and issued half the descriptors.

Coverage closes that observability gap by comparing:

- `param_layout`: the engine's complete refittable parameter layout;
- `all_params`: the parameters this receiver will install;
- `plan`: the actual wire and descriptor plan.

The record is emitted as JSON at WARNING level so ordinary benchmark logs retain
it without enabling INFO logging for every dependency.

## Where the record is emitted

Immediately after the transfer plan is built, and before `_prepare` fails closed
on unsupported sources. Emitting it at the end of `_prepare` would mean the runs
with a known coverage hole are exactly the runs with no coverage record, since
`plan.fallback` raises first. The rejection itself is unchanged: an unsupported
plan still ends `_prepare` with `UnsupportedReshard`.

## Record and gate

The `refit-coverage-v1` record includes parameter and byte coverage, never-written
parameter samples, captured copies, unsupported entries, planned wire bytes,
descriptor counts, full pulls, conversions, and fallback count.

The gate is disabled by default because partial refit is supported behavior.
Enable it with:

```text
MX_RESHARD_REQUIRE_FULL_COVERAGE=1
MX_RESHARD_COVERAGE_FLOOR=0.995
```

Both variables are registered in `modelexpress/envs.py` and documented in the
client README variable table, so they parse and warn like every other variable
and are read live rather than frozen at import.

The default floor is below 1.0 because some engine buffers, such as rotary
`inv_freq`, are legitimately outside the refit set. A floor outside `[0.0, 1.0]`
is rejected rather than honored: a negative floor passes every refit, silently
disabling the gate the caller just asked for, and a floor above 1.0 rejects even
a complete refit.

## What coverage does not prove

Coverage measures plan intent, not byte delivery. A record at 100% does not prove
that transport wrote the destination buffers, that installed values match the
source, or that generation is equivalent.

A benchmark row still requires parameter equality, generation and KL checks,
version agreement, fallback checks, and retained raw evidence. Coverage is one
necessary gate, not a replacement for those checks.

## Validation

- [x] Rebased onto current `main`; the already-merged #528 commit was dropped.
- [x] 53 complete `test_reshard_refit_*` tests pass, 15 of them focused on
      coverage.
- [x] Tests cover complete, partial, unsupported, empty-layout, and non-float
      buffer cases.
- [x] The opt-in floor rejects incomplete full refits, and an out-of-range floor
      is rejected in both directions.
- [x] A lifecycle test drives `_prepare` with a fallback-bearing plan and asserts
      the coverage record is emitted before the rejection.

Full GitHub CI is running on the rebased branch.

## Reviewer guide

1. `modelexpress/refit/reshard/receiver.py` — review `_log_coverage()` first.
   Confirm that coverage uses the engine-provided layout and installed set rather
   than an assumed sharding formula, and that it runs before the fail-closed
   rejection.
2. `tests/test_reshard_refit_coverage.py` — focus on the 51% partial-refit case,
   the opt-in gate, the emission-ordering test, and the non-float-buffer case
   that justifies the 0.995 floor.
3. `modelexpress/envs.py` and `modelexpress_client/python/README.md` — the two
   gate variables and their documented defaults.

The main policy question is whether WARNING is the right retention mechanism.
Changing the record back to INFO without a dedicated benchmark logger would
recreate the evidence-loss problem this PR addresses.

## Summary by CodeRabbit

* **New Features**
  * Added refit coverage reporting with machine-readable statistics and severity-based logging.
  * Added configurable coverage thresholds and optional enforcement for incomplete refits.
  * Added a dedicated error for refits that do not write all required parameter data.

* **Documentation**
  * Documented the new coverage enforcement and threshold environment settings, including the default threshold.

* **Bug Fixes**
  * Improved detection and handling of incomplete or unsupported refit plans.

### DIFF
--- a/modelexpress_client/python/modelexpress/envs.py
+++ b/modelexpress_client/python/modelexpress/envs.py
@@ -59,6 +59,8 @@ if TYPE_CHECKING:
     MX_MODEL_URI: Optional[str]
     MX_P2P_METADATA: str
     MX_RESHARD_FUSED_WIRE: bool
+    MX_RESHARD_REQUIRE_FULL_COVERAGE: bool
+    MX_RESHARD_COVERAGE_FLOOR: float
     # Kubernetes service backend
     MX_K8S_SERVICE_PATTERN: str
     MX_K8S_SOURCE_RETRIES: str
@@ -191,6 +193,15 @@ environment_variables: dict[str, Callable[[], Any]] = {
     "MX_MODEL_URI": lambda: os.environ.get("MX_MODEL_URI"),
     "MX_P2P_METADATA": lambda: os.environ.get("MX_P2P_METADATA", "1"),
     "MX_RESHARD_FUSED_WIRE": lambda: _env_bool("MX_RESHARD_FUSED_WIRE", True),
+    # Refit coverage gate. The floor is a fraction of the engine's parameter
+    # bytes; ReshardReceiver validates its range at the point of use.
+    "MX_RESHARD_REQUIRE_FULL_COVERAGE": lambda: os.environ.get(
+        "MX_RESHARD_REQUIRE_FULL_COVERAGE", ""
+    )
+    .strip()
+    .lower()
+    in _TRUTHY,
+    "MX_RESHARD_COVERAGE_FLOOR": lambda: _env_float("MX_RESHARD_COVERAGE_FLOOR", 0.995),
     # ── Kubernetes service backend ─────────────────────────────────────────
     "MX_K8S_SERVICE_PATTERN": lambda: os.environ.get("MX_K8S_SERVICE_PATTERN", "mx-sources"),
     "MX_K8S_SOURCE_RETRIES": lambda: os.environ.get("MX_K8S_SOURCE_RETRIES", ""),
--- a/modelexpress_client/python/modelexpress/envs.py
+++ b/modelexpress_client/python/modelexpress/envs.py
@@ -197,7 +197,7 @@ environment_variables: dict[str, Callable[[], Any]] = {
     # bytes; ReshardReceiver validates its range at the point of use. What a
     # complete refit scores is engine- and model-specific, so the default is set
     # loose enough to pass any complete refit and still catch a gross hole; see
-    # ReshardReceiver._coverage_floor.
+    # modelexpress.refit.reshard.receiver._coverage_floor.
     "MX_RESHARD_REQUIRE_FULL_COVERAGE": lambda: os.environ.get(
         "MX_RESHARD_REQUIRE_FULL_COVERAGE", ""
     )
--- a/modelexpress_client/python/modelexpress/envs.py
+++ b/modelexpress_client/python/modelexpress/envs.py
@@ -194,7 +194,10 @@ environment_variables: dict[str, Callable[[], Any]] = {
     "MX_P2P_METADATA": lambda: os.environ.get("MX_P2P_METADATA", "1"),
     "MX_RESHARD_FUSED_WIRE": lambda: _env_bool("MX_RESHARD_FUSED_WIRE", True),
     # Refit coverage gate. The floor is a fraction of the engine's parameter
-    # bytes; ReshardReceiver validates its range at the point of use.
+    # bytes; ReshardReceiver validates its range at the point of use. What a
+    # complete refit scores is engine- and model-specific, so the default is set
+    # loose enough to pass any complete refit and still catch a gross hole; see
+    # ReshardReceiver._coverage_floor.
     "MX_RESHARD_REQUIRE_FULL_COVERAGE": lambda: os.environ.get(
         "MX_RESHARD_REQUIRE_FULL_COVERAGE", ""
     )
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -23,7 +23,9 @@ transport, buffers and the router dtype-cast are shared here.
 
 from __future__ import annotations
 
+import json
 import logging
+import os
 
 import torch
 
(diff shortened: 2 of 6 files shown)


modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -40,6 +42,16 @@
...
 
+# Opt-in rather than always-on: partial and subset refit are intended features,
+# and for those a coverage below 1.0 is the point. What is never acceptable is a
+# *benchmark* row measuring an incomplete refit, because its wire volume and
+# timings are then the wrong magnitude and get compared against complete ones.
+_REQUIRE_FULL_COVERAGE = os.environ.get("MX_RESHARD_REQUIRE_FULL_COVERAGE", "0") == "1"
+# Deliberately not 1.0. A few engine parameters - rotary `inv_freq` and similar
+# non-float buffers - are legitimately not refit material, so a complete refit
+# lands slightly under full coverage.
+_COVERAGE_FLOOR = float(os.environ.get("MX_RESHARD_COVERAGE_FLOOR", "0.995"))
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `f6ef2c0`. Both coverage settings now use the shared environment registry instead of import-time constants. The coverage floor is validated when used, and the README documents both settings. Tests cover invalid ranges and changing the setting within a process.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -302,6 +314,81 @@ def _prepare(self, timeout: float) -> None:
             len(plan.unbounded_sources),
             len(plan.fallback),
         )
+        self._log_coverage(capture, param_layout, all_params, plan)
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `f6ef2c0`. Coverage is now recorded before an unsupported plan raises, so failed refits still explain what was missing. A regression test verifies the record is emitted before the exception.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -303,6 +328,85 @@ def _prepare(self, timeout: float) -> None:
...
+            "planned_wire_bytes": plan.bytes_planned(),
+            "extra_wire_bytes": plan.extra_wire_bytes(),
+            "descriptors": plan.descriptor_count(),
+            "descriptor_savings": plan.descriptor_savings(),
+            "full_pull_sources": len(plan.full_pulls),
+            "unbounded_sources": len(plan.unbounded_sources),
+            "converts": len(plan.converts),
+            "fallback": len(plan.fallback),
+        }
+        logger.warning("MX_REFIT_COVERAGE %s", json.dumps(record))
```
tanushriya910: nit: still feels a bit miscategorized as a warning and overloading the warning severity

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -303,6 +328,85 @@ def _prepare(self, timeout: float) -> None:
...
+
+        # Opt-in rather than always-on: partial and subset refit are intended
+        # features, and for those a coverage below 1.0 is the point. What is never
+        # acceptable is a *benchmark* row measuring an incomplete refit, because
+        # its wire volume and timings are then the wrong magnitude and get
+        # compared against complete ones.
+        if envs.MX_RESHARD_REQUIRE_FULL_COVERAGE and coverage < _coverage_floor():
+            raise RuntimeError(
```
tanushriya910: for consistency shouldn't this be UnsupportedReshard error

modelexpress_client/python/modelexpress/envs.py
```diff
@@ -174,6 +176,15 @@ def _env_float(name: str, default: float) -> float:
...
     "MX_P2P_METADATA": lambda: os.environ.get("MX_P2P_METADATA", "1"),
+    # Refit coverage gate. The floor is a fraction of the engine's parameter
+    # bytes; ReshardReceiver validates its range at the point of use.
+    "MX_RESHARD_REQUIRE_FULL_COVERAGE": lambda: os.environ.get(
+        "MX_RESHARD_REQUIRE_FULL_COVERAGE", ""
+    )
+    .strip()
+    .lower()
+    in _TRUTHY,
+    "MX_RESHARD_COVERAGE_FLOOR": lambda: _env_float("MX_RESHARD_COVERAGE_FLOOR", 0.995),
```
tanushriya910, line 207: would this number vary from model to model?

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: Fixed in `837b0cb`. A complete refit now logs coverage at INFO. WARNING is used only when parameters were not written, a source was unsupported, or fallback occurred. Tests cover the clean and warning cases.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: I used a separate `IncompleteRefit` error in `837b0cb`. `UnsupportedReshard` means a source layout cannot be handled and is caught per source during planning. Reusing it for incomplete model coverage could let callers accidentally ignore a partial refit. Both errors are still `RuntimeError` subclasses, so existing broad handlers continue to work.

modelexpress_client/python/modelexpress/envs.py
KavinKrishnan: Yes. The exact result depends on which load-time parameters the model and engine expose, and some derived buffers are not refit inputs. The default is intentionally loose enough to allow that small difference while still catching large gaps such as missing layers or experts. I documented this in `837b0cb`; deployments can tighten the floor after collecting coverage data for their model.

KavinKrishnan:
@coderabbitai review

modelexpress_client/python/modelexpress/envs.py
```diff
@@ -174,6 +176,18 @@ def _env_float(name: str, default: float) -> float:
...
+    # complete refit scores is engine- and model-specific, so the default is set
+    # loose enough to pass any complete refit and still catch a gross hole; see
+    # ReshardReceiver._coverage_floor.
+    "MX_RESHARD_REQUIRE_FULL_COVERAGE": lambda: os.environ.get(
+        "MX_RESHARD_REQUIRE_FULL_COVERAGE", ""
+    )
+    .strip()
+    .lower()
+    in _TRUTHY,
+    "MX_RESHARD_COVERAGE_FLOOR": lambda: _env_float("MX_RESHARD_COVERAGE_FLOOR", 0.995),
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `f99ef2d`. The comment now points to the module-level `modelexpress.refit.reshard.receiver._coverage_floor` function.

tanushriya910 (approved):

KavinKrishnan:
/ok to test [db12821](https://github.com/ai-dynamo/modelexpress/pull/551/commits/db1282195c2f89fa29f970937a072c8cc17ebc62)

PR #552, opened 2026-07-27
feat(reshard-refit): Megatron target lowering and zero-copy HF aliasing

KavinKrishnan:
## Summary

This PR maps native Megatron tensor geometry into the existing reshard transfer
planner and publishes zero-copy Hugging Face aliases for fused Megatron storage.
It allows an inference receiver to request canonical tensor names without making
the trainer unfuse and copy the model first.

The PR now targets current `main`; #536 is already merged and its former base
commits have been removed.

## Why this is needed

Inference loaders request Hugging Face-style tensor names. Megatron stores many
weights in rank-local fused representations:

- query, key, and value projections share one QKV tensor;
- gate and up projections share one gated-MLP tensor;
- expert layouts have their own tensor-parallel shard axes.

Copying and unfusing those tensors before every publication would add a model-size
allocation and copy on the trainer. Instead, this PR describes canonical aliases
as offsets into storage that is already registered for transfer.

### Target lowering

`lower_megatron_target()` converts explicit Megatron roles into the planner's
`CaptureResult` and destination layout contracts. Supported roles are closed and
validated: column-like, row-like, expert, vocabulary-parallel, and replicated.
Unknown roles raise instead of defaulting to a plausible but incorrect shard
axis.

The lowering uses the receiving tensor-parallel rank's destination slice. Name
translation remains after transfer, while planning still moves only the bytes
owned by that destination rank.

### Zero-copy aliases

`build_hf_aliases()` exposes fused QKV and gated-MLP storage under canonical names
by calculating addresses and shapes within the original registered tensor. It
does not allocate, copy, or register a second buffer.

Alias construction fails closed when a role/name combination has no supported
mapping. A valid-looking partial alias set would be more dangerous than an
explicit error because it could install wrong bytes from a valid address.

For the same reason, the fused gate/up split requires the publisher to state its
storage order as `extras["gated_mlp_order"] == "gate_then_up"`. The two halves are
assigned to `hf_names` positionally, so an up-then-gate layout would publish the
gate projection's bytes under the up projection's name, and no digest check can
see that: each name receives exactly the bytes its publisher advertised for it.
Megatron stores `linear_fc1` gate-then-up, and there is no verified up-then-gate
publisher to test a swap against, so any other value - including an absent one -
is rejected rather than assumed.

### Publication and receiver

`publish_megatron_reshard_view()` packages aliases into the existing rendezvous
format. The caller supplies the `MxReshardRendezvous` and owns its lifetime:
publishing starts the source's READY heartbeat thread, so constructing a
rendezvous inside the publish seam would leave that thread running with no handle
to stop it and no way to mark the source stale on shutdown. Duplicate publish
spec names are rejected, since collapsing them would publish one spec's shard
description under a name the other spec owns and the missing/extra check compares
key sets and cannot see it.

That rendezvous ownership change is worth more of a reviewer's attention than its
size suggests, because there is now hardware evidence for what the leak costs. On a
pre-#536 tree carrying the older bespoke heartbeat, which re-published the entire
blob on every beat rather than only refreshing status, creating a rendezvous per
refit accumulated threads that kept re-asserting an earlier step's shard table.
Two runs aborted reporting corrupt weights. The delivered bytes were correct in
both: the receiver was comparing them against digests from a table a step behind,
which a background thread had put back.

On current `main` that exact mechanism is already gone, since #536 replaced the
bespoke heartbeat with `PublisherThread`, which only sends `UpdateStatus` and never
re-publishes the table. What survives on `main` is narrower and still worth
closing: several threads advertising READY for one source means stopping one marks
it stale and another's next tick puts it back. Caller-owned lifetime removes the
accumulation that makes either failure reachable.

`MegatronReshardReceiver` supplies native target capture and framework translation
while reusing discovery, planning, transport, and buffer management from
`ReshardReceiver`.

## Scope and limits

This PR establishes the Megatron publish and target-lowering path. It does not:

- add digest production or destination verification;
- prove end-to-end refit correctness on a real model;
- canonicalize collective participant ordering for NCCL M2N;
- infer roles from framework parameter names;
- change transport or installation performance.

The tests use synthetic tensor geometry. They validate lowering and alias address
arithmetic, but they are not end-to-end model evidence.

## Validation

- [x] Rebased onto current `main`; merged #536 base commits were dropped.
- [x] 51 reshard tests pass, including 13 focused Megatron tests.
- [x] Tests cover destination TP partitioning, unsupported roles, fused QKV,
      gated MLP, expert aliases, publication, and receiver integration.
- [x] Tests cover the rejected gate/up orders, duplicate publish spec names, and
      the heartbeat being stoppable by the rendezvous owner.

Full GitHub CI is running on the rebased branch.

## Reviewer guide

1. `refit/reshard/megatron.py` — review role-to-axis selection and destination TP
   partitioning first.
2. `refit/reshard/megatron_aliases.py` — verify QKV, gated-MLP, and expert offset
   arithmetic remains inside registered storage, and the required storage-order
   metadata.
3. `refit/reshard/megatron_publisher.py` — verify aliases are published with
   stable names and the original registered addresses, and that rendezvous
   ownership stays with the caller.
4. `refit/reshard/megatron_receiver.py` — confirm the subclass changes only
   Megatron-specific capture and translation boundaries.
5. `tests/test_reshard_megatron.py` — use the alias-offset tests as the reference
   for the intended layouts.
6. `refit/reshard/__init__.py` — compatibility exports only.

Publish ordering is sufficient for independent NIXL reads. Canonical ordering
across collective participants remains a separate prerequisite for NCCL M2N.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
@@ -75,11 +75,25 @@ def _one_shard(
     )
 
 
+GATE_THEN_UP = "gate_then_up"
+
+
 def _build_gated_aliases(
     item: MegatronAliasInput, agent_name: str
 ) -> list[PublishedTensor]:
     if len(item.hf_names) != 2:
         raise ValueError(f"{item.name}: gated tensor requires gate/up HF names")
+    # The halves are assigned to hf_names positionally, so the storage order is
+    # what decides which HF tensor each half becomes. Getting it wrong publishes
+    # the gate projection's bytes under the up projection's name, which no digest
+    # gate can see: both names receive the bytes their publisher advertised. The
+    # order is therefore required rather than assumed.
+    order = item.extras.get("gated_mlp_order")
+    if order != GATE_THEN_UP:
+        raise ValueError(
+            f"{item.name}: fused gate/up aliasing requires extras"
+            f"['gated_mlp_order'] == {GATE_THEN_UP!r}, got {order!r}"
+        )
     axis = int(item.shard_axis if item.shard_axis is not None else 0)
     local_extent = int(item.tensor.shape[axis])
     if local_extent % 2:
--- a/modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Expose native Megatron storage as HF-canonical #496 source shards."""
+"""Expose native Megatron storage as HF-canonical reshard source shards."""
 
 from __future__ import annotations
 
@@ -31,6 +31,15 @@ def _source_rank_and_size(item: MegatronAliasInput, axis: int) -> tuple[int, int
     if item.local_shard_range is None:
         raise ValueError(f"{item.name}: SHARD has no local range")
     lo, hi = (int(value) for value in item.local_shard_range)
+    # A range can pass every check below and still lie outside the tensor it
+    # claims part of: (16, 24) against a global extent of 16 has the right width,
+    # divides evenly, and yields source rank 2 of a 2-rank group. The alias that
+    # follows would then address bytes the full tensor does not have.
+    if not 0 <= lo < hi <= global_extent:
+        raise ValueError(
+            f"{item.name}: source shard range {(lo, hi)} is outside the global "
+            f"extent {global_extent} on axis {axis}"
+        )
     if hi - lo != local_extent or global_extent % local_extent:
         raise ValueError(f"{item.name}: inconsistent source shard geometry")
     if lo % local_extent:
@@ -187,6 +196,16 @@ def build_hf_aliases(
 
     aliases = []
     for item in items:
+        # Every alias published below is a base address plus a shape, which tells
+        # a reader the bytes run contiguously from that address. A strided view
+        # satisfies neither, and nothing downstream can detect it: the read simply
+        # lands on whatever sits between the elements the view meant to select.
+        if not item.tensor.is_contiguous():
+            raise ValueError(
+                f"{item.name}: aliasing publishes an address and a shape, which "
+                f"requires contiguous storage; got a non-contiguous tensor of "
+                f"shape {tuple(int(dim) for dim in item.tensor.shape)}"
+            )
         if item.role == "qkv_column":
             aliases.extend(_build_qkv_aliases(item, agent_name))
             continue
--- a/modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
@@ -1,0 +1,210 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Expose native Megatron storage as HF-canonical #496 source shards."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+from modelexpress.refit.reshard.rendezvous import PublishedShard, PublishedTensor
+
+
+@dataclass(frozen=True)
+class MegatronAliasInput:
+    name: str
+    tensor: Any
+    role: str
+    hf_names: tuple[str, ...]
+    global_shape: tuple[int, ...]
+    placement_kind: str
+    shard_axis: int | None
+    local_shard_range: tuple[int, int] | None
+    extras: dict[str, str] = field(default_factory=dict)
+
+
+def _source_rank_and_size(item: MegatronAliasInput, axis: int) -> tuple[int, int]:
+    local_extent = int(item.tensor.shape[axis])
+    global_extent = int(item.global_shape[axis])
+    if item.placement_kind != "SHARD":
+        return 0, 1
+    if item.local_shard_range is None:
+        raise ValueError(f"{item.name}: SHARD has no local range")
+    lo, hi = (int(value) for value in item.local_shard_range)
+    if hi - lo != local_extent or global_extent % local_extent:
+        raise ValueError(f"{item.name}: inconsistent source shard geometry")
+    if lo % local_extent:
+        raise ValueError(f"{item.name}: non-uniform source shard is unsupported")
+    return lo // local_extent, global_extent // local_extent
+
+
+def _one_shard(
+    *,
+    name: str,
+    tensor: Any,
+    full_shape: tuple[int, ...],
+    agent_name: str,
+    shard_axis: int | None,
+    shard_range: tuple[int, int] | None,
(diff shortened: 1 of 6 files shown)


modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
```diff
@@ -0,0 +1,210 @@
...
+        _one_shard(
+            name=hf_name,
+            tensor=tensor,
+            full_shape=tuple(int(dim) for dim in full_shape),
+            agent_name=agent_name,
+            shard_axis=axis if shard_range is not None else None,
+            shard_range=shard_range,
+        )
+        for hf_name, tensor in zip(item.hf_names, (gate, up), strict=True)
+    ]
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `e1c6249`. We now require `gated_mlp_order="gate_then_up"` and reject missing or unsupported orders. This is the only layout we have verified, so failing clearly is safer than silently assigning gate and up weights incorrectly. Tests cover both a wrong value and missing metadata.

modelexpress_client/python/modelexpress/refit/reshard/megatron_publisher.py
```diff
@@ -0,0 +1,176 @@
...
+            "metadata_endpoint must be an explicit host:port reachable by receivers"
+        )
+    by_name = {spec.name: spec for spec in specs}
+    missing = sorted(set(by_name).difference(tensors))
+    extra = sorted(set(tensors).difference(by_name))
+    if missing or extra:
+        raise ValueError(
+            f"Megatron shard table/tensor mismatch: missing={missing[:10]} "
+            f"extra={extra[:10]}"
+        )
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in `e1c6249`. Duplicate tensor names are now rejected before building the lookup table. A regression test covers the case.

KavinKrishnan:
One caller-facing API changed during review: `publish_megatron_reshard_view()` and `publish_registered_shard_table()` now take a caller-owned `MxReshardRendezvous`.

The caller should create it, pass it to the publish function, and call `rendezvous.close()` during shutdown. This keeps the heartbeat thread tied to the same owner as the NIXL manager and prevents repeated publication from leaving extra heartbeat threads running.

modelexpress_client/python/modelexpress/refit/reshard/megatron_receiver.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""#496 receiver seam for native Megatron staging and translation."""
```
nv-hwoo: is this PR number?

modelexpress_client/python/modelexpress/refit/reshard/megatron_publisher.py
```diff
@@ -0,0 +1,174 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Publish registered native Megatron tensors through #496 rendezvous."""
```
nv-hwoo: for all docstrings and comments referring to #496, can we make agent to describe it in english, instead of leaving it as PR number

modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def _source_rank_and_size(item: MegatronAliasInput, axis: int) -> tuple[int, int]:
+    local_extent = int(item.tensor.shape[axis])
+    global_extent = int(item.global_shape[axis])
+    if item.placement_kind != "SHARD":
+        return 0, 1
+    if item.local_shard_range is None:
+        raise ValueError(f"{item.name}: SHARD has no local range")
+    lo, hi = (int(value) for value in item.local_shard_range)
```
nv-hwoo, line 33: nit: add check for `0 <= lo < hi <= global_extent`. For example, a local extent of 8, global extent of 16, and range (16, 24) passes every existing check. It then produces an alias outside the advertised full tensor.

modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
```diff
@@ -0,0 +1,224 @@
...
+            (item.hf_names[1], kv_heads_local * source_size * head_dim, k_shards),
+            (item.hf_names[2], kv_heads_local * source_size * head_dim, v_shards),
+        )
+    ]
+
+
+def build_hf_aliases(
+    items: list[MegatronAliasInput], *, agent_name: str
+) -> list[PublishedTensor]:
+    """Build zero-copy HF aliases whose addresses remain in registered storage."""
```
nv-hwoo, line 195: The function accepts any tensor for QKV and ordinary aliases and as a result non-contiguous view could be accepted and advertised as contiguous, causing NIXL to read the wrong bytes. 

Reject non-contiguous inputs before dispatching, or explicitly break them into contiguous address ranges. Add QKV and ordinary-alias rejection tests.

modelexpress_client/python/modelexpress/refit/reshard/megatron_receiver.py
KavinKrishnan: Yes, it was — #496 is the PR where the reshard path landed. Removed in 83395ca. This file now says "Reshard receiver seam" and "the generic reshard receiver".

modelexpress_client/python/modelexpress/refit/reshard/megatron_publisher.py
KavinKrishnan: Doh good catch

modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
KavinKrishnan: Good catch, fixed in 83395ca.

Your example is exactly right, and worth spelling out because it shows why none of the existing checks caught it: (16, 24) against a local extent of 8 and a global extent of 16 has the correct width, the global extent divides evenly by the local one, and the offset is a whole multiple of the local extent. So it passed all three and came back as source rank 2 of a 2-rank group, and the alias then described bytes past the end of the tensor.

`_source_rank_and_size` now requires `0 <= lo < hi <= global_extent` before the other checks, with a test using your exact numbers.

modelexpress_client/python/modelexpress/refit/reshard/megatron_aliases.py
KavinKrishnan: Agreed, fixed in 83395ca.

Only the fused gate/up path checked contiguity, and even there it checked the halves it had just narrowed rather than the tensor it was handed, so a strided input reached the QKV and single-name paths unchecked. Since every alias is published as one base address plus a shape, a strided view advertises the bytes sitting between the elements it meant to select, and nothing downstream can see it — the digest is computed over whatever the publisher advertised.

The check now runs in `build_hf_aliases` before dispatch, so all three paths are covered.

I went with rejecting rather than breaking the input into contiguous ranges, because we have no publisher that produces strided tensors here; if one appears, splitting is the better answer and this fails loudly rather than silently in the meantime.

Tests added for both cases you named: a column-sliced QKV tensor and a transposed ordinary alias.

nv-hwoo (approved):

PR #553, opened 2026-07-27
feat(reshard-refit): record refit stages and refuse impossible wire rates

KavinKrishnan:
## Why this is needed

A refit currently returns one total duration. When that number changes, reviewers and benchmark owners cannot tell whether the time moved in discovery, peer setup, transfer, conversion, or installation.

There is also a more serious failure mode: a transfer can report an impossibly short wire time even when the destination buffers were not updated. One observed run claimed 40.61 GB in 0.84 seconds, or 386.9 Gbps, on a receiver with roughly 190.8 Gbps of usable network capacity. Most existing health signals accepted that run, so it would have appeared as the best benchmark result.

This PR makes refit measurements explainable and rejects rates the configured network cannot physically produce.

## What changes

1. Each refit can emit one structured `MX_REFIT_STAGE` record (`refit-stage-v2`). It includes the emitting rank, preparation stages, wire stages, local tensor work, installation time, bytes moved, descriptor counts, and total accounted time.
2. `MX_RESHARD_MAX_GBPS` sets an optional per-rank throughput ceiling. If the measured wire rate exceeds it, the receiver emits `MX_REFIT_IMPOSSIBLE_THROUGHPUT` and stops before untrusted receive buffers are installed into live model parameters.
3. Stages that do not run are omitted. For example, an all-convert or all-full-pull plan does not report an empty exact-transfer phase.
4. Invalid or non-finite ceiling values disable the guard with a warning instead of breaking every refit.

The ceiling is disabled by default because the correct limit depends on the GPUs, network adapters, container placement, and rank placement of the specific deployment.

## Where it sits

```mermaid
flowchart TB
    A["<b>1. Discover peers and build plan</b>"] --> B["<b>2. Read weight bytes</b>"]
    B --> C["<b>3. Record wire time and bytes</b>"]
    C --> D{"<b>4. Rate within configured ceiling?</b>"}
    D -->|Yes| E["<b>5. Re-slice and convert</b>"]
    E --> F["<b>6. Install live parameters</b>"]
    F --> G["<b>7. Emit stage record</b>"]
    D -->|No| H["<b>5. Emit failure record and abort</b>"]

    classDef trainer fill:#E3F2FD,stroke:#1565C0,stroke-width:2px,color:#0D47A1,font-size:16px;
    class A,B,C mx;
    class D receiver;
    class E,F,G,H receiver;
```

The important ordering is step 4 before steps 5 and 6. An impossible rate means the transport completion cannot be trusted, so the guard must run before any receive buffer reaches live parameters.

## Stage-record contract

- Preparation costs appear only on the refit that paid for them.
- GPU stages synchronize at stage boundaries so time is charged to the work that launched it.
- `accounted_s` is the sum of non-overlapping measured stages. Callers subtract it from their own end-to-end duration to identify unattributed time.
- Fused transfer reports one wire span. Phased transfer reports only the exact, full-pull, and conversion spans that actually ran.
- Every structured record includes `rank`, because fleet-level analysis depends on the slowest rank rather than an anonymous average.

`MX_REFIT_STAGE_RECORD=0` disables the stage line. It is logged at warning level so benchmark harnesses can capture it without enabling info logs for every imported dependency.

## Scope

Included:

- stage timing and byte accounting;
- rank-aware structured records;
- the configurable throughput check;
- fail-before-install ordering;
- environment-variable documentation and tests.

Not included:

- digest verification timing, which belongs with the receiver-side verification gate;
- install batching or other performance changes;
- automatic selection of a network ceiling.

## How to review

Review the six files in this order:

1. **`modelexpress/refit/reshard/receiver.py` — required review.** Check that each timer brackets the work named by the field. Then verify that the throughput check runs after all wire spans are known but before re-slice, conversion, and installation.
2. **`tests/test_reshard_refit_ceiling.py` — required review.** Check both sides of the safety rule: impossible rates install nothing, while realistic rates still install. The bad and good rates come from observed runs rather than invented thresholds.
3. **`tests/test_reshard_refit_stage_record.py` — contract review.** Check rank attribution, setup attribution, absent stages, and `accounted_s`.
4. **`tests/test_reshard_refit_fused_wire.py` — fixture-only change.** The shared harness now carries a receiver rank for structured records.
5. **`modelexpress/envs.py` and `README.md` — configuration review.** Confirm defaults and handling of invalid, NaN, and infinite values.

Questions for reviewers:

- Does the configured ceiling fail closed at the right point, before live parameters change?
- Are fused and phased wire durations calculated from equivalent work?
- Is `accounted_s` still valid if any of these stages later become concurrent? If concurrency is introduced, the schema must change rather than silently summing overlapping spans.

## Validation

- The full local reshard and NIXL lifecycle selection passed: 158 tests.
- A regression test was run against the old ordering and failed because `_install` had already executed; it passes with the guard moved before installation.

## Benchmark status

The stage fields are the instrumentation used for the existing Topology A and B measurements. The ceiling arithmetic is pinned to both the observed invalid run and a valid Topology B run. Hardware validation of the guard itself remains pending because the bad adapter placement should not be reproduced on a shared qualification run.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 
 import json
 import logging
+import math
 import time
 from collections import deque
 
@@ -58,8 +59,22 @@ def _max_gbps() -> float:
 
     Read at call time so a harness can set it per run. Off unless configured
     because only the operator knows the real per-rank limit for their fabric.
+
+    A non-finite ceiling disables the check rather than being compared against.
+    ``float("nan")`` parses happily, and every comparison against NaN is False, so
+    a fat-fingered value would slip past the "is it configured" test and then fail
+    the "is the rate acceptable" test - turning the guard into a refit that always
+    aborts. Infinity is accepted by the comparison but means the same as off.
     """
-    return envs.MX_RESHARD_MAX_GBPS
+    ceiling = envs.MX_RESHARD_MAX_GBPS
+    if not math.isfinite(ceiling):
+        logger.warning(
+            "MX_RESHARD_MAX_GBPS=%r is not a finite number; the throughput ceiling "
+            "is disabled for this run",
+            ceiling,
+        )
+        return 0.0
+    return ceiling
 
 
 def handshake_endpoints_for_plan(
@@ -724,13 +739,25 @@ class ReshardReceiver:
             self._transport.read(descriptors + full_descriptors + convert_descriptors)
             stages["wire_fused_s"] = time.perf_counter() - _t
         else:
-            _t = time.perf_counter()
-            stats = execute_transfer(
-                self._plan,
-                resolve_param_ptr=lambda name: self._param_ptr[name],
-                transport=self._transport,
-            )
-            stages["wire_exact_s"] = time.perf_counter() - _t
+            # A plan can be all converts or all full pulls. Timing an exact phase
+            # that moved nothing records a duration that is not a wire duration: it
+            # inflates accounted_s, and because the phased implied rate divides the
+            # bytes by the summed wire stages, it drags that rate down and makes the
+            # throughput ceiling less likely to catch a run that deserves it.
+            if self._plan.exact_descriptor_count:
+                _t = time.perf_counter()
+                stats = execute_transfer(
+                    self._plan,
+                    resolve_param_ptr=lambda name: self._param_ptr[name],
+                    transport=self._transport,
+                )
+                stages["wire_exact_s"] = time.perf_counter() - _t
+            else:
+                stats = {
+                    "segments": 0,
+                    "bytes": 0,
+                    "fallback": list(self._plan.fallback),
+                }
             if full_descriptors:
                 _t = time.perf_counter()
                 self._transport.read(full_descriptors)
@@ -822,6 +849,10 @@ class ReshardReceiver:
         if envs.MX_REFIT_STAGE_RECORD:
             record = {
                 "schema": "refit-stage-v2",
+                # Every rank emits its own line, so without this a fleet-wide
+                # capture cannot tell them apart - and the number that matters for
+                # a refit is the slowest rank, not the average of an anonymous pile.
+                "rank": self._global_rank,
                 "step": step,
                 "bytes": stats["bytes"],
                 "segments": stats["segments"],
@@ -895,6 +926,9 @@ class ReshardReceiver:
             return
         detail = {
             "schema": "refit-impossible-throughput-v1",
+            # Which rank saw the impossible rate. One rank's adapters being
+            # misselected is a different diagnosis from the whole pod's.
+            "rank": self._global_rank,
             "step": step,
             "wire_bytes": wire_bytes,
             "wire_s": round(wire_s, 6),
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -25,7 +25,6 @@ from __future__ import annotations
 
 import json
 import logging
-import os
 import time
 from collections import deque
 
@@ -53,10 +52,6 @@ from modelexpress.refit.reshard.types import (
 
 logger = logging.getLogger("modelexpress.refit.reshard.receiver")
 
-# Emit one JSON stage record per refit. On by default: the timings are already
-# computed, and at INFO they were never captured by a benchmark run.
-_STAGE_RECORD = os.environ.get("MX_REFIT_STAGE_RECORD", "1") == "1"
-
 
 def _max_gbps() -> float:
     """Per-rank fabric ceiling in Gbps; 0 disables the check.
@@ -64,10 +59,7 @@ def _max_gbps() -> float:
     Read at call time so a harness can set it per run. Off unless configured
     because only the operator knows the real per-rank limit for their fabric.
     """
-    try:
-        return float(os.environ.get("MX_RESHARD_MAX_GBPS", "0") or 0)
-    except ValueError:
-        return 0.0
+    return envs.MX_RESHARD_MAX_GBPS
 
 
 def handshake_endpoints_for_plan(
@@ -820,7 +812,7 @@ class ReshardReceiver:
             len(stats["fallback"]),
         )
         metrics.update({k: round(v, 6) for k, v in stages.items()})
-        if _STAGE_RECORD:
+        if envs.MX_REFIT_STAGE_RECORD:
             record = {
                 "schema": "refit-stage-v2",
                 "step": step,
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -744,7 +744,13 @@ class ReshardReceiver:
             # inflates accounted_s, and because the phased implied rate divides the
             # bytes by the summed wire stages, it drags that rate down and makes the
             # throughput ceiling less likely to catch a run that deserves it.
-            if self._plan.exact_descriptor_count:
+            #
+            # Gated on segments, not exact_descriptor_count. The latter is the
+            # pre-bounding baseline that descriptor_savings() subtracts from, and it
+            # counts every copy including the ones that became full pulls or
+            # converts, so a bounded plan can carry a nonzero count with nothing
+            # left to issue. exact_descriptors() reads segments and nothing else.
+            if self._plan.segments:
                 _t = time.perf_counter()
                 stats = execute_transfer(
                     self._plan,
--- a/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/receiver.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import time
 from collections import deque
 
@@ -51,6 +52,22 @@ from modelexpress.refit.reshard.types import (
 )
 
 logger = logging.getLogger("modelexpress.refit.reshard.receiver")
+
+# Emit one JSON stage record per refit. On by default: the timings are already
+# computed, and at INFO they were never captured by a benchmark run.
+_STAGE_RECORD = os.environ.get("MX_REFIT_STAGE_RECORD", "1") == "1"
+
+
+def _max_gbps() -> float:
+    """Per-rank fabric ceiling in Gbps; 0 disables the check.
+
+    Read at call time so a harness can set it per run. Off unless configured
+    because only the operator knows the real per-rank limit for their fabric.
+    """
+    try:
+        return float(os.environ.get("MX_RESHARD_MAX_GBPS", "0") or 0)
+    except ValueError:
+        return 0.0
 
 
 def handshake_endpoints_for_plan(
@@ -274,7 +291,7 @@ class ReshardReceiver:
         local_rank: int,
         global_rank: int,
         num_trainer_sources: int,
-        device: "torch.device",
+        device: torch.device,
         listen_port: int,
         timeout: float = 1200.0,
     ) -> None:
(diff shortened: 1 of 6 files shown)


KavinKrishnan:
@coderabbitai full review

KavinKrishnan:
Fixed in 5b7e63b. This was a real bug and a good catch.

The guard's whole argument is that an impossible rate means the transport reported completions it did not earn, so the receive buffers cannot be trusted. Running it after `_install` meant it described weights it had already let through, which is the one outcome it exists to prevent. The `RuntimeError` was documenting the corruption instead of stopping it.

The check now runs right after the wire byte accounting, before the re-slice, the dtype cast and the install. Everything it reads is known by then.

One deliberate difference from the suggested diff: the impossible-throughput record is still emitted before the raise, so the evidence survives the abort, but the ordinary stage record is no longer reached on a breach. That is correct rather than a loss — a run that failed this check produced no trustworthy timing worth filing, and emitting one invites it into a benchmark table.

Two tests added through the real `update_weights` rather than against the guard in isolation, since the bug was entirely in where it was called from: a breached ceiling installs nothing, and a rate under the ceiling still installs. The first fails against the previous ordering.

KavinKrishnan:
All four addressed in c0d2325.

**Non-finite ceiling (envs.py).** Real, and the worst of the four in effect. `float("nan")` parses, so the value survived `_env_float`, and then every comparison against NaN is False — it slipped past the "is the guard configured" test and failed the "is this rate acceptable" test, so *every* refit aborted. `_max_gbps` now treats a non-finite value as unconfigured and warns, which matches what an unparseable value already did. Parametrised over `nan`, `NaN`, `inf`, `-inf`.

**Empty exact phase (receiver.py:733).** Correct. A plan can be all converts or all full pulls, and `wire_exact_s` for a phase that moved nothing is not a wire duration. It inflates `accounted_s`, and since the phased implied rate divides bytes by the summed wire stages, it also drags that rate down and makes the ceiling less likely to catch a run that deserves it — the guard is already weakest in phased mode. Now gated on `exact_descriptor_count`, with a phased all-full-pull test asserting `wire_exact_s` is absent and `wire_full_s` present.

**Rank on both records (receiver.py:851).** Agreed, and it should have been there from the start. Every rank writes its own line into one capture, and the number a refit is judged on is the slowest rank, not the average of an anonymous pile. Added to `refit-stage-v2` and `refit-impossible-throughput-v1`. For the impossible-throughput record it also separates one rank's adapters being misselected from the whole pod's, which are different diagnoses.

**Ceiling not cleared in two direct tests.** Correct, and newly load-bearing: now that the guard runs before `_install` rather than after, an inherited ceiling aborts those refits earlier than it used to. Both now clear it, with a comment saying why they cannot rely on the `_run` helper.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -667,47 +735,89 @@ def update_weights(self, step: int, *, timeout: float | None = None) -> dict:
...
-                self._plan,
-                resolve_param_ptr=lambda name: self._param_ptr[name],
-                transport=self._transport,
-            )
+            # A plan can be all converts or all full pulls. Timing an exact phase
+            # that moved nothing records a duration that is not a wire duration: it
+            # inflates accounted_s, and because the phased implied rate divides the
+            # bytes by the summed wire stages, it drags that rate down and makes the
+            # throughput ceiling less likely to catch a run that deserves it.
+            if self._plan.exact_descriptor_count:
```
yafshar:
```suggestion
            if self._plan.segments:
```

`exact_descriptor_count` is the wrong field for this guard — it is the *pre-bounding hypothetical* count, not the count of descriptors this phase will actually issue.

`plan_transfer` increments `exact_descriptor_count` for **every** copy, including ones that subsequently become `full_pulls` or `converts` (`transfer_plan.py:249`, `:265`). That is exactly why `descriptor_savings()` is defined as `exact_descriptor_count - descriptor_count()` — the field is the "what an unbounded plan would have cost" baseline. The descriptors this branch actually issues come from `exact_descriptors(plan, ...)`, which reads `plan.segments` and nothing else.

So on an all-full-pull plan the two diverge. Built one via `plan_transfer` with a dim-0-sharded source over 128 publishers, so the copy exceeds `max_segments_per_copy` and gets bounded:

```
segments               = 0
full_pulls             = 1
converts               = 0
exact_descriptor_count = 128

exact_descriptors(plan) yields: 0 descriptors

with exact_descriptor_count ->  TIMES an empty exact phase
with segments               ->  SKIPS (correct)
```

`execute_transfer` runs with zero descriptors and `wire_exact_s` is recorded for a phase that moved nothing — the precise defect the comment above the branch says it prevents: inflates `accounted_s`, drags down the phased implied rate, makes the ceiling less likely to fire on a run that deserves it.

Same for an all-convert plan: `converts` bumps `exact_descriptor_count` at `transfer_plan.py:249` while leaving `segments` empty.

One knock-on: `test_an_exact_phase_that_moved_nothing_is_not_timed` (`test_reshard_refit_stage_record.py:221`) currently sets *both*

```python
harness._plan.segments = []
harness._plan.exact_descriptor_count = 0
```

No plan `plan_transfer` can produce has that pair — zero segments with a nonzero count is the whole signature of a bounded plan. The test passes today because line 232 masks the bug. Dropping that line makes it fail against the current condition and pass against `segments`, which is what you want from the regression test.

Worth also asserting the ceiling arithmetic downstream of this, since that is where the consequence lands rather than in the record: an all-full-pull phased plan whose `wire_full_s` alone breaches the ceiling should still raise, and does not if a spurious `wire_exact_s` pads the denominator.

KavinKrishnan: You're right. The part I'd missed is that this field has a specific job — it's the "what would this have cost unbounded" baseline that `descriptor_savings()` subtracts from, so by design it counts copies that later became full pulls or converts. Switched the guard to `segments`.

Your point about the test is the more useful half. Setting the count to 0 alongside the empty segments described a plan `plan_transfer` can't actually produce, since zero segments with a nonzero count is precisely the signature of the bounded case. Dropped that line, and the test now asserts the nonzero count as a precondition so it can't drift back. It fails against the old condition.

On the ceiling assertion, I added one for the bounded shape, but I want to be straight about what it does and doesn't show. It pins that a bounded plan still reaches the guard and raises. It does not demonstrate the padding effect, because in this rig the spurious exact span is microseconds against a ceiling low enough to breach either way. The docstring says that rather than implying the test is doing more work than it is. The padding itself is asserted where it's observable, on the stage record.

yafshar (approved):
LGTM!

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -735,4 +851,102 @@ def update_weights(self, step: int, *, timeout: float | None = None) -> dict:
...
+        self, step: int, wire_bytes: int, stages: dict
+    ) -> None:
+        """Refuse a wire rate the fabric cannot physically produce.
+
+        One run delivered nothing and reported the fastest refit yet measured:
+        40.61 GB in 0.84 s, an implied 387 Gbps, on a pod holding two adapters
+        worth about 191 Gbps. Coverage read 100%, fallback and unsupported were 0,
+        and the addresses and digests were stable. The only signal that dissented
+        was the parameter-equality gate, which timing runs switch off by design,
+        so without this check that run becomes the best row in the matrix.
```
tanushriya910, line 901: I see the PR mentions timing runs switch off the equality gate. but in a real RL refit run, wouldn't the Mismatch KL be out of expected bounds in this case? Surprised the max off-policy detection didn't catch this and raise an error there too.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -735,4 +851,102 @@ def update_weights(self, step: int, *, timeout: float | None = None) -> dict:
...
+            "ceiling_gbps": ceiling,
+        }
+        logger.warning("MX_REFIT_IMPOSSIBLE_THROUGHPUT %s", json.dumps(detail))
+        raise RuntimeError(
+            f"[reshard] step {step} moved {wire_bytes} bytes in {wire_s:.4f}s, an "
+            f"implied {implied_gbps:.1f} Gbps against a per-rank ceiling of "
+            f"{ceiling:.1f} Gbps. The fabric cannot deliver that, so the transport "
+            f"reported completions without moving the payload - one known cause is "
+            f"the fabric library selecting adapters this container does not own. "
+            f"Treat this refit as failed, not fast."
```
tanushriya910, line 951: The ceiling requires the operator to know their per-rank bandwidth upfront, feel like that might be a non-trivial ask. Are there ways to detect bad adapter selection or the destination buffers not being written to as expected? Are there any UCX or NIXL level errors we should be relying on instead?

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: Good question, and the answer is that KL almost certainly would *not* have caught it — for a reason that makes this failure worse rather than better.

When the transport reports completions it didn't earn, the receive buffers aren't garbage. They're reused across refits, so they hold the **previous** refit's bytes. The engine then serves a policy that is exactly one step stale, not one that is broken. That's not out-of-distribution, it's mildly off-policy — and async RL is built to tolerate precisely that. Any off-policy bound wide enough to permit legitimate generation lag will permit this too, so the detector isn't failing, it's being handed something inside its tolerance.

The magnitude makes it worse. At the learning rate we run, only 7 of 870 tracked params move at all across five steps in bf16, so the divergence between the step N-1 and step N policies is tiny to begin with. Our own measured Generation KL Error on that rig sat at 0.019–0.051 and we already characterise that as a weak signal. A one-step-stale policy is well inside it.

So the detection story is: it eventually shows up as drift accumulating over many steps, by which point you've burned rollouts, and the real damage is that training still *converges*, just less sample-efficiently. It reads like a hyperparameter problem rather than a transport fault. Meanwhile the timing run that produced it recorded 386.9 Gbps and would have gone into the table as the fastest result on record.

The one case KL would catch loudly is a first refit, where the buffers hold uninitialised memory rather than a previous version. That's the easy version of this bug and it isn't the one that hurts.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
KavinKrishnan: Agreed that it's a non-trivial ask, which is why it's off unless configured — and I'd rather it end up as the last line of defence than the first. Taking the three signals you're pointing at in the order they should fire:

**NIXL/UCX errors — you're right that we should be leaning on these, and the reason we weren't is a bug we've since fixed.** `NixlTransferManager` has a `_data_plane_error` flag feeding `is_healthy()`, but it was only being set on the singular transfer wait, not on the batched READ wait — which is the path reshard refit actually uses. So timeouts and terminal NIXL statuses on the refit path never marked the agent unhealthy. That's #590, now merged. But it only helps when NIXL reports *something*. In the run that motivated this guard, NIXL reported the transfers COMPLETED. There was no error status to rely on; the transport claimed success and delivered nothing.

**Detecting that destination buffers weren't written — that's #591**, and it's strictly better than a throughput heuristic because it names the offending tensor instead of just failing the refit. Once the receiver-side comparison lands, that becomes the primary signal and this ceiling drops to a cheap smoke check.

**Detecting bad adapter selection — this we can do directly rather than infer, and we should.** The known trigger was two receiver replicas sharing one node's four EFAs, which produces silent no-op reads. That's a placement fact, not something to deduce from bandwidth: enumerate the adapters the agent actually bound and assert the count and exclusivity we expected. Deterministic, cheap, and it fires at startup instead of after a bad refit. Not built yet; happy to open it as a follow-up.

Which leaves the ceiling covering one specific residue: the transport affirmatively lying about completion, where there's no error to read, verification isn't enabled, and placement looks fine. That's exactly the run we hit, so I don't want to drop it — but it should be the backstop, not the mechanism.

On the configuration burden, I can remove most of it: rather than asking the operator for a number, derive the default by querying the link rates of the NICs the agent bound and summing them, and treat an explicit `MX_RESHARD_MAX_GBPS` as an override for when that's wrong. Want me to fold that into this PR, or keep this one as-is and do it alongside the adapter assertion?

tanushriya910: Fine with keeping as is for now if we follow up with the adapter check. Once that's in place though, I'd lean toward removing the bandwidth ceiling rather than maintaining both cause using wire rate to signal incorrect transfer is pretty indirect when we can just assert directly that the right adapters are bound.

tanushriya910 (approved):

PR #555, opened 2026-07-27
fix(client): disconnect NIXL peers before a reader exits

KavinKrishnan:
## Problem

A worker that pulled weights over P2P held a loaded remote agent — and a live QP — for the source it pulled from, and exited without invalidating either. That left the source's data plane permanently unusable:

- Every subsequent reader's transfer neither completed nor reported an error, so each stalled for the full 300 s budget and fell back to disk loading.
- The source kept heartbeating `READY`, so new readers kept selecting it.
- The only source-side trace was a single UCX error at the moment the *next* reader connected: `mlx5dv_devx_obj_modify(opcode=0x503) failed ... Remote I/O error`.
- Restarting the source was the only recovery.

Reproduced with a controlled arm: the same source served two consecutive transfers successfully while both readers stayed alive, and only failed for the next reader after one that had transferred was torn down. Source and readers pinned to one node, ruling out topology.

Note that opcode `0x503` is `INIT2RTR_QP` — a *new* QP coming up. The stale peer is the cause, but the casualty is the next connection, which matches the error firing on connect rather than on departure.

## Why the reader has to be the side that cleans up

In P2P the reader drives the metadata exchange: it sends `NIXLCOMM:SEND` and the source replies with its own metadata, so only the reader calls `loadRemoteMD`.

Two consequences, both verified against NIXL:

1. **The source has no peer record to invalidate.** `invalidateRemoteMD(reader)` finds nothing in `remoteSections_` or `remoteBackends_` and returns `NIXL_ERR_NOT_FOUND`. Having the source call `remove_remote_agent(peer)` would be a no-op — and so would the departing reader sending `NIXLCOMM:INVL`.
2. **The source cannot detect the departure.** The comm worker treats a cleanly closed socket as "nothing to read yet" (`recv` returning 0 with nothing buffered), and only a thrown error erases the peer entry. A gracefully terminated peer is indistinguishable from an idle one.

So the RDMA READ creates real connection state on the source that the source can neither track nor invalidate, and nothing but the reader can close the pair. Hence the permanence.

## Changes

**Readers disconnect before exiting** — the fix.

- `NixlTransferManager` tracks the remote agents it loads, via both `fetch_remote_and_wait` (P2P) and `add_remote_agent` (centralized).
- New `remove_remote_agent()`, the missing counterpart. Wraps `invalidateRemoteMD`, which frees cached metadata *and* disconnects the backend. Never raises — `NOT_FOUND` is expected if the peer already left.
- `shutdown()` disconnects **before** dropping `_agent`; once that reference is gone there is no handle left to disconnect through.
- `NixlTransferManager` owns process-wide teardown: it registers its own `atexit` hook and disconnects all tracked peers before dropping `_agent`. Generic publisher stop only marks that publisher STALE and cannot tear down peers shared with receive or artifact paths.

The manager is the right owner because it owns both the local NIXL handle and the complete remote-peer inventory. Centralized receives now route through its tracking wrapper, and manager-owned process teardown closes peers after successful transfers without coupling cleanup to any one publisher.

**A broken data plane stops being advertised.**

- `is_healthy()` now accounts for transfer timeouts and error statuses, cleared by a later success so one transient blip cannot sideline a worker for the life of the process. A timeout counts because a wedged QP produces neither a completion nor an error status.
- An unhealthy worker is **demoted** rather than skipped. Skipping only stopped refreshing `updated_at`, leaving the entry selectable until the reaper timed it out (90 s default), during which every reader that picked it paid a full timeout. Demotion is reversible, and workers that never reached `READY` are still not demoted.

**The transfer budget is configurable.** `MX_TRANSFER_TIMEOUT` now applies to the RDMA receive path, **only when explicitly set**.

That caveat is deliberate: the variable defaults to 900, three times this path's historical 300. Adopting `envs.MX_TRANSFER_TIMEOUT` wholesale would have stalled every unconfigured deployment for 900 s per candidate — up to 45 minutes across three — while claiming to fix a stall. `test_the_900_default_of_the_env_var_is_not_inherited` guards that. Zero and negative values fall back rather than meaning "unbounded", since `receive_from_source` treats `None` as wait-forever.

## Scope — what this does not do

`is_healthy()` cannot catch a **passive source**. Such a source issues no transfers — a reader's one-sided READ is invisible to its Python layer — so there is no outcome for it to observe. `test_health_does_not_claim_to_catch_a_passive_source` pins that limit so it is not mistaken for coverage.

A durable cross-reader guard needs a server-side change, deliberately **not** included here. The signal exists but on the wrong side: the reader knows the transfer timed out and which source it was, and `UpdateStatusRequest` already carries the source's IDs, so a reader could demote a source with no proto change. Two reasons not to:

- A wedged source keeps heartbeating `READY` every 30 s, so the demotion is undone within one interval unless the server records a penalty that survives heartbeats.
- Letting any reader demote any source is a poisoning vector — a reader with a broken local NIC would evict healthy sources.

The shape that works is a server-side strike count fed by reader-reported failures, which the source's own heartbeat cannot clear. Happy to follow up if maintainers want that.

## Two upstream NIXL issues

Both are the reason the source cannot help itself, and neither is fixable from the client:

1. The P2P fetch flow leaves the responder with untracked connection state. Either the responder should record peers it has served, or `invalidateRemoteMD` should reach connection state it has no metadata for.
2. The comm worker cannot detect graceful peer disconnect; EOF should close and erase the peer entry rather than being read as "no data".

Also worth a NIXL/UCX answer: why a departed peer's half-open QP makes `INIT2RTR_QP` fail for an *unrelated* new peer. That step is below both MX and NIXL, and if it is a UCX RC resource shared across endpoints, other consumers are exposed to the same failure.

## Tests

Regression coverage includes P2P and centralized peer tracking, manager-owned process teardown, publisher replacement isolation, timeout handling, demotion, and recovery. The focused lifecycle/timeout suite plus existing publisher tests passes: 57 passed.

```
pytest modelexpress_client/python/tests/test_nixl_peer_lifecycle.py       # 18
pytest modelexpress_client/python/tests/test_publisher_peer_teardown.py   #  9
pytest modelexpress_client/python/tests/test_rdma_transfer_timeout.py     # 10
```

The wider suite has 74 pre-existing failures in this environment (CUDA, vLLM, protobuf, HF network) in files this PR does not touch; verified byte-identical against a pristine `main` worktree.

## Verification still needed on hardware

These tests cover the client contract — that a reader disconnects before exiting — **not** the UCX behaviour that contract avoids. Confirming the wedge is gone needs the original InfiniBand rig: run the repro, and after the first reader is scaled to zero, check that the next reader's transfer succeeds and the source logs no `mlx5dv_devx_obj_modify(opcode=0x503)` error. I could not reproduce faithfully on the cluster available to me, which uses EFA/libfabric rather than IB/UCX, so this specific mlx5 QP failure mode does not exist there and a green run would prove nothing.

## Summary by CodeRabbit

* **New Features**
  * Added configurable RDMA transfer timeouts through `MX_TRANSFER_TIMEOUT`, with validation and a 300-second fallback.
  * Improved detection and reporting of data-plane failures, including transfer timeouts and errors.
  * Workers now transition to `STALE` during unhealthy periods and return to `READY` after recovery.
  * Remote peer connections are tracked and cleanly disconnected during shutdown.

* **Documentation**
  * Documented timeout configuration, worker state transitions, and shutdown behavior.

* **Tests**
  * Added coverage for timeout configuration, health reporting, peer lifecycle, demotion, recovery, and shutdown handling.

---

## Known limitation: this teardown depends on a cooperative exit

Reviewers should weigh this explicitly rather than discover it later.

The disconnect runs from the NIXL manager's process-wide `atexit` hook. CPython does not run `atexit` handlers when the default `SIGTERM` disposition terminates the process, and this code lives in the vLLM EngineCore subprocess — so whether it fires on `kubectl scale --replicas=0` depends on the engine turning that signal into a clean interpreter exit.

If maintainers would rather not depend on engine behaviour, the alternative is a chaining `SIGTERM` handler in the client that saves and delegates to any handler already installed. I left it out deliberately: installing signal handlers from a library embedded in someone else's engine process is intrusive, and vLLM may legitimately own that signal. Happy to add it if preferred.

More broadly, **prevention is the most a client-side fix can offer.** A reader killed without running cleanup — `SIGKILL` after the grace period, OOM kill, node loss, `--force` deletion — recreates the same wedge, and the source can neither detect nor recover from it. Those are routine events in Kubernetes, not edge cases. Closing that hole needs the two NIXL changes described above: the source currently has no record of its readers to invalidate, and no signal that one has departed.

### DIFF
--- a/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import logging
+import math
 import os
 import time
 
@@ -75,9 +76,9 @@ def _transfer_timeout_seconds() -> float:
             DEFAULT_RDMA_TRANSFER_TIMEOUT_S,
         )
         return DEFAULT_RDMA_TRANSFER_TIMEOUT_S
-    if configured <= 0:
+    if not math.isfinite(configured) or configured <= 0:
         logger.warning(
-            "MX_TRANSFER_TIMEOUT=%r is not positive; using %.1fs for the RDMA "
+            "MX_TRANSFER_TIMEOUT=%r is not finite and positive; using %.1fs for the RDMA "
             "receive budget rather than waiting indefinitely",
             raw,
             DEFAULT_RDMA_TRANSFER_TIMEOUT_S,
--- a/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import time
 
 from .. import envs
@@ -36,6 +37,53 @@ from .. import p2p_pb2
 logger = logging.getLogger("modelexpress.strategy_rdma")
 
 MAX_SOURCE_RETRIES = 3
+
+# Fallback when MX_TRANSFER_TIMEOUT is unset or unusable. Matches the value this
+# path used when the timeout was hard-coded, so behaviour is unchanged by default.
+DEFAULT_RDMA_TRANSFER_TIMEOUT_S = 300.0
+
+
+def _transfer_timeout_seconds() -> float:
+    """Per-source RDMA receive budget, in seconds.
+
+    Every candidate costs this long when a source is wedged, because a wedged QP
+    produces neither a completion nor an error status - the timeout is the only
+    thing that ends the wait. With MAX_SOURCE_RETRIES candidates the worst case is
+    that multiplied by three before the target falls back to disk, so operators
+    need to be able to size it against their model rather than accept a constant.
+
+    Honours MX_TRANSFER_TIMEOUT only when it is *explicitly set*, and reads
+    os.environ directly to tell that apart from the default. This is deliberate:
+    that variable defaults to 900, so simply adopting ``envs.MX_TRANSFER_TIMEOUT``
+    would triple this path's budget from 300s to 900s and make the stall being
+    fixed three times more expensive for anyone who never configured it.
+
+    A non-positive or unparseable value falls back rather than being honoured. In
+    particular 0 must not become "no timeout": ``receive_from_source`` treats None
+    as wait-forever, which is the failure mode this exists to bound.
+    """
+    raw = os.environ.get("MX_TRANSFER_TIMEOUT")
+    if raw is None:
+        return DEFAULT_RDMA_TRANSFER_TIMEOUT_S
+    try:
+        configured = float(raw)
+    except (TypeError, ValueError):
+        logger.warning(
+            "MX_TRANSFER_TIMEOUT=%r is not a number; using %.1fs for the RDMA "
+            "receive budget",
+            raw,
+            DEFAULT_RDMA_TRANSFER_TIMEOUT_S,
+        )
+        return DEFAULT_RDMA_TRANSFER_TIMEOUT_S
+    if configured <= 0:
+        logger.warning(
+            "MX_TRANSFER_TIMEOUT=%r is not positive; using %.1fs for the RDMA "
+            "receive budget rather than waiting indefinitely",
+            raw,
+            DEFAULT_RDMA_TRANSFER_TIMEOUT_S,
+        )
+        return DEFAULT_RDMA_TRANSFER_TIMEOUT_S
+    return configured
 
 
 class RdmaStrategy(LoadStrategy):
@@ -471,7 +519,7 @@ class RdmaStrategy(LoadStrategy):
             bytes_transferred, tensor_count, _ = ctx.nixl_manager.receive_from_source(
                 source_metadata=source_worker.nixl_metadata,
                 source_tensors=source_tensors,
-                timeout_seconds=300.0,
+                timeout_seconds=_transfer_timeout_seconds(),
                 remote_agent_name=remote_agent_name_override,
                 require_exact_match=require_exact_match,
             )
--- a/modelexpress_client/python/modelexpress/metadata/publisher.py
+++ b/modelexpress_client/python/modelexpress/metadata/publisher.py
@@ -92,6 +92,9 @@ class PublisherThread:
         self._publish_started_at: float | None = None
         self._publish_given_up = False
         self._cleaned_up = False
+        # True once we have demoted this worker for an unhealthy data plane, so the
+        # demotion and its log line happen once rather than every interval.
+        self._unhealthy = False
 
         self._interval = (
             interval_secs
@@ -133,14 +136,55 @@ class PublisherThread:
         if self._thread is not None:
             self._thread.join(timeout=self._interval + 5)
         self._mark_stale()
+        self._disconnect_nixl_peers()
         self._cleanup()
         logger.info(f"[Worker {self._worker_rank}] Publisher thread stopped")
 
     def _on_exit(self) -> None:
-        """atexit handler: mark STALE on clean shutdown (SIGTERM)."""
+        """atexit handler: mark STALE and disconnect NIXL peers on clean shutdown.
+
+        Note on reach: CPython runs atexit handlers on normal interpreter exit and
+        on SystemExit, but *not* when the default SIGTERM disposition terminates
+        the process. A worker killed by a plain SIGTERM therefore runs none of
+        this, so the engine embedding the client must either install a handler
+        that raises SystemExit or call ``stop()`` itself. Both paths land here.
+        """
         self._stop_event.set()
         self._mark_stale()
+        self._disconnect_nixl_peers()
         self._cleanup()
+
+    def _disconnect_nixl_peers(self) -> None:
+        """Disconnect NIXL remote agents this worker loaded, before exiting.
+
+        A worker that pulled weights holds a loaded remote agent - and a live QP -
+        for the source it pulled from. Exiting without invalidating that leaves the
+        source with a half-open connection it cannot clean up on its own, because in
+        P2P the source never loads the reader's metadata and so has no peer record
+        to invalidate. The next reader's connection setup then fails against the
+        source's stale state, and the source stays wedged until it restarts while
+        still heartbeating READY.
+
+        Ordered after ``_mark_stale`` so the registry has already stopped offering
+        this worker before we start tearing its transport down. Best-effort and
+        silent on failure: this runs on exit paths where raising would replace a
+        clean shutdown with a crash.
+        """
+        if self._nixl_manager is None:
+            return
+        try:
+            disconnected = self._nixl_manager.disconnect_remote_agents()
+            if disconnected:
+                logger.info(
+                    f"[Worker {self._worker_rank}] Disconnected "
+                    f"{disconnected} NIXL peer(s) on shutdown"
+                )
+        except Exception:
+            logger.debug(
+                f"[Worker {self._worker_rank}] Failed to disconnect NIXL peers "
+                f"on shutdown",
+                exc_info=True,
+            )
 
     def _mark_stale(self) -> None:
         """Best-effort UpdateStatus(STALE). Swallows all errors."""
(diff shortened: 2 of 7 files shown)


KavinKrishnan:
/ok to test [50c49d5](https://github.com/ai-dynamo/modelexpress/pull/555/commits/50c49d5bc9743755dddd4f75016c3404cfea6e75)

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -37,6 +38,53 @@
...
+    if configured <= 0:
+        logger.warning(
+            "MX_TRANSFER_TIMEOUT=%r is not positive; using %.1fs for the RDMA "
+            "receive budget rather than waiting indefinitely",
+            raw,
+            DEFAULT_RDMA_TRANSFER_TIMEOUT_S,
+        )
+        return DEFAULT_RDMA_TRANSFER_TIMEOUT_S
+    return configured
+
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in eda60a3. `_transfer_timeout_seconds()` now rejects NaN and both infinities with `math.isfinite`, and parametrized tests pin all three values to the 300s fallback.

modelexpress_client/python/modelexpress/metadata/publisher.py
```diff
@@ -133,15 +136,56 @@ def stop(self) -> None:
...
+                    f"[Worker {self._worker_rank}] Disconnected "
+                    f"{disconnected} NIXL peer(s) on shutdown"
+                )
+        except Exception:
+            logger.debug(
+                f"[Worker {self._worker_rank}] Failed to disconnect NIXL peers "
+                f"on shutdown",
+                exc_info=True,
+            )
+
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in eda60a3. `_on_exit()` now performs the same bounded join as `stop()` before marking STALE; the new ordering test asserts join -> STALE -> peer disconnect.

modelexpress_client/python/modelexpress/metadata/publisher.py
```diff
@@ -159,6 +203,40 @@ def _mark_stale(self) -> None:
...
+                    f"[Worker {self._worker_rank}] NIXL agent unhealthy, marked "
+                    f"STALE so targets stop selecting it: {reason}"
+                )
+            except Exception:
+                logger.debug(
+                    f"[Worker {self._worker_rank}] Failed to mark STALE for "
+                    f"unhealthy agent",
+                    exc_info=True,
+                )
+
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in eda60a3. `_unhealthy` now latches only after the STALE RPC succeeds. A regression test makes the first status RPC fail, confirms the latch stays clear, then confirms the next call retries and succeeds.

KavinKrishnan:
Addressed all CodeRabbit findings in `eda60a3`: non-finite timeout rejection and tests; bounded publisher-thread join before STALE; demotion retry after transient RPC failure; corrected SIGTERM/atexit wording; and trimmed the long implementation docstrings while retaining the detailed rationale in regression-test documentation. Focused suite: 73 passed. Pre-commit: clean.

KavinKrishnan:
/ok to test [eda60a3](https://github.com/ai-dynamo/modelexpress/pull/555/commits/eda60a3201db38adec104110f73593ffee90eb6f)

KavinKrishnan:
/ok to test [eda60a3](https://github.com/ai-dynamo/modelexpress/pull/555/commits/eda60a3201db38adec104110f73593ffee90eb6f)

KavinKrishnan:
/ok to test [eda60a3](https://github.com/ai-dynamo/modelexpress/pull/555/commits/eda60a3201db38adec104110f73593ffee90eb6f)

KavinKrishnan:
/ok to test [6319ac2](https://github.com/ai-dynamo/modelexpress/pull/555/commits/6319ac2eb55c3d21735b89f73ac0b497faa397de)

KavinKrishnan:
/ok to test [6319ac2](https://github.com/ai-dynamo/modelexpress/pull/555/commits/6319ac2eb55c3d21735b89f73ac0b497faa397de)

KavinKrishnan:
/ok to test [e7da720](https://github.com/ai-dynamo/modelexpress/pull/555/commits/e7da7203dfc0be9399b073f63223000813bda79e)

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -572,8 +604,51 @@ def add_remote_agent(self, source_metadata: bytes) -> str:
             remote_agent_name,
             len(source_metadata),
         )
+        self._remote_agents.setdefault(remote_agent_name, None)
```
zhengluo-nv, line 617: [P1] Track centralized receives through this wrapper

`receive_from_source()` still calls `self._agent.add_remote_agent(source_metadata)` directly when `remote_agent_name is None`, so the centralized receive path never executes this `setdefault`. Both the centralized `RdmaStrategy` path and `trtllm_live_transfer` can later call `shutdown()`, which then sees no tracked peer and drops `_agent` without invalidating that remote metadata/QP. Please route the call at `receive_from_source():702` through `self.add_remote_agent(source_metadata)` and add a regression test that receives without a preloaded agent name, then verifies shutdown removes it.

modelexpress_client/python/modelexpress/metadata/publisher.py
```diff
@@ -133,15 +137,37 @@ def stop(self) -> None:
         if self._thread is not None:
             self._thread.join(timeout=self._interval + 5)
         self._mark_stale()
+        self._disconnect_nixl_peers()
```
zhengluo-nv: [P2] Do not tear down every shared peer on a generic publisher stop

`PublisherThread` does not own the manager's remote-peer set, and `stop()` is also used for routine artifact/source replacement—not only process shutdown. For example, `artifact_lifecycle.py` calls `previous.stop()` at lines 165 and 261 while the same `ctx.nixl_manager` remains shared by weight serving, other artifact publishers, and receive paths. This call therefore invalidates every peer loaded by unrelated users while the process and manager continue running. Please keep global peer disconnection in manager/process-wide shutdown (or make it explicit for the publisher that actually owns that shutdown), and add a replacement test showing unrelated tracked peers remain connected.

zhengluo-nv (commented):
Two lifecycle gaps remain in the current head. The `_agent` / `_remote_agents` split itself is reasonable: `_agent` is the local NIXL handle, while `_remote_agents` is the Python-side inventory needed for deterministic teardown because the public NIXL API does not expose remote enumeration. The problems are that one production path bypasses that inventory and a generic publisher is allowed to tear down inventory it does not own.

modelexpress_client/python/modelexpress/nixl_transfer.py
KavinKrishnan: Fixed in c51f971. `receive_from_source()` now routes centralized metadata through `self.add_remote_agent()`, so the peer enters the manager-owned inventory. Added a regression that performs a centralized receive and verifies `shutdown()` removes that peer.

modelexpress_client/python/modelexpress/metadata/publisher.py
KavinKrishnan: Fixed in c51f971. Removed peer disconnection from generic `PublisherThread.stop()` and `_on_exit()`. `NixlTransferManager` now registers manager-owned process teardown, while publisher replacement only marks its own source STALE. Added regression coverage that routine publisher stop/exit leaves unrelated tracked peers connected, plus manager atexit registration coverage.

KavinKrishnan:
Addressed Zheng's lifecycle review in 1492261:

- centralized receives now route through the manager's peer-tracking wrapper;
- process-wide peer disconnection is owned by NixlTransferManager shutdown/atexit;
- routine PublisherThread stop or artifact replacement no longer disconnects unrelated shared peers;
- added regressions for centralized teardown, manager atexit ownership, and publisher replacement isolation.

Focused suite: 57 passed. Pre-commit: clean.

/ok to test 14922619f1df6918c859b8a2957bfaebcf66ee63

KavinKrishnan:
/ok to test [2a4a7c0](https://github.com/ai-dynamo/modelexpress/pull/555/commits/2a4a7c02cfe8e4a1a5530fb5613dd181d0c53d23)

zhengluo-nv (approved):

KavinKrishnan:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/4af180c7070b0cc8d72bbed3d0b2c1f028009574

PR #557, opened 2026-07-27
test(ci): add Redis backend P2P coverage and vCluster sync diagnostics

tanushriya910:


nicolasnoble:
I went digging on this tonight since it's blocking everything. Most of what I found is negative results, but one positive one changes where I'd look.

The rollout timeouts track almost perfectly with which AKS node the mx-server pod lands on. Pulling the node column out of the `-o wide` dumps across 137 rollouts in 9 runs, `aks-default-4px8c` was 16 for 16 and every other node in the default pool was 4 of 121 combined. It holds inside a single run, so it isn't cluster state or version. In run 30158450235 two rollouts started a second apart on the same node and one finished at 62s while the other timed out. Then the default pool turned over between the 27th and the 28th, 4px8c went away, and the pass rate dropped to 2 of 36.

I can't tell from logs whether those nodes are unhealthy or just busier. Rollouts on 4px8c finish in about 10 seconds and the one success on a bad node took 62, so there's a latency gradient underneath and 120s may just be too tight when the syncer is behind. Cheap way to tell: bump the readiness timeout to ten minutes on a scratch branch and see whether pods eventually go Ready. Converging around three or four minutes would mean throughput and the timeout is the fix. Never converging would point at the nodes, which is an infra ticket rather than a CI one. OPS-7934 mentions deploy-operator failures traced to a bad AKS node, possibly the same pool.

On things that look ruled out, so you don't spend runs on them: the gate swap in `b02be43` reads the same `.status.conditions[Ready]` that `rollout status` does, and run 30318184057 shows it stuck the full 120s. Relaxing the gate further doesn't help either, `ci/k8s/server/mx-server.yaml` has no `publishNotReadyAddresses` and the clients reach the server over Service DNS, so a not-Ready pod means no endpoints and we'd fail at connect time instead. Cluster recreation is covered too, that same run deleted and rebuilt `mx-ci-cluster` at 00:47:36Z and all 13 P2P jobs still failed. Your always-bootstrap commit is the same one-line change Zheng made in `b98192cc` on #540 a few hours earlier.

One thing to double check in the branch: with the CLI at v0.36.0 the delete step on `10afc51e` produced no output in half a second and create then logged `Upgrade vcluster`, so 0.36.0 may not recognize a cluster made by 0.35 and the delete silently no-ops.

If Zheng's readiness probe gets revived, adding `NODE:.spec.nodeName` to the custom-columns would answer the node question directly. It currently prints readiness without node, which is why this was easy to miss.

PR #559, opened 2026-07-28
fix(ci): bump vCluster control plane to 1.34 to match AKS host schema

tanushriya910:
## Summary

CI has been failing broadly since ~July 24 with `kubectl rollout status deployment/mx-server` times out even though the server comes up in seconds, hit-or-miss across tests.

**Root cause:** the AKS host cluster auto-upgraded to Kubernetes **1.34**, which added the `observedGeneration` field on pod conditions. The vCluster control plane was on **1.33**, so the vCluster API dropped that field on sync; the syncer saw the dropped field as a new change and entered an infinite reconcile loop, continuously overwriting `Ready=True` with stale data. As a result pods never surfaced as Ready in the vCluster API, and everything derived from it stalled.

Reference: https://www.vcluster.com/docs/vcluster/troubleshoot/pod-stuck-ready-false-version-skew

## Fix

- **Bump the vCluster control plane to `v1.34.10`** so its API shares the `observedGeneration` schema with the 1.34 host. This is the load-bearing change.
- **Bump the vCluster CLI to `v0.36.0`** (supports 1.34 hosts).
- **Pull CoreDNS from `registry.k8s.io`** instead of the `dynamoci.azurecr.io` mirror, which 401s for the 1.34-era CoreDNS image on a fresh cluster.
- **Print the host Kubernetes version** in vCluster diagnostics so a future host auto-upgrade skew is caught immediately.

## Validation

On a run with this fix, `Ensure vCluster` plus all P2P tests (vllm, sglang, trt-llm, sglang-mooncake, full-manifest same/mixed-version) and both Dynamo tests pass — the entire cascade of failures is resolved by the version match.

## Summary by CodeRabbit

* **New Features**
  * Diagnostics now include the host cluster’s Kubernetes version.
  * vCluster setup uses Kubernetes version 1.34.10 by default and pins CoreDNS to version 1.12.0.
  * The default vCluster CLI version is now v0.36.0.

### DIFF
--- a/.github/actions/install-vcluster-cli/action.yml
+++ b/.github/actions/install-vcluster-cli/action.yml
@@ -8,7 +8,7 @@ inputs:
   vcluster_version:
     description: 'vCluster CLI version to install'
     required: false
-    default: 'v0.33.0'
+    default: 'v0.36.0'
 
 runs:
   using: "composite"
--- a/.github/actions/setup-mx-vcluster/action.yml
+++ b/.github/actions/setup-mx-vcluster/action.yml
@@ -31,9 +31,16 @@ inputs:
     required: false
     default: 'mx-ci'
   vcluster_k8s_version:
-    description: 'Kubernetes version for the vCluster control plane'
+    description: |
+      Kubernetes version for the vCluster control plane. Must be >= the host
+      cluster's minor version. Kubernetes 1.34 added the observedGeneration
+      field on pod conditions; if the host is 1.34+ but the vCluster is 1.33 or
+      older, the vCluster API drops that field and the syncer enters an infinite
+      reconcile loop that keeps overwriting Ready=True with stale data, so pods
+      never surface as Ready (Deployment availableReplicas and Service endpoints
+      stay stuck). Keep this at or above the host minor to match schemas.
     required: false
-    default: 'v1.32.13'
+    default: 'v1.34.10'
   dynamo_ref:
     description: |
       Git ref of the ai-dynamo/dynamo repo to clone for the helm chart.
@@ -143,7 +150,8 @@ runs:
           --set 'controlPlane.statefulSet.scheduling.nodeSelector.kubernetes\.azure\.com/agentpool=system' \
           --set 'controlPlane.statefulSet.scheduling.tolerations[0].key=CriticalAddonsOnly' \
           --set 'controlPlane.statefulSet.scheduling.tolerations[0].operator=Exists' \
-          --set 'controlPlane.statefulSet.scheduling.tolerations[0].effect=NoSchedule'
+          --set 'controlPlane.statefulSet.scheduling.tolerations[0].effect=NoSchedule' \
+          --set controlPlane.coredns.deployment.image=registry.k8s.io/coredns/coredns:v1.12.0
         echo "::endgroup::"
 
     - name: Wait for vCluster pod to be ready
--- a/.github/actions/dump-vcluster-diagnostics/action.yml
+++ b/.github/actions/dump-vcluster-diagnostics/action.yml
@@ -44,6 +44,14 @@ runs:
         # Always read from the host kubeconfig written by connect-vcluster.
         KCTL="kubectl --kubeconfig=${{ github.workspace }}/.kubeconfig-host"
 
+        echo "::group::Host cluster Kubernetes version"
+        # Captures the AKS host server version so we can correlate sudden
+        # vCluster sync regressions with host k8s auto-upgrades. Kubernetes 1.34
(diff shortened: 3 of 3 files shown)


.github/actions/install-vcluster-cli/action.yml
```diff
@@ -8,7 +8,7 @@ inputs:
   vcluster_version:
     description: 'vCluster CLI version to install'
     required: false
-    default: 'v0.33.0'
+    default: 'v0.36.0'
```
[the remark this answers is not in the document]
tanushriya910: Good catch on our self-hosted runners a previously-installed binary does satisfy command -v vcluster, so a default bump can be silently skipped. Deferring it out of this PR, though: this is a validated fix for the host/vCluster version skew that's currently blocking all of CI, and reworking the (currently-working) install path adds untested version-parsing logic I don't want to risk in the same change. For this run the runners picked up v0.36.0 correctly, so the fix is in effect

.github/actions/setup-mx-vcluster/action.yml
```diff
@@ -31,9 +31,16 @@ inputs:
...
+      Kubernetes version for the vCluster control plane. Must be >= the host
+      cluster's minor version. Kubernetes 1.34 added the observedGeneration
+      field on pod conditions; if the host is 1.34+ but the vCluster is 1.33 or
+      older, the vCluster API drops that field and the syncer enters an infinite
+      reconcile loop that keeps overwriting Ready=True with stale data, so pods
+      never surface as Ready (Deployment availableReplicas and Service endpoints
+      stay stuck). Keep this at or above the host minor to match schemas.
     required: false
-    default: 'v1.32.13'
+    default: 'v1.34.10'
```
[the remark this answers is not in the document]
tanushriya910: Agreed on the intent, but the suggested location wouldn't actually catch the failure mode this PR fixes. setup-mx-vcluster runs behind the Bootstrap vCluster gate (if: steps.vcluster-check.outputs.exists != 'true'), so it only executes when the vCluster is being created. A preflight there guards the creation path only.

zhengluo-nv (approved):

.github/actions/setup-mx-vcluster/action.yml
nicolasnoble, line 43: One thought, not blocking: this pins a new constant, and what broke us was the host auto-upgrading past a pinned constant. So I think we're set up to hit this again whenever AKS moves to 1.35.

You already fetch the host version in the diagnostics change here. Could we use it rather than just print it? Either derive the control plane version from the host minor, or keep the pin and have setup fail early with something like "host is 1.35, vcluster pinned at 1.34, bump vcluster_k8s_version". Either way the next one is a red step with the answer in it instead of another few days of digging.

Fine as a follow-up if you'd rather land this as-is, we need CI back more than we need this.

.github/actions/setup-mx-vcluster/action.yml
```diff
@@ -143,7 +150,8 @@ runs:
           --set 'controlPlane.statefulSet.scheduling.nodeSelector.kubernetes\.azure\.com/agentpool=system' \
           --set 'controlPlane.statefulSet.scheduling.tolerations[0].key=CriticalAddonsOnly' \
           --set 'controlPlane.statefulSet.scheduling.tolerations[0].operator=Exists' \
-          --set 'controlPlane.statefulSet.scheduling.tolerations[0].effect=NoSchedule'
+          --set 'controlPlane.statefulSet.scheduling.tolerations[0].effect=NoSchedule' \
+          --set controlPlane.coredns.deployment.image=registry.k8s.io/coredns/coredns:v1.12.0
```
nicolasnoble, line 154: Could we get a short comment here? The reason is in the PR description but whoever bumps the control plane next probably won't have that open, and it isn't obvious from the line that the coredns image is coupled to the k8s version or that we're deliberately going around the ACR mirror.

Separate from this PR, if the mirror 401s on that image that's probably worth filing against whoever owns dynamoci. If we only route around it here it stays broken.

nicolasnoble (approved):
Left two comments, neither blocking. Approving so this can land, having CI back matters more than either of them.

Worth linking DGH-1192 or filing an MX one too, this cost us four days and there's currently nothing that makes it findable from the repo side.

nicolasnoble:
And this got merged while I was putting some feedback on, alright :)

.github/actions/setup-mx-vcluster/action.yml
tanushriya910: you are right, I'll follow up in a different PR

PR #577, opened 2026-07-30
fix(client): release the source peer as part of the load

KavinKrishnan:
Follow-up to #555. The disconnect that PR added runs only from `atexit`, which the vLLM EngineCore process does not reach when the pod is torn down, so it never fired. Release the peer on the load path instead.

## Change

A weight load is one-shot: once the bytes are installed the reader has no further use for the source's QP, so it can be returned while the process is still healthy rather than at interpreter exit. That makes cleanup independent of *how* the process dies, which also covers `SIGKILL` and OOM.

Only the reader can do this. In P2P the reader drives the metadata exchange, so only the reader loads a remote agent; the source has no peer record to invalidate and no signal that a peer departed.

* `rdma_strategy` releases in a `finally`, so a failed transfer cleans up too. That is the wedged-source case, where the reader gives up and moves to the next candidate.
* The centralized branch now loads the peer itself instead of leaving it to `receive_from_source`, so the strategy holds the name it is responsible for releasing. The same asymmetry applies there.
* The release sits after the device sync, pinned by a test, so it cannot tear the QP down under an in-flight READ.

I deliberately did not take the alternative of a `SIGTERM`/`SIGINT` handler: installing signal handlers from a library embedded in someone else's engine process is intrusive, vLLM may legitimately own those signals, and it would still miss `SIGKILL`.

## Not fixed here

If the reader dies *mid-transfer* its `finally` cannot run and the wedge returns. Only the reader can close the pair, so recovering from that needs NIXL-side peer-death detection or source-side invalidation — both described in #555.

MX could separately bound the blast radius by having readers report transfer timeouts and quarantining the source, so later readers stop selecting it. That stops the amplification without repairing the QP. Not in this PR.

Also unchanged: a passive source cannot self-demote. `is_healthy()` reads `_data_plane_error`, which is only ever set by the side that *issues* a transfer, and a P2P source issues none — the reader's one-sided READ is invisible to it.

Related suites: 217 passed. Unit pipeline green.

### DIFF
--- a/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
@@ -477,7 +477,7 @@ class RdmaStrategy(LoadStrategy):
         register_tensors(result, ctx)
 
         is_p2p = bool(source_worker.worker_grpc_endpoint)
-        remote_agent_name_override = None
+        remote_agent_name = None
 
         if is_p2p:
             # _fetch_worker_metadata() prefetched and generation-validated
@@ -506,7 +506,7 @@ class RdmaStrategy(LoadStrategy):
                 f"[Worker {ctx.global_rank}] [TIMING] P2P NIXL metadata fetch: "
                 f"{nixl_fetch_time:.3f}s"
             )
-            remote_agent_name_override = source_worker.agent_name
+            remote_agent_name = source_worker.agent_name
         else:
             source_tensors = [
                 TensorDescriptor(
@@ -518,6 +518,16 @@ class RdmaStrategy(LoadStrategy):
                 )
                 for t in worker_tensor_descriptors(source_worker)
             ]
+            # Load the peer here rather than letting receive_from_source do it,
+            # so this method holds the name it is responsible for releasing.
+            add_start = time.perf_counter()
+            remote_agent_name = ctx.nixl_manager.add_remote_agent(
+                source_worker.nixl_metadata
+            )
+            logger.info(
+                f"[Worker {ctx.global_rank}] [TIMING] add_remote_agent: "
+                f"{time.perf_counter() - add_start:.3f}s (agent={remote_agent_name})"
+            )
 
         logger.info(
             f"[Worker {ctx.global_rank}] Receiving {len(source_tensors)} tensors from source"
@@ -542,11 +552,20 @@ class RdmaStrategy(LoadStrategy):
                 source_metadata=source_worker.nixl_metadata,
                 source_tensors=source_tensors,
                 timeout_seconds=_transfer_timeout_seconds(),
-                remote_agent_name=remote_agent_name_override,
+                remote_agent_name=remote_agent_name,
                 require_exact_match=require_exact_match,
             )
         except Exception as e:
             raise SourceTransferError(f"RDMA receive failed: {e}") from e
+        finally:
+            # Release the source now: this is a one-shot weight load, so the QP
+            # has no further use here, and waiting for process exit is what left
+            # sources wedged. The manager's atexit hook does not run when the
+            # engine process is torn down (SIGTERM, SIGKILL, OOM), and the source
+            # cannot invalidate a reader it never loaded - in P2P only the reader
+            # loads the peer - so nothing else can close the pair. Also runs on
+            # failure, where the source is the more likely casualty.
+            ctx.nixl_manager.remove_remote_agent(remote_agent_name)
         transfer_time = time.perf_counter() - transfer_start
 
         bandwidth_gbps = (
--- a/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
@@ -479,65 +479,75 @@ class RdmaStrategy(LoadStrategy):
         is_p2p = bool(source_worker.worker_grpc_endpoint)
         remote_agent_name = None
 
-        if is_p2p:
-            # _fetch_worker_metadata() prefetched and generation-validated
-            # this manifest before _load_as_target() prepared target tensors.
-            tensor_protos = worker_tensor_descriptors(source_worker)
-            source_tensors = [
-                TensorDescriptor(
-                    name=t.name,
-                    addr=t.addr,
-                    size=t.size,
-                    device_id=t.device_id,
-                    dtype=t.dtype,
-                )
-                for t in tensor_protos
-            ]
-            nixl_fetch_start = time.perf_counter()
-            ep = source_worker.metadata_endpoint
-            host, port_str = ep.rsplit(":", 1)
-            ctx.nixl_manager.fetch_remote_and_wait(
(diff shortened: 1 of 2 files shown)


modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -497,6 +497,16 @@ def _receive_from_peer(
...
+            # Load the peer here rather than letting receive_from_source do it,
+            # so this method holds the name it is responsible for releasing.
+            add_start = time.perf_counter()
+            remote_agent_name = ctx.nixl_manager.add_remote_agent(
+                source_worker.nixl_metadata
+            )
+            logger.info(
+                f"[Worker {ctx.global_rank}] [TIMING] add_remote_agent: "
+                f"{time.perf_counter() - add_start:.3f}s (agent={remote_agent_name})"
+            )
```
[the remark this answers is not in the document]
KavinKrishnan: Restructured — the release is now in an outer `finally` covering acquisition through the device sync, so it is last and a failure before the transfer starts releases the peer too.

One correction on the reasoning, since it matters for anyone reading this later: the QP was not being torn down with device work in flight. `receive_from_source` awaits the transfer to completion and then calls `self._accelerator_backend.synchronize(self._device_id)` before returning, specifically because GPUDirect writes bypass torch streams (`nixl_transfer.py:832-834`). The bytes were already visible at the old release site, and the `ctx.accelerator_backend.synchronize()` that followed is a broader device sync, not the one that makes the transfer visible.

The leak-on-setup-failure half was a real gap, so I made the change anyway. The added value is that the ordering is now explicit here instead of depending on an internal detail of another module — if that internal sync ever moves, `test_the_peer_outlives_the_transfer_and_the_device_sync` fails loudly rather than silently corrupting weights.

modelexpress_client/python/tests/test_rdma_peer_release.py
```diff
@@ -0,0 +1,154 @@
...
+        mgr.remove_remote_agent.assert_called_once_with(P2P_AGENT)
+
+    def test_the_transfer_error_is_not_masked_by_cleanup(self):
+        """Retry and fallback both depend on the original failure propagating."""
+        mgr = _manager()
+        mgr.receive_from_source.side_effect = TimeoutError("Transfer timed out")
+        mgr.remove_remote_agent.side_effect = RuntimeError("invalidate failed")
+
+        with pytest.raises(RuntimeError, match="invalidate failed"):
+            _receive(mgr, p2p=True)
```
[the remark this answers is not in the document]
KavinKrishnan: Good catch, and worse than flagged: the test asserted the cleanup error surfaced, which documented cleanup *masking* the transfer failure — the exact opposite of what the name claimed, against a manager that swallows its own failures by contract.

Rewritten as `test_the_transfer_error_reaches_the_caller`: cleanup behaves normally and the assertion is that `SourceTransferError` reaches the caller with the `TimeoutError` as its `__cause__`. That is the invariant retry and disk fallback actually depend on.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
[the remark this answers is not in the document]
KavinKrishnan: You were right that the scope stopped short — pushed in `bca72a5`. The `try` now starts at acquisition rather than after it.

The case that made this worth doing is a half-finished P2P fetch: `fetch_remote_and_wait` dials the source and only *then* waits for `check_remote_metadata`, so a timeout can leave metadata that lands after we gave up, with nothing left to release it. Covering that needed the agent name claimed *before* the dial rather than taken from its return, which is possible in P2P because the name is known up front. Safe to release a peer that never finished loading, since `remove_remote_agent` documents an unknown peer as expected and never raises.

The centralized path is not symmetric and I did not force it to be: the name only exists once `add_remote_agent` returns, so a failure inside it leaves nothing to release. The `finally` is guarded rather than calling the manager with `None`.

Two tests added for the newly covered windows: `test_a_half_finished_p2p_fetch_is_released` and `test_a_failed_blob_load_releases_nothing`. 10 tests total, 9 of which fail against `main`.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -477,88 +477,126 @@ def _receive_from_peer(
...
+        # the reader can do it - in P2P only the reader loads a remote agent, so
+        # the source has no peer record to invalidate.
+        #
+        # The scope starts at acquisition and ends after the device sync, so the
+        # peer outlives every use of the bytes it served and no failure in between
+        # can leak it. A half-finished acquisition is the case most worth covering:
+        # fetch_remote_and_wait dials the source first and only then waits, so a
+        # timeout can leave metadata that lands afterwards with nothing to release
+        # it. remove_remote_agent treats an unknown peer as expected, so releasing
+        # one that never finished loading is harmless.
```
nv-hwoo:
```suggestion
```

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -477,88 +477,126 @@ def _receive_from_peer(
...
-        logger.info(
-            f"[Worker {ctx.global_rank}] [TIMING] RDMA transfer complete: "
-            f"{tensor_count} tensors, {bytes_transferred / 1e9:.2f} GB, "
-            f"{transfer_time:.3f}s, {bandwidth_gbps:.1f} Gbps"
-        )
 
-        ctx.accelerator_backend.synchronize()
+            ctx.accelerator_backend.synchronize()
+        finally:
+            if remote_agent_name is not None:
```
nv-hwoo, line 586: Let's add a short comment here

nv-hwoo (commented):
One small comment change. The PR looks good otherwise 👍

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
KavinKrishnan: Applied in `e35c7bf` — block dropped. Agreed on placement: it was sitting ~100 lines from the code it explained, and the retest narrative belongs in the PR and the bug record rather than the source.

Trimmed the two acquisition comments the same way, and shortened the test module docstring, which CodeRabbit had flagged for the same reason.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
KavinKrishnan: Added in `e35c7bf`:

```python
finally:
    # A weight load is one-shot, so release the source here instead of at
    # process exit: the engine process is torn down without running atexit
    # hooks, and in P2P only the reader holds a record of the peer to
    # invalidate. None means acquisition failed with nothing to release.
    if remote_agent_name is not None:
        ctx.nixl_manager.remove_remote_agent(remote_agent_name)
```

Kept it to the two things that are not evident from the code: why the release is here rather than at process exit, and what the `None` guard is for.

nv-hwoo (approved):

PR #579, opened 2026-07-30
feat(reshard): stamp the published shard table with its training step

KavinKrishnan:
## Why this is needed

A receiver discovers a table that says where each published weight shard lives. The table does not currently say which training step those shards belong to.

That gap matters once shard digests are checked. If discovery returns a table from step 10 while the receiver pulls weights from step 11, the bytes can transfer correctly and still fail verification because the expected digest is one step old. With many publishers updating at slightly different times, the receiver cannot safely infer freshness from whether a digest changed.

This PR adds an explicit training-step stamp to the published table. Later verification code can compare step numbers instead of guessing whether metadata is fresh.

## What changes

- `wrap_rendezvous_blob` accepts an optional `publisher_step` and adds it to the JSON payload.
- `unwrap_rendezvous_blob` returns the stamp as `RendezvousPayload.publisher_step`.
- A missing stamp decodes to `None`, meaning “unknown.” It does not decode to step 0, because step 0 is valid.
- An unstamped publisher omits the key entirely. This preserves the previous wire payload for publishers that do not use the feature.
- Existing callers use named fields instead of unpacking a fixed four-item tuple.

## Where it sits

```mermaid
sequenceDiagram
    participant T as Trainer publisher
    participant C as MX catalog
    participant R as Refit receiver

    T->>C: 1. Publish shard table + optional step
    R->>C: 2. Discover shard table
    C-->>R: 3. Return shards + publisher_step
    alt Stamp is present
        R->>R: 4. Record the observed training step
    else Stamp is absent
        R->>R: 4. Treat freshness as unknown
    end
    Note over R: Future PR compares the observed step before digest verification
```

This PR adds the observation in step 4. It does not add the future freshness decision or the digest comparison.

## Compatibility contract

Mixed-version deployments are expected while trainer and receiver images roll out.

- New reader + old publisher: the missing key becomes `None`.
- New reader + new publisher: the integer step is available by name.
- Unstamped publication: every existing JSON field keeps the same value and no `publisher_step` key is written.
- Step 0 remains distinguishable from an unknown step.

The payload named tuple now has five fields. The first four retain their positions, but callers should use named access because fixed-length positional unpacking is brittle when optional metadata grows.

## Scope

Included:

- wire-format support for the optional step;
- decoding into the existing `RendezvousPayload` type;
- compatibility and shard-table tests.

Not included:

- wiring the current optimizer/training step into every publisher;
- rediscovery during a refit;
- rejecting or retrying stale metadata;
- digest comparison.

Those pieces should land with the consumer so the policy and the data it acts on can be reviewed together.

## How to review

Review the four files in this order:

1. **`refit/reshard/rendezvous.py` — required review.** Check the meaning of `None`, preservation of step 0, and omission of the JSON key when no stamp is supplied.
2. **`tests/test_reshard_refit_step_stamp.py` — required review.** Check the mixed-version properties and that adding a stamp leaves all other payload fields unchanged.
3. **`tests/test_reshard_refit_rendezvous.py` — mechanical review.** Existing assertions now prefer named fields; positional stability is still tested explicitly.
4. **`tests/test_reshard_megatron.py` — mechanical review.** The Megatron test reads the same payload through named fields.

Questions for reviewers:

- Is `None` the right representation for “publisher did not provide freshness”?
- Should the wire key remain optional during mixed-version rollout?
- Is extending the existing named tuple preferable to adding a second unwrap function and return type?

## Validation

- The full local reshard/refit selection passed: 119 tests.
- The focused step-stamp suite passed: 9 tests.

## Follow-up

The next consumer should rediscover metadata for the target refit, compare the observed publisher step with the requested step, and only then evaluate shard digests. A missing stamp must remain “not verified,” not “verified successfully.”

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
@@ -217,13 +217,31 @@ def merge_shard_tables(tables: list) -> list:
 
 
 def wrap_rendezvous_blob(
-    agent_metadata: bytes, agent_name: str, metadata_endpoint: str, tensors: list
+    agent_metadata: bytes,
+    agent_name: str,
+    metadata_endpoint: str,
+    tensors: list,
+    publisher_step: int | None = None,
 ) -> bytes:
     """Pack ``{agent_meta, agent_name, metadata_endpoint, shard_table}`` into one
     JSON blob. ``metadata_endpoint`` (``host:listen_port`` of the trainer's NIXL
     listen thread) is what the receiver's ``fetch_remote_and_wait`` connects to
     for the P2P memory-registration handshake (the central agent-metadata blob
-    alone does not make the registrations resolvable for RDMA reads)."""
+    alone does not make the registrations resolvable for RDMA reads).
+
+    ``publisher_step`` stamps the table with the training step whose weights it
+    describes. A receiver otherwise has no way to tell a current table from one
+    published a step ago, and it needs to: the shard table carries the per-shard
+    digests a receiver verifies against, so a table one step behind makes correctly
+    delivered bytes read as corruption. Inferring freshness instead - asking whether
+    any digest changed since planning - holds only when a table is wholly stale or
+    wholly current, and breaks under partial propagation across many publishers,
+    where it reports one lagging publisher's shard as a hard defect. The stamp turns
+    that inference into an observation.
+
+    The key is omitted rather than written as null when absent, so a blob from an
+    unstamped publisher is byte-identical to one an older client would have produced.
+    """
     payload = {
         "schema": _SCHEMA,
         "agent_name": agent_name,
@@ -231,6 +249,8 @@ def wrap_rendezvous_blob(
         "agent_meta_b64": base64.b64encode(agent_metadata).decode("ascii"),
         "tensors": json.loads(encode_shard_table(tensors).decode("utf-8"))["tensors"],
     }
+    if publisher_step is not None:
+        payload["publisher_step"] = int(publisher_step)
     return json.dumps(payload).encode("utf-8")
 
 
@@ -239,13 +259,21 @@ class RendezvousPayload(NamedTuple):
 
     A plain tuple until now, which read as ``payload[3]`` at the point where the
     interesting question is asked - whether this rank published any tensors. Being
-    a tuple subclass, it still unpacks and indexes as before.
+    a tuple subclass, it still indexes as before.
+
+    ``publisher_step`` is ``None`` for a publisher that does not stamp, and that
+    must be read as "unknown", never as step 0: a publisher read as step 0 looks
+    permanently behind, and a consumer that excuses lagging publishers would then
+    excuse all of its shards, going quiet instead of strict. It carries a default
+    so that reading the stamp is a new field rather than a second unwrap function,
+    which is also why unpacking must now name five values.
     """
 
     agent_metadata: bytes
     agent_name: str
     metadata_endpoint: str
     tensors: list
+    publisher_step: int | None = None
 
 
 def unwrap_rendezvous_blob(blob: bytes) -> RendezvousPayload:
@@ -259,7 +287,11 @@ def unwrap_rendezvous_blob(blob: bytes) -> RendezvousPayload:
     tensors = decode_shard_table(
         json.dumps({"schema": _SCHEMA, "tensors": payload["tensors"]}).encode("utf-8")
     )
-    return RendezvousPayload(agent_metadata, agent_name, metadata_endpoint, tensors)
+    raw_step = payload.get("publisher_step")
+    publisher_step = None if raw_step is None else int(raw_step)
+    return RendezvousPayload(
+        agent_metadata, agent_name, metadata_endpoint, tensors, publisher_step
+    )
 
 
 class MxReshardRendezvous:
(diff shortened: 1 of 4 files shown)


KavinKrishnan:
@coderabbitai full review

KavinKrishnan:
Fixed in c51305a. The name overclaimed.

The test called the current `unwrap_rendezvous_blob`, so it showed current decoding rather than anything about a reader that predates the field, and its positional assertion already duplicated `test_the_first_four_fields_keep_their_positions` directly above it.

Replaced with the property such a reader actually depends on: the stamp is additive, so every other key holds the same value whether or not it is present. A reader that looks up only the keys it knows is unaffected by that. Running a released reader against the blob is the stronger check, and it is not something an in-process unit test can do, so it belongs in a version-compatibility run instead of a claim in a test name here.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -217,20 +217,40 @@ def merge_shard_tables(tables: list) -> list:
...
+    unstamped publisher is byte-identical to one an older client would have produced.
+    """
     payload = {
         "schema": _SCHEMA,
         "agent_name": agent_name,
         "metadata_endpoint": metadata_endpoint,
         "agent_meta_b64": base64.b64encode(agent_metadata).decode("ascii"),
         "tensors": json.loads(encode_shard_table(tensors).decode("utf-8"))["tensors"],
     }
+    if publisher_step is not None:
```
tanushriya910, line 252: thinking if we should make this required?

KavinKrishnan: Good question. I think it splits into three different kinds of "required" that land in different places.

**The wire field can't be required**, and not only for compatibility. `None` here means "this publisher doesn't stamp", which is genuinely different from any number it could carry instead. Step 0 is a real step — the first refit — so there's no spare value to press into service as "unknown". If we force a number, an unstamped publisher starts looking permanently behind, and a consumer that's lenient toward lagging publishers would then be lenient toward every one of its shards. That fails quiet, which is the worst direction for something whose whole job is catching stale data.

**The Python argument could be required**, with no default, so every caller has to pass something even if it's `None`. That does force the decision at each call site. Right now there's one production caller and it changes anyway in the propagation PR, so it'd be a line that gets rewritten immediately — but I'm happy to do it if you'd rather have the forcing function now than later.

**Where I think "required" really belongs is the consumer**, which isn't upstream yet. Once the receiver is actually checking freshness, an unstamped publisher is one it can't check. The right behavior is probably to refuse rather than skip the check silently, because otherwise turning verification on gives you a weaker guarantee than you'd expect against any publisher that hasn't been upgraded. That's a policy call I'd rather make explicitly in that PR than imply by making a field mandatory here.

Does that split match what you had in mind?

tanushriya910: I meant the python argument so trainer/sender always stamps the step and its not optional and receiver side also requires this to ensure its pulling the right step/version. fine with following up on this in later PRs where this is factored into receivers as well.

tanushriya910 (approved):

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
KavinKrishnan: That makes sense, and I agree with both halves — the sender should always stamp, and the receiver should require the stamp rather than treat it as a bonus. Following up in the propagation PR as you suggest. Concretely what lands there:

- `wrap_rendezvous_blob` loses the default on `publisher_step`, so every caller has to pass one.
- `megatron_publisher` passes the real training step instead of nothing.
- The receiver reads the stamp and compares it against the step it planned for, so a table one step behind is caught as staleness instead of surfacing as a digest mismatch on bytes that transferred fine.

One thing worth deciding there rather than discovering during a rollout: making the receiver *require* a stamp is a fail-closed change, and trainer and inference images get upgraded separately. Mid-rollout there will be old trainers that don't stamp, and a strict receiver refuses them. So we need to pick whether the requirement is unconditional, or only enforced when verification is switched on — the latter keeping a plain unverified refit working against an old publisher while anyone who asked for verification gets the real guarantee. I lean toward the second, since it means turning verification on never quietly gives you less than you asked for, but it's a judgement call about how much mixed-version tolerance we want and I'd rather have your read on it before I pick.

Rebased onto main and re-triggered CI.

KavinKrishnan:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/5716ee541e3ffd08a58023d48dc92bec3596556a

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
KavinKrishnan: Picking the option I flagged, so this isn't left hanging: the receiver requirement will be **enforced only when verification is switched on**, not unconditionally.

The reasoning is that an operator who hasn't enabled verification hasn't asked for the staleness guarantee, so refusing their refit against a not-yet-upgraded trainer costs them something they didn't ask for. Whereas anyone who does enable it should get the real guarantee rather than a check that quietly skips any publisher that doesn't stamp. That also keeps a partially rolled-out fleet able to refit, which an unconditional requirement wouldn't.

Sender side is unconditional as you asked — `wrap_rendezvous_blob` loses the default and `megatron_publisher` passes the real step, so every publisher in our tree stamps. Both land in the propagation PR. Shout if you'd rather have it strict on both sides and I'll flip it there.

PR #581, opened 2026-07-31
fix: honor ignore_weights when streaming model files from the server

AndyDai-nv:
## What

`Client::request_model` takes an `ignore_weights` flag but never forwarded it to `stream_model_files_from_server`, which hard-coded `file_selector: None` and streamed **every** file including weight shards. So a config-only request still re-streamed the full weight set from the server whenever shared storage was disabled (`MODEL_EXPRESS_NO_SHARED_STORAGE=1`).

This PR:
- adds an `ignore_weights` field to `ModelFilesRequest` (mirrors `ModelDownloadRequest.ignore_weights`),
- threads the flag from `request_model` → `stream_model_files_from_server` → the gRPC request,
- has the server's `collect_model_files` skip weight files when set (applies to both `StreamModelFiles` and `ListModelFiles`),
- extracts the weight-file check into a shared `providers::is_weight_file` free function so the streaming path and the provider download path share one definition.

`StreamModelFiles`/`ListModelFiles` already supported selective streaming via `file_selector`; this makes weightless streaming a first-class option.

## Why

Defense-in-depth for the redundant-weight-stream class of bug (context: investigation of ai-dynamo/dynamo#12097). When a caller only needs config/tokenizer (e.g. building a model card) but runs with shared storage disabled, it should not re-pull the full weight set from the server. Today the miss is masked because callers that pass `ignore_weights=true` also cause the server to hold no weight files — but if the server already has the full model cached, the stream would pull the weights anyway. This makes the client request match its intent.

## Testing

- New unit test `test_collect_model_files_ignore_weights` (weight shards excluded at any depth; config/tokenizer kept).
- `cargo build --workspace`, `cargo clippy --workspace --all-targets` (no warnings), `cargo test -p modelexpress-server --lib` (230 passed), client stream tests, and `modelexpress-common` provider tests all green.

No Python change: the Python client is metadata-only and does not consume `ModelService`. The weight-file extension list is already documented in `docs/GCS_PROVIDER.md`.

## Summary by CodeRabbit

- **New Features**
  - Model file downloads, listings, and streaming can now exclude weight files when requested.
  - Weight files are detected consistently, including nested weight shards.
  - Existing file selection and missing-file validation continue to work when weights are excluded.

- **Bug Fixes**
  - Improved consistency between model downloads and streamed file transfers when filtering out weight files.

### DIFF
--- a/modelexpress_client/go/gen/modelexpress/model/model.pb.go
+++ b/modelexpress_client/go/gen/modelexpress/model/model.pb.go
@@ -371,7 +371,10 @@ type ModelFilesRequest struct {
 	// Chunk size in bytes for streaming (0 = use server default)
 	ChunkSize uint32 `protobuf:"varint,3,opt,name=chunk_size,json=chunkSize,proto3" json:"chunk_size,omitempty"`
 	// Optional file selector. If unset, all files are streamed.
-	FileSelector  *ModelFileSelector `protobuf:"bytes,4,opt,name=file_selector,json=fileSelector,proto3" json:"file_selector,omitempty"`
+	FileSelector *ModelFileSelector `protobuf:"bytes,4,opt,name=file_selector,json=fileSelector,proto3" json:"file_selector,omitempty"`
+	// When true, weight files are skipped and only non-weight files
+	// (config, tokenizer, etc.) are streamed. Mirrors ModelDownloadRequest.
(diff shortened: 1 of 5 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/581/commits/681cc3809bb4914c690e910116f73ec39b4fab9a

zhengluo-nv (approved):

PR #582, opened 2026-07-31
docs(p2p): record the source wedge after reader teardown as a known issue

KavinKrishnan:
Adds a README known-issues entry for an issue observed on UCX/InfiniBand, tracked as nvbug 6519532.

Once a reader that pulled from a P2P source is torn down, the source can stop serving new readers: their RDMA reads stall to the transfer timeout and fall back to disk, while the source keeps advertising `Ready` and serving inference normally. Loading stays correct; the cost is that P2P is lost for that source until its worker pod is restarted. Lowering `MX_TRANSFER_TIMEOUT` bounds the stall.

Not fixable from ModelExpress — the responder holds UCX queue-pair state for a requester it keeps no NIXL-level record of, so nothing on either side reaches it. Being followed up upstream.

Docs only, one line.


ganeshku1 (approved):

KavinKrishnan:
/ok to test [c501900](https://github.com/ai-dynamo/modelexpress/pull/582/commits/c5019008a3fba258ddc12fc1e69f9c6dcf66e842)

nv-nmailhot (approved):

Issue #471, opened 2026-07-01
Topology-aware source metadata for MoE and P/D scaleout


athreesh:
## Summary

ModelExpress source identity is currently mostly model + backend + TP/PP/EP + dtype/quantization.

That is enough for simple P2P cold-start cases, but it isn't enough for mapping topologies that users now care about: WideEP, TEP/DEP, P/D disagg, shadow/failover engines, and EPLB-style expert layout.

We should explore an engine-agnostic topology metadata contract so MX can tell when two workers are actually compatible sources for each other. 

## Motivation
Large MoE deployments have new requirements, especially when served with EP. Oftentimes, the EPLB info is essential for ensuring the transfer starts a new worker that matches the source worker.

A worker's compatibility can depend on factors like:
- whether it's a P/D worker, agg worker, or shadow engine,
- sharding layout,
- WideEP rank layout,
- expert placement,
- whether EPLB /redundant experts are enabled,
- whether expert traffic expects NVL/MNNVL, but KV or weight movement can use RDMA,
- engine runtime (SGL, vLLM, TRT-LLM, etc.)

Today MX has no first-class place to represent this. We need to explore what fields affect source compatibility, which fields are per-worker runtime state, and how different engines want to describe the same topology. Especially for MoE, (wide)EP, P/D deployments.

## Proposal
Add a topology descriptor to MX metadata that is common across engines. The first version should focus on source compatibility and discovery, not on making MX own runtime scheduling.

Some possible fields:
- serving worker role: `prefill`, `decode`, `aggregated`, `shadow`, etc.
- parallelism shape: TP, PP, DP, EP, CP/PCP where available,
- rank identity: global rank, local rank, DP rank, EP rank, engine group id,
- topology domain: node, rack, NVL domain, RDMA fabric, compute domain,
- expert metadata: logical experts, physical slots, redundant experts, placement strategy,
- EPLB metadata: enabled/disabled plus a stable config/layout digest,
- transport expectations: RDMA, NVL/MNNVL, local-only, or mixed paths,
- framework-specific escape hatch for fields that do not normalize cleanly yet.

This ideally should stay backend-agnostic. vLLM DP+EP, SGLang WideEP, and TRT-LLM disagg/WideEP should all be able to populate the same shape, but map to engine specific implementations.

## Acceptance
- MX can distinguish incompatible MoE/P/D workers that currently look identical.
- MX can track info such as EPLB and use that to select the right worker/replicate for target new workers.
[...]

PR #585, opened 2026-08-02
fix(client): rebuild engine state from received RDMA weights

abatilo:
## Problem

On the SGLang `remote_instance` / modelexpress load path, a receiving engine builds the model with random-initialized weights and runs all post-load derivation **before** the P2P transfer lands. The transfer then overwrites only the manifest tensors (parameters and buffers), so state derived from weight content stays computed from random values. One example is SGLang's `_attn_res_cw_cache` dictionary on each residual projection module, which `get_cw` reads and never recomputes on a hit.

The resulting engine passes every manifest check (name, shape, dtype all match) but produces invalid output: correct text at first, then degradation into invalid text. On a 32-GPU mixture-of-experts cell, measured speculative-decoding acceptance was **0.0 across 1,076 verify calls** while an identical S3-loaded replica was healthy.

This matches the source-identity coverage gap reported in #471: two workers can look identical as sources but still produce different models when derived state differs.

## Fix

`SglangAdapter.after_rdma_receive` now resets the value-derived caches (the one engine-engaged structure the manifest provably cannot carry today) and re-runs the engine's model-level `post_load_weights` from the received weights.

Deliberately not re-run: `_process_weights_after_loading` — the received tensors are already post-processed, so re-packing a packed model would corrupt it. Model-level `post_load_weights` is idempotent by construction.

## Verification (4-node Kubernetes LeaderWorkerSet, 8 H200 GPUs/node, 1.5Ti Kimi-K3 MXFP4, sglang 0.5.16 kimi-k3 branch, modelexpress 0.5.0)

- **Before:** P2P receivers generated invalid text and accepted no speculative drafts while the identical S3-loaded replica was healthy.
- **After:** the same group passes short generations and a 102,340-token long-context request.
- **Spec decode** returned to the healthy replica's level: ~3 accepted tokens/call at ~32% acceptance (vs baseline ~3.33 at ~33%).
- **Transfer telemetry unchanged:** 2,858 tensors, 60.85 GB/rank at 285 Gbps in 1.7 s.

## Tests

`tests/test_sglang_adapter.py` — cache reset & re-derivation, missing-model error path, cache-less model tolerance (3/3 passing). Existing adapter suite still green.

## Context

- Discovered and root-caused against a multi-node (LWS) SGLang TP32 deployment; the P2P failure mode was deterministic on every receive attempt until this hook.
- Related: #471 (topology-aware source metadata for MoE and P/D scaleout)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model handling after RDMA transfers by clearing stale derived caches and rerunning required post-load setup.
  * Added validation for missing models and support for models without cache or post-load hooks.

* **Tests**
  * Added coverage for cache clearing, post-load processing, validation errors, and models without optional components.


AndyDai-nv (commented):
/ok to test https://github.com/ai-dynamo/modelexpress/pull/585/commits/43f86cd8b2098ac22d6554d924404ac73cf9682d

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/585/commits/43f86cd8b2098ac22d6554d924404ac73cf9682d

modelexpress_client/python/modelexpress/engines/sglang/adapter.py
```diff
@@ -97,6 +97,34 @@ def before_rdma_receive(self, result: LoadResult) -> LoadResult:
...
+        would corrupt it. Model-level ``post_load_weights`` only computes
+        value-derived views/caches and is idempotent by construction.
+        """
+        if result.model is None:
+            raise RuntimeError("SGLang post-receive derivation requires result.model")
+        for _, module in result.model.named_modules():
+            if getattr(module, "_attn_res_cw_cache", None) is not None:
+                module._attn_res_cw_cache = {}
+        logger.info("re-deriving engine state from received RDMA weights")
+        return self._post_load_weights(result)
```
zhengluo-nv: **[P1] Do not rerun model-wide `post_load_weights()` unconditionally**

`before_rdma_receive()` has already invoked this hook before tensor registration and transfer, but SGLang does not guarantee that it is idempotent. In the pinned `v0.5.13.post1` release, `DeepseekV4ForCausalLM.post_load_weights()` calls `_setup_fp8_wo_a_scales()` when `SGLANG_OPT_FP8_WO_A_GEMM` is enabled (the default on Blackwell); that method replaces `weight_scale_inv.data` with `transform_sf_into_required_layout(...)`. The source tensor arriving here is already in that transformed layout, so calling the hook again double-transforms the scale factors. Replacing the storage after NIXL/TransferEngine registration can also leave the receiver's subsequent source registration referring to the pre-hook allocation. A DeepSeek-V4 P2P receiver on B200 can therefore produce corrupted inference and be unsafe as a source even though the existing Hopper SGLang checks pass.

Please make the post-receive operation a narrowly defined idempotent hook (or restrict it to the proven Kimi cache path), and add a DeepSeek-V4/Blackwell regression test verifying that the already-postprocessed scale tensor is neither transformed twice nor reallocated. Upstream implementation: https://github.com/sgl-project/sglang/blob/85fd90072d1a9f2432842b03588f63b745e524e4/python/sglang/srt/models/deepseek_v4.py#L1856-L1884

zhengluo-nv (commented):
I found one compatibility regression that should be addressed before merge.

AndyDai-nv (commented):
Thanks for your contribution, @abatilo some models perform a non-idempotent layout transform inside post_load_weights as @zhengluo-nv points out, so it might not be safe to re-run the method

PR #588, opened 2026-08-04
refactor(weight_transfer): execute NIXL reads through the existing batch API

nicolasnoble:
The weight_transfer package landed with its own copy of the NIXL issue-and-wait sequence: prep_xfer_dlist twice, make_prepped_xfer, transfer, a check_xfer_state poll, release_xfer_handle. NixlTransferManager.post_read_batch and await_read_batches already do exactly that, so this collapses NixlExecutor onto them. Getting there also required reaching into the manager's private attributes, which goes away with it. After this there is no NIXL call left anywhere under weight_transfer/.

Three commits, in order. The first adds characterization tests for NixlExecutor, asserted at the NIXL agent boundary rather than against the executor's internals, so they hold unchanged across the reimplementation. Nothing in the suite exercised NixlExecutor or PullRole before, which means the existing tests would have stayed green whether or not the swap was faithful.

The second deletes PushRole and M2nExecutor. Neither has a construction site anywhere in the repo, tests included, and M2nExecutor was itself a third copy of the same block. Dropping PushRole takes the NIXL WRITE direction with it, which is what lets the READ-only batch API be reused.

The third does the swap. execute() groups descriptors by remote agent, posts one batch per agent, and awaits the set once. Polling gets slightly better as a side effect, since _wait_for_xfers sweeps the whole handle set per iteration where the old loop spun on one handle at a time.

Two smaller things worth a look in review. NixlExecutor now refuses construction when its device_id disagrees with the manager's, because post_read_batch builds local descriptors from the manager's device id and the swap changes which value reaches them. The two are equal today only because TrainerPullStrategy passes the same ctx.device_id to both, and nothing enforced it. This turns a silent transfer to the wrong device into a startup failure. Separately, await_read_batches takes a label parameter now, defaulting to the existing string, so a failure on the pull path stops reporting itself as the reshard path.

The remote device id stays at 0, since RdmaDescriptor has no field for it. It has a name now instead of being a bare literal, but lifting it properly means adding the field and threading it through every planner that builds a descriptor.

Excluding seven test files that fail collection on a missing google_crc32c locally, the suite goes from 67 failed, 831 passed, 45 skipped on main to 67 failed, 844 passed, 45 skipped here. Same failures both sides, and the difference is the thirteen new tests.

None of this exercises real RDMA. The new tests run against a mocked agent, so the device id handling in particular wants a run on real hardware before it merges.

## Summary by CodeRabbit

* **Updates**
  * Simplified weight synchronization to use pull-based transfers.
  * Removed trainer-side push synchronization support.
  * Consolidated M2N transfers under the standard pull workflow.
  * Streamlined transfer execution to support batched read operations.

* **Bug Fixes**
  * Improved transfer timeout and failure messages with operation-specific labels.
  * Added validation for matching device identifiers and safer cleanup after transfer errors.

* **Tests**
  * Added coverage for device handling, grouping, empty transfers, failures, cleanup, and error reporting.


KavinKrishnan (dismissed):

modelexpress_client/python/modelexpress/weight_transfer/transport/nixl_executor.py
```diff
@@ -21,11 +25,14 @@
...
 logger = logging.getLogger("modelexpress.weight_transfer.transport")
 
-NixlOperation = Literal["READ", "WRITE"]
+# RdmaDescriptor has no remote-device field, so the trainer side is assumed to
+# sit on device 0. Lifting this means adding the field to the descriptor and
+# threading it through every planner that builds one.
+REMOTE_DEVICE_ID = 0
 
 
 class NixlExecutor:
```
tanushriya910, line 34: this looks very similar to the NixlReshardTransport class here: https://github.com/ai-dynamo/modelexpress/blob/main/modelexpress_client/python/modelexpress/refit/reshard/transport/nixl.py#L41. thinking if we can combine further

modelexpress_client/python/modelexpress/weight_transfer/transport/nixl_executor.py
```diff
@@ -21,11 +25,14 @@
 
 logger = logging.getLogger("modelexpress.weight_transfer.transport")
 
-NixlOperation = Literal["READ", "WRITE"]
+# RdmaDescriptor has no remote-device field, so the trainer side is assumed to
+# sit on device 0. Lifting this means adding the field to the descriptor and
+# threading it through every planner that builds one.
+REMOTE_DEVICE_ID = 0
```
KavinKrishnan, line 31: Can we avoid assuming that every trainer keeps its weights on GPU 0? That may work in a simple setup, but it can read from the wrong GPU when the model is spread across several GPUs. I think the descriptor should carry the trainer GPU number instead.

modelexpress_client/python/modelexpress/weight_transfer/transport/nixl_executor.py
```diff
@@ -67,74 +73,38 @@ def execute(
...
-                    agent.release_xfer_handle(handle)
-                except Exception as e:
-                    logger.warning("Failed to release NIXL %s handle: %s", operation, e)
-
-        self._manager._accelerator_backend.synchronize(self._device_id)
+            except Exception as exc:  # noqa: BLE001 - cleanup must not mask the cause
+                logger.warning("Drain after a failed READ post: %r", exc)
+            raise
+
+        total_bytes, _num_reads, _duration = self._manager.await_read_batches(
```
KavinKrishnan, line 100: This should land with #590, or after it. Right now a failed batch can raise here while the manager still says the connection is healthy. #590 fixes that and adds tests for timeouts and transfer errors.

modelexpress_client/python/modelexpress/weight_transfer/transport/nixl_executor.py
```diff
@@ -67,74 +73,38 @@ def execute(
...
-                dst_prepped = agent.prep_xfer_dlist(
-                    agent_name="" if operation == "READ" else remote_name,
-                    xfer_list=dst_list,
-                    mem_type=mem_type,
-                    backends=backends,
+                posted.append(
+                    self._manager.post_read_batch(
+                        remote_agent_name=remote_name,
+                        ranges=[
+                            (d.src_addr, d.dst_addr, d.nbytes, REMOTE_DEVICE_ID)
```
KavinKrishnan, line 83: Before we build these reads, can we fail if any trainer is missing instead of quietly leaving it out? Otherwise part of the model keeps old weights while the rest is updated, and the run can look successful.

modelexpress_client/python/modelexpress/weight_transfer/transport/nixl_executor.py
KavinKrishnan: I agree. Both classes now do the same basic job: group reads by trainer, start them together, wait for them, and clean up if one fails. I would prefer one shared implementation. Before combining them, though, we need to keep the safer behavior from NixlReshardTransport: use the actual trainer GPU number and stop if a trainer is missing.

nicolasnoble:
closing. all of 481 is coming out, so this refactors code that won't exist.

your two points outlive it and the replacement needs both: carry the trainer gpu number rather than assuming 0, and fail loudly on a missing trainer.

PR #589, opened 2026-08-04
fix(sglang): adopt container-held derived tensors into the RDMA manifest

AndyDai-nv:
## Problem

`capture_tensor_attrs` promotes bare `Tensor` attributes assigned on an `nn.Module`, which is how derived state like MLA's `w_kc`/`w_vc` reaches the manifest. It does not reach tensors held **inside a container**, or on a **non-Module object**:

- `module._attn_res_cw_cache = {}` assigns a `dict`, which fails the `isinstance(value, torch.Tensor)` check
- `cache[dtype] = cw` goes through `dict.__setitem__`, which never touches `nn.Module.__setattr__` at all

Such tensors are invisible to `named_parameters()`/`named_buffers()`, so they are absent from the RDMA manifest. On a target they keep whatever `before_rdma_receive` derived from **random-init** weights. The transfer then overwrites only manifest tensors, so the engine runs with part of its state computed from noise — while every manifest check (name, shape, dtype) passes and the transfer reports success.

Concretely on Kimi-K3:

| state | storage | reaches manifest today |
|---|---|---|
| `_attn_res_cw_cache` | `dict` keyed by dtype, short-circuited by `get_cw` | no |
| `_k3_fused_decode_args` | `tuple` consumed by the fused KDA decode kernel | no |

`cw = norm.weight * proj.weight.squeeze()` is the `[hidden_size]` scoring vector that weights the attention-residual mixture at every aggregation point, so a stale one silently distorts every layer's residual stream. SGLang's MXFP4 FusedMoE `quant_method` is a third case — it holds swizzled weights after deleting the original params.

## Fix

`discover_tensors` calls `adopt_hidden_tensors`, matching `engines/vllm/adapter.py:184`. The two adapters' `discover_tensors` were structurally identical apart from this line, which was commented out with a note to enable it "when SGLang nested-object coverage is in scope". Kimi-K3 is that case.

`adopt_hidden_tensors` registers the **same tensor objects** as non-persistent buffers, so the container entries alias the transferred memory — an RDMA write into the manifest entry is visible through the dict or tuple, with no engine-side change.

This keeps the receive path on its existing contract: **transfer the final state, derive nothing afterwards.** That contract is what keeps engine post-load hooks off the receive path, where re-running them would double-transform already-processed weights (e.g. `DeepseekV4ForCausalLM.post_load_weights` calls `_setup_fp8_wo_a_scales`, which re-lays-out `weight_scale_inv` with no already-transformed guard, under a default-on env flag) and would move tensors that are already registered and published.

## Tests

`tests/test_sglang_hidden_tensors.py` — dict-held, tuple-held and quant-method-held tensors reach the manifest; the manifest entry **aliases** the container entry so an RDMA write propagates; non-tensor tuple members are ignored; the non-tensor tuple member is not adopted (asserted by count, since the collector returns tensors by construction); aliases of existing parameters are not duplicated.

`tests/test_sglang_rdma_receive_invariants.py` — drives the real adapter through `before_rdma_receive -> discover -> register -> transfer` against a model transcribed from SGLang's `kimi_k3.py` / `attn_residual.py`, and pins two invariants: the value-derived cache matches a fresh derivation from the received weights, and no registered tensor moves or detaches from the model.

Python suite: 1083 passed, 1 skipped (excluding `tests/gpu`, which needs `transformers` and a GPU).

## Open question for maintainers: transfer-ABI version

This adds `_mx_*` manifest names. Same-family transfers tolerate subset manifests (`require_exact_match` defaults `False`), so **a target on this build paired with a source without it silently leaves the adopted entries at dummy values** -- the same failure mode this PR fixes.

`SourceIdentity` derives `mx_version` from the `modelexpress` package version and has no free-form capability field, so there is no surgical way to separate old and new workers short of a workspace version bump. Per `CLAUDE.md` that is a release action touching ~10 files plus regenerated source-id hashes, so I have not done it here.

**Please decide whether this should land together with a version bump, or wait for the next one.** Flagging rather than deciding, since it affects the release train.

## Hardware validation status

Mechanism and coverage are confirmed offline. A full Kimi-K3 run is staged but has not completed: the image now builds on `lmsysorg/sglang:kimi-k3` (thanks @zhengluo-nv -- the source checkout it replaced was unnecessary) with a build-time assertion on the two private symbols this depends on, and the 1.5T checkpoint is staged on a cluster PVC. Seed rank 0 came up and initialised correctly; rank 1 never scheduled -- nodes reporting `Ready`, uncordoned and fully unallocated were still rejected with `Insufficient nvidia.com/gpu`. Raised with the cluster maintainers.

Still unmeasured, and not claimed here:

- source/target `_mx_*` name symmetry on a real K3 (the collision suffix in `adopt_hidden_tensors` is order-dependent)
- what each quantization path newly adopts — MXFP4, CUTLASS FP8/W4A8 — and that none of it is device-local scratch such as the pointer arrays in `fp8.py`'s CUTLASS buffers, which are lazily created at first forward and therefore absent at discover time today, but only by timing
- `adopt_hidden_tensors` walk cost on a 92-layer model
- transfer size delta

A dry-run mode that reports what *would* be adopted, per quantization path, would de-risk the last three before enabling by default.

Related: #585 addresses the same Kimi-K3 symptom by re-running `post_load_weights` after the transfer; #471 (source-identity coverage for derived state).

## Summary by CodeRabbit

* **New Features**
  * Added an option to stream, list, or download model files without weight files.
  * Weight-file detection now works consistently across supported file formats and nested paths.
  * SGLang model discovery now includes hidden tensors in nested containers and cache state.

* **Bug Fixes**
  * Preserved tensor aliases and live tensor registrations during SGLang loading and RDMA transfers.

* **Tests**
  * Added coverage for hidden tensor discovery, weight exclusion, alias handling, and RDMA receive invariants.

### DIFF
--- a/examples/p2p_transfer_k8s/client/sglang/Dockerfile.k3
+++ b/examples/p2p_transfer_k8s/client/sglang/Dockerfile.k3
@@ -1,0 +1,50 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# SGLang Kimi-K3 support lives on the sgl-project/sglang `kimi-k3` branch and
+# is not in any published tag. Build it over a released image so the CUDA /
+# torch / flashinfer stack comes from the base and only the Python package is
+# replaced.
+#
+# Build:
+#   docker build -f examples/p2p_transfer_k8s/client/sglang/Dockerfile.k3 \
+#     --build-arg MX_REF=pull/589/head \
+#     -t <registry>/sglang-k3-mx:pr589 .
+#
+# MX_REF selects what ModelExpress client to install:
+#   pull/589/head  -> adopt container-held derived tensors into the manifest
+#   pull/585/head  -> re-derive after the transfer instead
+#   main           -> baseline for A/B runs
+
+ARG SGLANG_IMAGE=lmsysorg/sglang:v0.5.16
+FROM ${SGLANG_IMAGE}
+
+# Pinned tip of the kimi-k3 branch. Repin deliberately: the branch moves and
+# `_attn_res_cw_cache` / `get_cw` are private symbols the MX fix keys off.
+ARG SGLANG_K3_COMMIT=1f9d689b2918b17c2f3876997464ac686978d971
+RUN git clone --filter=blob:none https://github.com/sgl-project/sglang /opt/sglang-k3 && \
+    cd /opt/sglang-k3 && \
+    git checkout ${SGLANG_K3_COMMIT} && \
+    python3 -m pip install --no-cache-dir --no-deps -e python/
+
+RUN python3 -c "from sglang.srt.layers.attn_residual import get_cw; \
+    import inspect; assert '_attn_res_cw_cache' in inspect.getsource(get_cw)"
+
+ARG INSTALL_MOONCAKE=true
+RUN if [ "${INSTALL_MOONCAKE}" = "true" ]; then \
+        python3 -m pip install --no-cache-dir mooncake-transfer-engine-cuda13; \
+    fi
+
+ARG MX_REF=pull/589/head
+RUN python3 -m pip uninstall -y modelexpress 2>/dev/null || true && \
+    rm -rf "$(python3 -c 'import site; print(site.getsitepackages()[0])')"/modelexpress* && \
+    git clone --filter=blob:none https://github.com/ai-dynamo/modelexpress /opt/mx && \
+    cd /opt/mx && \
+    git fetch origin ${MX_REF}:mx-under-test && \
+    git checkout mx-under-test && \
+    python3 -m pip install --no-cache-dir --no-deps ./modelexpress_client/python
+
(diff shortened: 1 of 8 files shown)


examples/p2p_transfer_k8s/client/sglang/Dockerfile.k3
```diff
@@ -0,0 +1,50 @@
...
+#
+# MX_REF selects what ModelExpress client to install:
+#   pull/589/head  -> adopt container-held derived tensors into the manifest
+#   pull/585/head  -> re-derive after the transfer instead
+#   main           -> baseline for A/B runs
+
+ARG SGLANG_IMAGE=lmsysorg/sglang:v0.5.16
+FROM ${SGLANG_IMAGE}
+
+# Pinned tip of the kimi-k3 branch. Repin deliberately: the branch moves and
```
zhengluo-nv: Are these workaround required? 

There is a kimi-k3 release image, we could try this
lmsysorg/sglang:kimi-k3

AndyDai-nv: sounds good, this pr is not verified yet due to limited compute resources, just put it here first

PR #590, opened 2026-08-04
fix(nixl): record data-plane failures on the batched READ wait

KavinKrishnan:
## Why this is needed

ModelExpress uses a health flag to stop advertising a transfer agent after the data plane fails. The single-transfer wait updates that flag, but the batched READ wait did not.

This matters because batched READs are now the normal path for reshard refits, and PR #588 also uses them for the `weight_transfer` path. Before this fix, a batch could time out or return a terminal NIXL error while `is_healthy()` continued to report the agent as healthy. Other workers could keep selecting a source whose network path was already known to be broken.

The failure is not theoretical. A wedged UCX/InfiniBand queue pair can remain pending until the application timeout without returning an error status. In that case, the timeout is the only signal that the data plane failed.

## What changes

`NixlTransferManager._wait_for_xfers` now follows the same health rules as the existing single-transfer wait:

- A timeout records `_data_plane_error` before raising `TimeoutError`.
- A terminal NIXL status records `_data_plane_error` before raising `RuntimeError`.
- The error is cleared only after every handle in a non-empty batch completes successfully.
- An empty batch does not clear an earlier failure because waiting on nothing does not prove the network recovered.

A later successful batch is allowed to clear an earlier failure. This lets a worker recover after a transient fault instead of remaining unhealthy for the life of the process.

## Where it sits

```mermaid
flowchart TB
    A["<b>1. Post a batch of NIXL READs</b>"] --> B["<b>2. Wait for all handles</b>"]
    B --> C{"<b>3. Batch result</b>"}
    C -->|Timeout| D["<b>4. Record data-plane error</b>"]
    C -->|NIXL error| D
    D --> E["<b>5. is_healthy returns false</b>"]
    C -->|All complete| F["<b>4. Clear earlier error</b>"]
    F --> G["<b>5. is_healthy returns true</b>"]
    C -->|No handles| H["<b>4. Leave health unchanged</b>"]

    class A,B mx;
    class C,D,E,F,G,H receiver;
```

## Scope

Included:

- health-state updates for the existing batched wait;
- recovery semantics after a successful batch;
- unit coverage through both the private helper and public `await_read_batches` entry point.

Not included:

- changes to NIXL transfer posting;
- retry policy;
- source eviction or publisher teardown policy;
- changes specific to PR #588.

The bug predates PR #588 and affects every caller of `await_read_batches`, so the fix stays in the shared transfer manager.

## How to review

This PR has two files. Review them in this order:

1. **`modelexpress/nixl_transfer.py` — required review.** The source change is small. Confirm that failure is recorded before either exception, and that success clears the flag only after the full non-empty batch completes.
2. **`tests/test_nixl_peer_lifecycle.py` — required review.** Focus on `TestHealthReflectsTheBatchDataPlane`. It mirrors the existing single-transfer tests and covers timeout, terminal error, mixed handle outcomes, recovery, empty batches, and the public batch-wait method.

Questions for reviewers:

- Should one completed batch be sufficient evidence to restore health after a transient failure?
- Is whole-batch success the correct clearing boundary? Clearing per handle would allow a completed sibling to hide another handle's failure.
- Should an empty batch leave a previous failure latched? This PR says yes because no data-plane work was observed.

## Validation

- `tests/test_nixl_peer_lifecycle.py`: 27 passed.
- The broader reshard, refit, NIXL, publisher, and many-to-many transfer selection passed: 202 tests.
- Five new tests fail against the old helper, covering timeout, terminal error, recovery, one failed handle among siblings, and the public entry point.

No real Remote Direct Memory Access (RDMA) hardware is used in these unit tests; the NIXL agent is stubbed, consistent with the existing lifecycle suite.

### DIFF
--- a/modelexpress_client/python/modelexpress/nixl_transfer.py
+++ b/modelexpress_client/python/modelexpress/nixl_transfer.py
@@ -505,16 +505,27 @@ class NixlTransferManager:
 
         Sleeps only when a full sweep completed nothing, so the polling slop is
         paid once for the whole set rather than once per handle.
+
+        Records data-plane failures exactly as :meth:`_wait_for_xfer` does, for the
+        same reason: a wedged QP yields neither a completion nor an ERR status, so
+        the timeout is the only evidence anything went wrong, and recording it is
+        what lets ``is_healthy()`` stop advertising this agent.
         """
         if self._agent is None:
             raise RuntimeError("NIXL agent not initialized")
         pending = list(handles)
+        waited_on_something = bool(pending)
         wait_start = time.perf_counter()
         while pending:
             if (
                 timeout_seconds is not None
                 and time.perf_counter() - wait_start >= timeout_seconds
             ):
+                self._data_plane_error = (
+                    f"{label} timed out after {timeout_seconds:.1f}s with "
+                    f"{len(pending)} transfer(s) outstanding and no error status "
+                    f"from NIXL"
+                )
                 raise TimeoutError(
                     f"{label} timed out with {len(pending)} transfer(s) outstanding"
                 )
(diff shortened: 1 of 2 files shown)


KavinKrishnan:
@coderabbitai full review

tanushriya910 (approved):

KavinKrishnan:
/ok to test https://github.com/ai-dynamo/modelexpress/commit/ec84b79d60335d3972adc4a7e8f966ced2352ed4

PR #591, opened 2026-08-04
feat(reshard-refit): publish a per-shard digest for later verification

KavinKrishnan:
## Why this is needed

A refit can move the expected number of bytes and still install the wrong values. The transfer may succeed while the plan points a copy at the wrong region of a tensor. End-to-end loss or generation checks can reveal that the model changed, but they do not identify which published shard was wrong.

A receiver-side check needs an expected value for each shard: “these are the bytes the publisher held when it advertised this region.” This PR adds that expected value to the publish and discovery path. It is the producer half of verification; it does not yet reject a refit.

## What changes

- Publishers can compute a digest for each advertised shard when `MX_RESHARD_PUBLISH_DIGEST=1`.
- The optional digest is carried in the shard-table JSON, decoded during discovery, and preserved in the planner's `Shard` input.
- Megatron publishers compute the digest over the exact region they advertise. For Query-Key-Value (QKV) aliases, that means the narrow shard rather than the fused parent tensor.
- `shard_region` extracts the matching publisher-owned box from a full-pull staging buffer for the later receiver-side comparison.
- With the flag off, no digest is computed and the wire payload keeps its previous shape.

The feature is off by default because digesting every published tensor adds a full read pass over model weights. It is intended for correctness qualification, not throughput measurements.

## Where it sits

```mermaid
flowchart TB
    A["<b>1. Publisher owns a tensor shard</b>"] --> B{"<b>2. Digest publication enabled?</b>"}
    B -->|No| C["<b>3. Publish the existing shard entry</b>"]
    B -->|Yes| D["<b>3. Compute shard digest on GPU</b>"]
    D --> E["<b>4. Add digest to shard table</b>"]
    C --> F["<b>5. Catalog discovery</b>"]
    E --> F
    F --> G["<b>6. Planner carries expected digest</b>"]
    G -. Future PR .-> H["<b>7. Receiver recomputes and compares</b>"]

    classDef trainer fill:#E3F2FD,stroke:#1565C0,stroke-width:2px,color:#0D47A1,font-size:16px;
    class A,B,D,E trainer;
    class C,F,G mx;
    class H receiver;
```

The dashed step is not part of this PR.

## Digest design and limits

This is a position-sensitive reduction, not a cryptographic byte hash.

The implementation views bytes as 32-bit words, reduces them in 4 KiB rows on the GPU, and folds the row results in index-weighted order before producing a compact host-side value. Index weighting matters because a copy plan can place correct data at the wrong offsets; a plain sum or sum-of-squares would not detect that permutation.

The current granularity catches whole-row reordering and value changes. It does not guarantee detection of every possible word permutation within one 4 KiB row. That tradeoff avoids copying full tensors to the CPU or adding a custom GPU hash kernel. Reviewers should treat the result as a qualification signal tailored to the known planning failure mode, not as a general integrity or security hash.

Ragged bytes after the last complete 32-bit word are included. The helper also rejects mismatched tensor ranks instead of allowing `zip` to silently truncate the requested shard coordinates.

## Compatibility contract

- A publisher without a digest omits the JSON key.
- A missing digest decodes to `None` and means “not checked,” not “matched.”
- Existing `Shard` construction remains valid because the field is optional with a default.
- With publication disabled, the normal path does not make a digest pass over model tensors.

## Scope

Included:

- digest generation;
- optional wire-format field;
- propagation through discovery and planner inputs;
- Megatron publication sites;
- shard-region extraction for a later comparison;
- environment-variable and compatibility documentation.

Not included:

- receiver-side recomputation and comparison;
- metadata freshness checks;
- retry or failure policy after a mismatch;
- performance claims for digest-enabled runs.

The receiver comparison must refresh discovery first. Comparing current weights against metadata retained from an earlier step would report normal training updates as corruption. PR #579 adds the training-step stamp needed to make that freshness decision explicit.

## How to review

Review the ten files in this order:

1. **`refit/reshard/verify.py` — required algorithm review.** Check position sensitivity, integer-overflow assumptions, ragged-byte handling, row granularity, device/host synchronization cost, and rank validation in `shard_region`.
2. **`tests/test_reshard_refit_verify.py` — required contract review.** Confirm the tests distinguish a position-sensitive digest from a plain checksum and cover the optional wire field in both directions.
3. **`refit/reshard/rendezvous.py` — required wire-format review.** Check that an absent digest remains absent on encode and becomes `None` on decode, then verify that `build_sources` carries it into the planner.
4. **`refit/reshard/megatron_aliases.py` and `megatron_publisher.py` — publication review.** Confirm each digest covers the same address range and shape that the publisher advertises.
5. **`refit/reshard/slice_plan.py`, `envs.py`, and `__init__.py` — mechanical review.** Optional field, flag registration, and exports.
6. **`README.md` and `refit/README.md` — documentation review.** Confirm that they describe producer support without claiming the receiver gate already exists.

Questions for reviewers:

- Does the reduction detect the planning mistakes that matter without making qualification runs impractical?
- Is 4 KiB the right detection granularity, or should a later implementation use a stronger GPU-side hash?
- Do all publication sites digest exactly the memory region represented by their shard metadata?
- Is keeping digest publication opt-in the right default until its cost is measured on the target model?

## Validation

- The focused verification suite covers stable bytes, value changes, row permutations, ragged tails, shard-region extraction, rank mismatch, optional publication, and wire round trips.
- Existing reshard tests that construct `Shard` without a digest continue to pass.

## Follow-up and merge relationship

PR #579 should land before the receiver-side comparison. It provides the publisher training-step stamp needed to distinguish stale expectations from incorrect bytes. If this PR lands after #553, the `envs.py` and README environment-variable blocks will need a small mechanical conflict resolution; the implementations do not otherwise depend on one another.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/verify.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/verify.py
@@ -69,14 +69,28 @@ def tensor_digest(tensor) -> str:
     while the reduction itself stays one cheap kernel over the full tensor.
 
     The tensor is made contiguous first, since a strided view has no meaningful byte
-    order to digest. Returns a hex string so it can ride in JSON alongside the shard
-    table.
+    order to digest, and is rebased to a word-aligned start. Returns a hex string so it
+    can ride in JSON alongside the shard table.
     """
     import torch
 
     flat = tensor.detach().reshape(-1)
     if not flat.is_contiguous():
         flat = flat.contiguous()
+    elif (flat.storage_offset() * flat.element_size()) % 4:
+        # What a publisher actually passes. ``narrow`` returns a view that is
+        # contiguous *and* offset, so the branch above never fires, and
+        # ``.contiguous()`` would return self anyway - only a copy rebases the start.
+        # Without this the int32 view below raises on any shard whose byte offset is
+        # not a multiple of 4, which is every odd bf16 split: the gate/up narrow at an
+        # odd half, the QKV narrow at an odd row.
+        #
+        # Rebasing rather than skipping the leading bytes to reach an aligned
+        # boundary, which would be cheaper and wrong: the digest has to be a function
+        # of the shard's contents alone. A publisher digesting at offset 2818 and a
+        # receiver digesting the same bytes at offset 0 must agree, and they only do
+        # if neither one's result depends on where the bytes happen to sit.
+        flat = flat.clone()
     raw = flat.view(torch.uint8)
     n = int(raw.numel())
 
--- a/modelexpress_client/python/modelexpress/refit/reshard/verify.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/verify.py
@@ -134,9 +134,19 @@ def shard_region(full_tensor, global_shape, shard_offset, shape):
     Here rather than with the comparison gate because it is the inverse of what a
     publisher digests: the publisher hashes its own contiguous shard, and this is how
     a receiver recovers the same box out of the buffer it assembled.
+
+    The three ranks must agree. A short ``shard_offset`` would otherwise stop the
+    narrow loop early and return a larger region than the publisher digested, which
+    surfaces as a mismatch on a shard that transferred correctly - the most expensive
+    kind of wrong answer for a gate whose whole job is to be trusted.
     """
+    if not len(global_shape) == len(shard_offset) == len(shape):
+        raise ValueError(
+            f"rank mismatch: global_shape {tuple(global_shape)}, shard_offset "
+            f"{tuple(shard_offset)} and shape {tuple(shape)} must have equal rank"
+        )
     view = full_tensor.reshape(tuple(global_shape))
-    for axis, (start, extent) in enumerate(zip(shard_offset, shape)):
+    for axis, (start, extent) in enumerate(zip(shard_offset, shape, strict=True)):
         view = view.narrow(axis, int(start), int(extent))
     return view.contiguous()
 
--- a/modelexpress_client/python/modelexpress/refit/reshard/verify.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/verify.py
@@ -24,18 +24,21 @@ costs a pass over every published tensor.
 
 Scope, because it is narrower than "the refit is verified" and the difference matters
 when planning the consumer. A publisher digest covers a whole shard, so a receiver can
-only check it where it holds that whole shard - the full-pull path, which stages the
-entire logical source tensor contiguously and can recover any publisher's box with
-:func:`shard_region`. On the exact-segment path the receiver reads a sub-range straight
-into the live parameter and never materialises the shard, so there is nothing to
-recompute the digest over. That is not a corner case: where trainer and inference
-parallelism differ, most sources take the exact path. For strict qualification, the
-receiver must full-pull and verify the publisher's whole shard, derive the expected
-destination slice from that verified staging buffer, and compare it with the destination
-produced by the ordinary exact-segment read. Both reads must target the same immutable,
-step-stamped source version. A fingerprint over the contributing publishers is useful
-freshness evidence, but it is not a byte-for-byte expectation for a partial read and must
-not be presented as one.
+compare it directly only when the ordinary plan gives that receiver every byte of the
+shard. That includes a coarser inference topology which assembles whole trainer shards
+into known destination sub-boxes. A finer inference topology instead reads only part of
+each trainer shard, so its partial destination has no value comparable to the published
+whole-shard digest.
+
+For strict qualification of that partial-read case, the receiver must full-pull and
+verify the publisher's whole shard, derive the expected destination slice from the
+verified staging buffer, and compare it with the destination produced by the ordinary
+exact-segment read. Both reads must target the same immutable, step-stamped source
+version. :func:`shard_region` serves the full-pull reference path; the direct case needs
+the transfer plan's source-to-destination provenance to locate each complete publisher
+shard in the assembled destination. A fingerprint over the contributing publishers is
+useful freshness evidence, but it is not a byte-for-byte expectation for a partial read
+and must not be presented as one.
 """
 
 from __future__ import annotations
--- a/modelexpress_client/python/modelexpress/refit/reshard/verify.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/verify.py
@@ -1,0 +1,154 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Shard digests for the reshard refit path.
+
+A refit that reports plausible timings can still install the wrong bytes: the wire
+can deliver correctly while the *plan* points a copy at the wrong sub-box, and
+generation-level metrics such as KL against a reference are too coarse to localise
+that. This module supplies the cheap half of the answer - a digest a publisher can
+compute over the shard it owns, and which a receiver can recompute over the bytes it
+actually received - so a mismatch names the offending tensor instead of showing up
+as a slightly worse loss curve three steps later.
+
+Why not a byte hash: hashing 30 GB per rank per refit on the GPU means either a
+device-side hash kernel we do not have, or a copy to host we cannot afford. The
+digest below is a *position-sensitive* reduction instead - see :func:`tensor_digest`
+for why plain sums are not enough.
+
+This module is the digest and the wire format for it. The receiver-side gate that
+recomputes and compares is deliberately not here; it needs the fresh-discovery
+refresh to avoid reporting ordinary training updates as corruption, which is its own
+change. Publishing is opt-in via ``MX_RESHARD_PUBLISH_DIGEST`` because the reduction
+costs a pass over every published tensor.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import logging
+
+logger = logging.getLogger(__name__)
+
+# Words per digest row. Row sums are position-sensitive *between* rows, since the row
+# vector is hashed in order, so this is the granularity at which a permutation is
+# detected: reordering whole 4 KiB rows is caught, reordering words inside one row is
+# not. 1024 int32 words keeps the hashed vector ~4 KiB per MiB of tensor, small enough
+# to move to host, while staying coarse enough that the reduction is a single fast
+# kernel.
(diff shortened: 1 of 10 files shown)


modelexpress_client/python/modelexpress/refit/reshard/verify.py
```diff
@@ -0,0 +1,154 @@
...
+    and the box is generally strided inside the full tensor.
+
+    Here rather than with the comparison gate because it is the inverse of what a
+    publisher digests: the publisher hashes its own contiguous shard, and this is how
+    a receiver recovers the same box out of the buffer it assembled.
+    """
+    view = full_tensor.reshape(tuple(global_shape))
+    for axis, (start, extent) in enumerate(zip(shard_offset, shape)):
+        view = view.narrow(axis, int(start), int(extent))
+    return view.contiguous()
```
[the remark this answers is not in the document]
KavinKrishnan: Good catch, fixed in 181f583.

This one mattered more than the lint hint suggests. A short coordinate stopped the narrow loop early and returned a *larger* region than the publisher digested, so the gate would report a mismatch on a shard that transferred perfectly well. For a check whose entire value is being trusted when it fires, a false positive is the expensive failure — we already lost a run's worth of time to exactly that shape of thing on Topology B.

Now validates that `global_shape`, `shard_offset` and `shape` agree in rank, with `strict=True` so a future edit cannot quietly reintroduce the truncation. Three parametrized cases cover a short offset, a short shape, and a short pair against a higher global rank; all three fail without the fix.

modelexpress_client/python/modelexpress/refit/reshard/verify.py
```diff
@@ -0,0 +1,164 @@
...
+
+    The tensor is made contiguous first, since a strided view has no meaningful byte
+    order to digest. Returns a hex string so it can ride in JSON alongside the shard
+    table.
+    """
+    import torch
+
+    flat = tensor.detach().reshape(-1)
+    if not flat.is_contiguous():
+        flat = flat.contiguous()
```
yafshar, line 79:
```suggestion
    if not flat.is_contiguous():
        flat = flat.contiguous()
    elif (flat.storage_offset() * flat.element_size()) % 4:
        # narrow yields a contiguous view at a nonzero offset, so .contiguous()
        # is a no-op and cannot rebase it; only a copy can.
        flat = flat.clone()
```

`published_digest` raises on real Megatron geometry once `MX_RESHARD_PUBLISH_DIGEST=1`.

`narrow` returns a contiguous view with a nonzero storage offset. Line 78 is therefore `False`, thus `.contiguous()` never runs, and line 86 throws:

```
RuntimeError: self.storage_offset() must be divisible by 4 to view Byte as Int
(different element sizes), but got 2818
```

Repro'd geometries:

| tensor | result |
|---|---|
| 1-D bf16 gated bias, `half=1409` (odd) | FAIL, offset 2818 |
| 1-D bf16 gated bias, `half=1408` (even) | OK |
| 2-D fp8 gated, hidden=6, `half=17` | FAIL, offset 102 |
| 2-D bf16 gated, hidden=4096 | OK |

Callers producing these views: `megatron_aliases.py:83` and `:178`, both `published_digest(tensor)` on a narrow. `_build_gated_aliases` narrows at `half`; `_build_qkv_aliases` narrows at `q_rows`/`head_dim`. Anything where `narrow_offset_elems * itemsize % 4 != 0` — every odd 1-byte offset, every odd bf16/fp16 offset.

Worth noting why widening the line-78 condition alone is not enough: `.contiguous()` on an already-contiguous tensor returns `self`, so it detects the case and then does nothing about it. Only a copy rebases the offset, hence the separate `elif` branch rather than an `or`.

Why the current tests miss it: `test_digest_handles_a_strided_input` (`test_reshard_refit_verify.py:82`) uses `dense[:, ::2]`, which is non-contiguous, so line 79 does run. `test_digest_covers_a_non_word_multiple_tail` (`:70`) uses a fresh `arange` at offset 0. No test constructs a contiguous view at a nonzero offset, which is the only shape publishers actually pass. Worth a case that does:

```python
def test_digest_handles_a_contiguous_view_at_an_unaligned_offset():
    """What a publisher actually passes. ``narrow`` is contiguous *and* offset, so
    the non-contiguous branch never fires and the int32 view has to be given an
    aligned base of its own."""
    fused = torch.arange(2 * 1409, dtype=torch.bfloat16)
    half = fused.narrow(0, 1409, 1409)

    assert half.is_contiguous() and half.storage_offset() % 2, "precondition"
    assert tensor_digest(half) == tensor_digest(half.clone())
```

yafshar (changes requested):

modelexpress_client/python/modelexpress/refit/reshard/verify.py
KavinKrishnan: Good catch, and confirmed — reproduced it at exactly your offset 2818. Fixed with your suggestion.

One thing I expanded in the comment, because it's the part that would tempt someone to "optimize" this later: the copy is doing more than satisfying the int32 view. Skipping the leading bytes forward to the next 4-byte boundary would be cheaper and would quietly break the feature, because the digest would then depend on where the bytes happen to sit in memory. A publisher hashes its shard at whatever offset the fused parent gives it, and the receiver hashes the same bytes at offset 0 in its own staging buffer. Those two only agree if neither answer depends on the offset, so this has to rebase rather than skip.

Took your test and added the odd fp8 case alongside it, plus one that pins the property directly: the same bytes at offset 0, at an odd offset, and at an even offset all have to digest the same. All three fail against the old code and pass now.

yafshar (approved):
LGTM!

modelexpress_client/python/modelexpress/refit/reshard/verify.py
```diff
@@ -0,0 +1,178 @@
...
+        scalars.append(int(raw[body:].to(torch.int64).sum().item()))
+        scalars.append(n % 4)
+
+    h = hashlib.blake2b(digest_size=16)
+    for value in scalars:
+        h.update((value & _U64).to_bytes(8, "little"))
+    return h.hexdigest()
+
+
+def shard_region(full_tensor, global_shape, shard_offset, shape):
```
tanushriya910, line 140: Since the receiver part isn't there yet, just checking if the overall verification would work or not with this. Based on my understanding this shard_region is going be used on receiver side to get the exact trainer published box from the full pulled tensor, but how does it work in say the following scenario:

trainer TP=4 and inference TP=8 and for a tensor sharded along the row axis, inference pulls half the rows the trainer has. So the slice plan issues a NIXL READ for only that subset (not the full shard), then:

-> Publisher digested rows 0–1023
-> Receiver pulled rows 0–511
-> There's no staging buffer holding the full trainer shard to run shard_region against

doesn't the comparison break in that case?

KavinKrishnan: You're right to push on this. My first reply identified the limitation, but it
stopped before answering how verification should work for the resharded read and
then blurred publisher freshness evidence with byte equality. The precise answer
is:

**The whole-shard digest in this PR cannot directly verify your partial-read
example.** If trainer rank 0 publishes rows 0–1023 with digest `D(T0)`, and
inference rank 0 reads only rows 0–511, then `D(rows 0–511)` is not comparable to
`D(T0)`. The current digest is deliberately not range-composable. A destination
digest or a fingerprint over the contributing publishers does not manufacture
that missing expected value either.

For strict reshard qualification, the receiver needs two paths over the **same
immutable source version**:

```text
trainer T0, step N, rows 0–1023, published digest D(T0)
                  |
          +-------+----------------------+
          |                              |
 ordinary exact path              reference path
 READ rows 0–511                  full-pull source at step N
          |                              |
          v                              v
 candidate destination            verify whole T0 against D(T0)
 rows 0–511                              |
                                         v
                                  replay the transfer plan /
                                  slice rows 0–511 locally
                                         |
                                         v
                                  expected destination

             candidate destination == expected destination
                              |
                       yes -> install
                       no  -> fail before install
```

That is how the TP=4 → TP=8 case is checked. `shard_region` belongs only to the
reference side: it extracts each publisher-owned box from the full-pull staging
buffer so we can first prove that staging contains what the publisher advertised.
The receiver then replays the *actual exact plan* against those verified staged
bytes to create the expected TP=8 destination slice. Finally it compares that
with the buffer produced by the ordinary exact NIXL READ.

The step stamp is necessary but not sufficient: the trainer must keep the
published step's buffers immutable (or leased) until both reads finish. Otherwise
the exact and reference reads can straddle an optimizer update and disagree even
when both transfers are correct.

So the contracts should be:

* full-pull source available + whole-shard digest matches + exact replay matches:
  strict evidence for this destination;
* only an exact partial read, with no verified reference materialised: **no
  evidence**, never a pass;
* lower-overhead production verification would need a different wire contract,
  such as per-range/chunk digests or source-computed digests for the regions in
  the completed transfer plan. A whole-shard digest cannot be sliced after the
  fact.

#591 supplies the publisher truth needed for the first path, but it is not by
itself the overall reshard verification gate. I updated the module documentation
to state this exact two-path contract and removed the implication that a
publisher fingerprint can serve as the expected bytes for a partial read.

tanushriya910: Fine with this as a first pass, as long as the follow-up verification covers the coarser-inference-TP case (e.g. trainer TP=4, inference TP=2). There the receiver holds whole trainer shards even on the normal sliced path, so it should be verifiable with a direct digest recompute rather than needing the full-pull reference path.

tanushriya910 (approved):

modelexpress_client/python/modelexpress/refit/reshard/verify.py
KavinKrishnan: Agreed. The follow-up should branch on the actual source overlap, not simply on
whether the planner labelled the operation exact or full-pull:

- **Trainer TP=4 -> inference TP=2:** each contributing trainer shard is present
  in full inside a known destination sub-box. Use the transfer plan's
  source-to-destination provenance to extract that complete sub-box from the
  ordinary destination buffer, recompute its digest, and compare directly with
  the publisher digest. No second/full-pull reference read is needed.
- **Trainer TP=4 -> inference TP=8:** each receiver has only part of a published
  shard. A whole-shard digest cannot be compared directly, so strict
  qualification needs the verified full-pull reference plus exact-plan replay
  described above.
- If neither a complete shard nor a verified reference was materialised, report
  `NO_EVIDENCE`, never pass.

I updated the module contract to state this topology split explicitly. That
keeps the cheap direct check for coarser inference TP and confines the expensive
dual-path check to finer inference TP where it is actually necessary.

Issue #569, opened 2026-07-28
feat: integrate server-side model-cache fallback with P2P loaders


zhengluo-nv:
## Summary

ModelExpress 0.4.1 implements P2P weight transfer and server-side model caching as separate flows. Engine-native ModelExpress loaders do not currently combine them into one fallback strategy.

For example, a vLLM worker started with `--load-format mx` can register `MxModelLoader`, find no P2P source, then check its empty local Hugging Face cache and fail under `HF_HUB_OFFLINE=1`. It does not ask ModelExpress Server to fetch and stream the model first.

The intended cold-miss flow is:

1. Try a compatible P2P source.
2. If none is available, ask ModelExpress Server for the model.
3. On a server cache miss, the server downloads and caches the complete model from Hugging Face.
4. The server streams the complete model files to the worker.
5. The worker installs a valid local Hugging Face snapshot, loads it, becomes ready, and can serve as a P2P source.

The worker must not require direct Hugging Face access.

## Current behavior

With an empty worker cache, no P2P source, and `HF_HUB_OFFLINE=1`, the vLLM MX loader fails while resolving the local snapshot:

```text
huggingface_hub.errors.LocalEntryNotFoundError:
Cannot find an appropriate cached snapshot folder for the specified revision
on the local disk and outgoing traffic has been disabled.
```

The direct `--load-format mx` cold-miss path does not automatically invoke the server-side cache flow.

## Proposed behavior

Add the server-side model cache as a fallback in the engine-native ModelExpress load strategy:

```text
P2P source -> ModelExpress Server cache/download -> existing local/native fallback
```

This should cover:

- vLLM `--load-format mx` / `modelexpress`
- SGLang `remote_instance` with the ModelExpress backend
- model configuration, tokenizer, weight, and other required repository files—not weights alone
- pinned revisions and private Hugging Face repositories
- atomic installation of a complete Hugging Face-compatible worker snapshot

The SGLang path must account for configuration/tokenizer resolution that may happen before the weight-loader callback.

## Configuration and compatibility

The server fallback should be opt-in through the existing no-shared-storage configuration, for example:

```yaml
MODEL_EXPRESS_URL: <model-express-server-address>
MODEL_EXPRESS_NO_SHARED_STORAGE: "1"
MODEL_EXPRESS_CACHE_DIRECTORY: /home/dynamo/.cache/huggingface/hub
HF_HUB_CACHE: /home/dynamo/.cache/huggingface/hub
[...]

PR #592, opened 2026-08-05
feat(client): load models from the server cache when workers have no shared storage

scydas:
## What

Part of #569 — deliberately not a full close. The issue lists five things the fallback should cover:

| Coverage required by #569 | This PR |
|---|---|
| vLLM `--load-format mx` / `modelexpress` | done |
| SGLang `remote_instance` with the ModelExpress backend | partial, see below |
| config, tokenizer, weights and other repository files, not weights alone | done |
| pinned revisions and private Hugging Face repositories | **not done**, see below |
| atomic installation of a complete HF-compatible worker snapshot | done |

Lets a vLLM worker get both the model metadata and the weights from the ModelExpress Server, so the fallback chain becomes **P2P source → server cache/download → existing local/native fallback** and a worker never needs direct Hugging Face access.

The metadata and the weights are fetched at different times, because the engine asks for them at different times:

| When | Trigger | Needs |
|---|---|---|
| T0 | `EngineArgs.__post_init__` → `get_model_path` → `snapshot_download` | config/tokenizer + a resolvable HF cache layout |
| T1 | tokenizer/processor load | metadata |
| T2 | `MxModelLoader.load_model` → `LoadStrategyChain` | weights |

Only T2 reaches the strategy chain. T0 happens while engine args are still being parsed — before any loader object exists — so a strategy-chain-only implementation cannot serve it. P2P does not help either: it transfers GPU tensors and never repository files, so **every** worker needs local metadata regardless of whether P2P later hits.

Hence the split:

- **Metadata** is pulled at T0, unconditionally, by patching `snapshot_download`. No weight moves on this path, so P2P keeps first refusal on the part that matters.
- **Weights** stay in the strategy chain. `ServerCacheStrategy` is inserted *after* `RdmaStrategy`, so it only runs on a P2P miss.

New modules (all under `modelexpress_client/python/modelexpress/`):

- `model_client.py` — `ModelCacheClient`, wrapping `EnsureModelDownloaded` / `ListModelFiles` / `StreamModelFiles`, with stream-protocol validation (chunk offset continuity, `is_last_file` semantics, commit-hash agreement) and two install entry points: metadata-only and weights-only.
- `model_snapshot.py` — the Hugging Face cache layout: `models--<org>--<name>/{refs/main, snapshots/<commit>/}`, `flock`-guarded staging, per-file rename on publish, and `SnapshotPatch.rollback()` so an interrupted weight install leaves no partial files.
- `model_prefetch.py` — T0 orchestration, the enable check, and repo-id recovery from a cache path.
- `load_strategy/server_cache_strategy.py` — the strategy itself; `requires = (EngineAdapter.load_via_native,)`.
- `engines/vllm/patches/patch_hf_snapshot_prefetch.py` — takes over `snapshot_download` across `huggingface_hub._snapshot_download` / `hf_api` / `huggingface_hub`; a prefetch error is logged as a warning and delegates to the original, never blocking startup.

Also:

- `generate_proto.sh` now generates `model_pb2` as well as `p2p_pb2` (the Python client previously did not consume `ModelService` at all).
- Three env vars: `MODEL_EXPRESS_NO_SHARED_STORAGE` (the switch), `MODEL_EXPRESS_CACHE_DIRECTORY`, `MODEL_EXPRESS_TRANSFER_CHUNK_SIZE`.
- `engines/vllm/loader.py`: `download_model` returns early when the feature is on, since the snapshot is already in place.
- Docs: `docs/ARCHITECTURE.md`, `docs/DEPLOYMENT.md`.

Everything is behind `MODEL_EXPRESS_NO_SHARED_STORAGE`; with the switch off, none of this code runs (verified on hardware, see below).

`model_snapshot.py` carries its own weight-suffix list mirroring `providers.rs`, because the split has to happen client-side: the metadata phase needs everything *except* weights and the weight phase needs the inverse, and only the former is expressible server-side. #581 has since landed `is_weight_file` plus an `ignore_weights` flag on `ModelFilesRequest`, so the metadata half could stop duplicating the list — worth a follow-up issue rather than widening this PR into a filtering-protocol change.

SGLang is partly covered already: its `nixl` transport runs the same `LoadStrategyChain`, and `SglangAdapter` implements `load_via_native`, so `ServerCacheStrategy` is live there with no extra code. Two gaps remain for a separate PR — the `transfer_engine` transport falls straight to a native load without consulting the chain, and SGLang has no `vllm.general_plugins` equivalent, so T0 needs an explicit entry point.

The full list of requirements and operational limits lives in `docs/DEPLOYMENT.md` (Server-Backed Model Cache), and the phase split and stream/rollback contract in `docs/ARCHITECTURE.md`. Repeating the three that need a reviewer's judgement rather than a reader's awareness — the first two are the open half of #569's fourth bullet:

- **Pinned revisions are not supported.** The model RPC has no revision field, so the server always answers with what its cache holds for `main`. Reproduced on hardware: DGD pinned `a338b55`, server had `7ae5576` — the client warns and refuses to mix revisions rather than silently serving the wrong one. Fixing this needs a proto change plus a registry key change, which is why it is not folded in here.
- **Private repositories are untested end to end.** The server already supports `HF_TOKEN` and the worker never talks to Hugging Face, so the path should work by construction, but it has not been exercised on hardware and is not claimed.
- **The weight stream is not deduped across pods.** The server-side claim dedups the upstream fetch (experiments 6A/6B below), but each pod still streams its own copy, so N replicas cold-starting cost N x model size in server egress.

> Note for reviewers of #581: its description states "the Python client is metadata-only and does not consume `ModelService`". That is no longer true after this PR — the Python client now calls all three `ModelService` RPCs.

## Why

Without this, a worker with no shared storage still has to reach Hugging Face itself for config and tokenizer even when the ModelExpress Server already holds the model, and it has no way to get weights from the server when P2P misses (cold start, first replica, no rank match). The server cache existed but nothing in the Python engine path could use it.

The design constraint that shaped this is that **P2P and repository files are disjoint**. It is tempting to make the whole thing one strategy, but the engine resolves the model path long before any strategy runs, and it fails at `parse_args` with `LocalEntryNotFoundError` if the cache layout is not already valid. So the metadata half has to be earlier, and it has to be unconditional.

## Testing

Unit: 1205 passed / 46 skipped for the full Python suite. New test files cover the client's stream protocol, the snapshot layout and rollback, the prefetch, the `snapshot_download` patch, and the strategy.

On real hardware — 2 nodes, 1× A800 80GB each, vLLM via Dynamo DGD, ModelExpress Server 0.4.0 with an internal HF mirror. Nine scenarios, every one re-run against this branch's single commit after the review fixes were folded in. "upstream" counts Hugging Face fetches by the server; "streams" counts `StreamModelFiles` calls served from the server's own cache, split into the metadata phase and the weight phase.

| # | Scenario | upstream | metadata streams | weight streams | Result |
|---|---|---|---|---|---|
| 1 | Server cache cold | 1 | 1 | 1 | `COMPLETE in 11.83s`, READY |
| 2 | Server cache warm | 0 | 1 | 1 | metadata 0.157s |
| 3 | **P2P available** | 0 | 1 | **0** | RDMA 0.99 GB / 86ms / **92.5 Gbps**, `COMPLETE in 1.75s` |
| 4 | Switch off | 0 | 0 | 0 | Server delta log 0 lines, none of this code executes |
| 5 | 4 processes, same pod | 0 | 1 | 1 | 4/4 ok; 8 calls collapsed to 2 streams by `flock`, zero leftovers |
| 6A | 4 independent pods | 1 | 4 | 4 | 4/4 ok in 37s; claim dedups upstream, streams are per-pod |
| 6B | DGD 2 replicas, simultaneous cold start | 1 | 2 | 2 | both released at the same instant, 12.48s / 9.19s |
| 7 | Server unreachable | 0 | 0 | 0 | degrades in 20.1s, no half-written files, failure semantics unchanged |
| 8 | 7B / 4 shards / 15.24 GB | 1 | 1 (8 files) | 1 (4 files) | metadata 0.17s vs weights 114.27s, all shards present |
| 9 | Server killed mid-transfer | 1 | 1 | 1, aborted | weights fully rolled back, metadata kept, retry resumes |

Rows 6A and 6B show the dedup boundary: one upstream fetch no matter how many workers race, but one stream per worker in each phase. That is the cross-pod limit noted above, measured.

Row 3 is the one that carries the design claim. `server-cache` appears in `Eligible loaders`, so it *could* have run, but `rdma` is tried first and wins — and the server's own log for that window shows exactly one stream, `Found 7 files`, the metadata set. Not one weight byte came from the server while a P2P source was alive. Fetching metadata early does not weaken P2P-first.

### Cross-process state loss

`server-cache` initially filtered itself out with `No Hugging Face repo id for '/home/dynamo/.cache/.../snapshots/7ae5576...', skipping server cache`. vLLM rewrites `ModelConfig.model` in place with the resolved path *and* loads weights in a separate EngineCore process, so the in-process `path → repo id` map recorded at T0 is empty where the strategy runs. Unit tests could not catch this because they run in one process. Fixed by deriving the repo id deterministically from the `models--<org>--<name>` directory name; `TestRepoIdFromCachePath` now pins that behaviour with the reasoning in its docstring.

## Summary by CodeRabbit

* **New Features**
  * Added server-backed model caching for environments without shared storage.
  * Model metadata and weight files can now be fetched, validated, and installed locally.
  * Added prefetch support for Hugging Face model snapshots.
  * Added the public `ModelCacheClient` for managing cached models.
  * Added configuration options for cache location, storage mode, and transfer chunk size.

* **Documentation**
  * Documented server-backed caching, deployment configuration, fallback behavior, and operational considerations.

* **Tests**
  * Added extensive coverage for caching, streaming validation, snapshot management, prefetching, and loader integration.


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/592/commits/dd3fcc2f5216372d8c0034d1f8e68902f2014253

modelexpress_client/python/modelexpress/model_snapshot.py
```diff
@@ -0,0 +1,443 @@
...
+        if self._cache.has_files(snapshot_path, expected_files):
+            logger.info(
+                "Snapshot %s already complete, discarding staged copy", snapshot_path
+            )
+            shutil.rmtree(staging_path, ignore_errors=True)
+            self._staging_path = None
+            self._cache.write_main_ref(commit_hash)
+            return snapshot_path
+
+        stale_path: Path | None = None
```
AndyDai-nv, line 272: When install_metadata_snapshot falls through to staging, publish() checks has_files(snapshot_path, expected_files) against the metadata set only. If that check fails while snapshots/<commit>/ already holds a complete weight set, the whole directory is moved to .modelexpress-stale-* and rmtree'd - the cached weights are deleted to install metadata.

The commit hash here is necessarily the same as the one refs/main points at (both come from the server's resolution of main), so this is an in-place overwrite, not the safe "new revision, new directory" case. Trigger: any state where the local metadata set doesn't match the server manifest at the same commit - e.g. the mirror gains a chat_template.jinja, or one metadata file's size changes.

scydas: Confirmed and fixed in df8b51b. You're right that the same-commit case makes this an in-place overwrite: the manifest covers non-weight files only, so the weight set sat outside every check but inside the directory being replaced.

publish() now merges into an existing snapshots/<commit>/ one atomic rename at a time instead of swapping the directory. Same commit means same content, so the merge only ever adds files. The wholesale-replace path is kept for the case where the directory does not exist yet.

Added test_republish_keeps_files_the_manifest_does_not_mention, which reproduces the loss (a mirror gaining chat_template.jinja while a weight set is already installed) and fails against the previous code.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/592/commits/422df8bea1afb5f791599ad661b3646adb61eaf3

AndyDai-nv (approved):
Thanks @scydas, lgtm, need your quick look @zhengluo-nv  as well to see whether it is good to close 569

modelexpress_client/python/modelexpress/model_client.py
```diff
@@ -0,0 +1,393 @@
...
+        metadata_paths, _ = split_by_weight(manifest.keys())
+        if not metadata_paths:
+            raise ModelCacheError(
+                f"ModelExpress returned no non-weight files for {model_name}"
+            )
+        expected = {path: manifest[path] for path in metadata_paths}
+
+        cache = ModelSnapshotCache(model_name, self.cache_directory)
+        with cache.lock():
+            existing = cache.resolve_snapshot(expected)
```
zhengluo-nv: [P1] Verify the cached snapshot revision before reusing it

`resolve_snapshot` only checks the paths and sizes from `ListModelFiles`, but that response carries no commit identity. If the server advances `main` and the new revision has the same file names and sizes, this returns the old local snapshot without opening `StreamModelFiles`, so the later commit-hash validation never runs. I reproduced a local `aaaa...` snapshot being accepted while the server stream represented `bbbb...`, with zero stream calls. Please make the server revision part of the manifest/reuse contract and fail closed unless it matches the snapshot directory or `refs/main`.

modelexpress_client/python/modelexpress/model_client.py
```diff
@@ -0,0 +1,393 @@
...
+
+        Always requests weights: a metadata-only download would register the
+        model as complete on the server and stop any later weight fetch.
+        """
+        request = model_pb2.ModelDownloadRequest(
+            model_name=model_name,
+            provider=provider,
+            ignore_weights=False,
+        )
+        for update in self.stub.EnsureModelDownloaded(request):
```
zhengluo-nv, line 143: [P2] Preserve P2P-first when the server cache is cold

This metadata-phase call always sends `ignore_weights=False`, so before the strategy chain can try RDMA, a cold ModelExpress Server downloads the complete weight set. A live P2P source avoids the later weight stream but does not avoid this upstream weight download or its startup delay. The hardware result in the PR covers P2P with a warm server, not this cold-server case. Please provide a metadata-only server claim that does not mark the full model complete, or otherwise defer the full server download until after the P2P miss.

modelexpress_client/python/modelexpress/model_snapshot.py
```diff
@@ -0,0 +1,479 @@
...
+
+    def rollback(self) -> None:
+        """Undo this patch, leaving the snapshot as it was before it started.
+
+        A half-applied patch is worse than none: the engine would see a subset
+        of the weights and load it as if it were complete.
+        """
+        self.close()
+        while self._published:
+            self._published.pop().unlink(missing_ok=True)
```
zhengluo-nv: [P2] Restore overwritten shards during rollback

`_finalize` replaces an existing target, but `_published` records only the target path; rollback therefore unlinks the newly published file without restoring the previous one. If a refresh publishes shard A and then fails while receiving shard B, shard A from the previously complete snapshot is deleted. I reproduced this with two existing safetensors: after the interrupted patch, only the untouched second shard remained. Please move existing targets aside and restore them on rollback, while cleaning the backups only after the whole patch succeeds.

docs/DEPLOYMENT.md
```diff
@@ -723,6 +723,37 @@ artifact transfer only within a trusted deployment, and network-isolate the MX
...
+MODEL_EXPRESS_CACHE_DIRECTORY: /home/dynamo/.cache/huggingface/hub
+HF_HUB_CACHE: /home/dynamo/.cache/huggingface/hub
+HF_HUB_OFFLINE: "1"
+```
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `MODEL_EXPRESS_NO_SHARED_STORAGE` | `0` | Fetch repository files from ModelExpress Server. When unset, nothing changes: no extra RPCs and no change to P2P or local loading. |
+| `MODEL_EXPRESS_CACHE_DIRECTORY` | `HF_HUB_CACHE` | Where the worker installs snapshots. Point it at the same path as `HF_HUB_CACHE` so the engine reads what ModelExpress wrote. |
+| `MODEL_EXPRESS_TRANSFER_CHUNK_SIZE` | `1048576` | gRPC file-stream chunk size in bytes. |
```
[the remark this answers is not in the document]
scydas: Fixed in 422df8b. Both address variables now have their own rows in the Server-Backed Model Cache table, right after the switch:

| `MODEL_EXPRESS_URL` | unset | ModelExpress Server address. Required together with the switch above; without an address the feature stays off. |
| `MX_SERVER_ADDRESS` | unset | Alternative spelling of the server address, accepted for parity with the P2P client. Either variable satisfies the requirement. |

The `MODEL_EXPRESS_TRANSFER_CHUNK_SIZE` row is unchanged; its `1048576` default matches `DEFAULT_CHUNK_SIZE`, as you confirmed.

modelexpress_client/python/modelexpress/model_client.py
```diff
@@ -0,0 +1,384 @@
...
+        self.close()
+
+    @property
+    def stub(self) -> model_pb2_grpc.ModelServiceStub:
+        """Return (and lazily create) the model-service stub."""
+        if self._stub is None:
+            options = [
+                ("grpc.max_send_message_length", self._max_message_size),
+                ("grpc.max_receive_message_length", self._max_message_size),
+            ]
```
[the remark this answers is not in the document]
scydas: Fixed in 422df8b. The channel now carries keepalive, with the reasoning recorded next to the options:

```python
# No RPC here carries a deadline, because a cold-cache download
# is legitimately slow. Keepalive is what separates "slow" from
# "dead": without it a silently dropped connection blocks the
# engine's startup path forever, so the pod neither becomes
# ready nor crash-loops.
("grpc.keepalive_time_ms", 30_000),
("grpc.keepalive_timeout_ms", 10_000),
("grpc.keepalive_permit_without_calls", 1),
("grpc.http2.max_pings_without_data", 0),
```

`max_pings_without_data=0` is the one addition to your suggestion: without it gRPC stops pinging once a stream goes quiet, which is exactly the window this is meant to cover.

I did not add a deadline to the unary `ListModelFiles`. It walks the model directory server-side (`collect_model_files`), so its duration scales with shard count and with how cold the server's filesystem is; a fixed deadline would turn a slow-but-healthy server into a startup failure. The dead-peer case is covered by keepalive on that call as well.

modelexpress_client/python/modelexpress/model_snapshot.py
```diff
@@ -0,0 +1,443 @@
...
+                os.replace(stale_path, snapshot_path)
+            raise
+        self._staging_path = None
+
+        if stale_path is not None:
+            try:
+                shutil.rmtree(stale_path)
+            except OSError:
+                logger.warning("Failed to clean up stale snapshot %s", stale_path)
+        return snapshot_path
```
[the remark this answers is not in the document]
scydas: Fixed in 422df8b. The restore now runs on `BaseException`, with the leak you describe spelled out in the comment:

```python
except BaseException:
    # BaseException, not Exception: a KeyboardInterrupt here would
    # otherwise strand the moved-aside directory with no owner and no
    # cleanup path, leaking a partial model's worth of disk.
    if stale_path is not None and not snapshot_path.exists():
        os.replace(stale_path, snapshot_path)
    raise
```

modelexpress_client/python/tests/test_model_snapshot.py
```diff
@@ -0,0 +1,280 @@
...
+        assert leftovers == []
+
+    def test_discard_removes_staging(self, cache):
+        staging = cache.staging()
+        staging.begin_file("config.json")
+        staging.write(b"{}")
+        staging.discard()
+
+        assert not staging.path.exists() if staging._staging_path else True
+        assert not (cache.repo_root / "snapshots").exists()
```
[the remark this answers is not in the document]
scydas: Fixed in 422df8b. The staging path is captured before `discard()`, so the removal assertion actually runs:

```python
def test_discard_removes_staging(self, cache):
    staging = cache.staging()
    staging_path = staging.path
    staging.begin_file("config.json")
    staging.write(b"{}")
    staging.discard()

    assert not staging_path.exists()
    assert not (cache.repo_root / "snapshots").exists()
```

modelexpress_client/python/modelexpress/model_client.py
scydas: Fixed in `7d77085`. The metadata phase now sends `ignore_weights=true`.

The claim you asked for turned out to already exist upstream. #598 folded the weight mode into the registry entry key, and `registry/entry_key.rs` names this exact case — "a metadata-only download (`ignore_weights`) must not satisfy a later full-weight request, which would hand back a snapshot with no weights in it". So the client only had to stop asking for the full download; the weight phase claims its own key and gets its own.

The `ensure_downloaded` docstring that claimed the opposite is gone.

modelexpress_client/python/modelexpress/model_snapshot.py
scydas: Fixed in `c45294e`, following your prescription: `_finalize` moves an existing target aside before the rename, `rollback()` puts it back, and a new `commit()` drops the backups once the whole patch has landed.

`commit()` also clears the published list, so a rollback after a successful commit cannot delete the files it just installed.

Covered by a test that reproduces your case: two existing safetensors, replace one, roll back, and assert both the replaced shard and the untouched one still hold their original content.

modelexpress_client/python/modelexpress/model_client.py
scydas: Confirmed, and it still holds verbatim after the rebase onto #598: that PR put `revision` on the requests and `resolved_revision` on the download status, but `ModelFileList` still carries no commit, so the manifest cannot identify itself.

Fixed in `3caef7a`, following your prescription. The server revision comes from `resolved_revision` on the `EnsureModelDownloaded` status stream, and both of your anchors are in use — each phase gets the one that matches its ground truth:

- **Metadata**: `resolve_snapshot` refuses reuse unless the reported revision equals `refs/main`; no reported revision means no reuse. Fails closed both ways.
- **Weights**: the same hole existed before the `has_files` shortcut — weights left by an earlier revision satisfy a name-and-size comparison — so the reported revision is checked against the snapshot directory name there, the directory the engine has already resolved and is reading from.
- Every call after the first in a phase is pinned via `ModelFilesRequest.revision`, so the manifest and the stream cannot come from different commits — the window closes rather than merely being detected.
- Your repro is now a test: same file names and sizes, advanced `main` — the client restreams and publishes the new commit's directory instead of returning the old one.

One cost stated plainly: the server's already-downloaded path reports the revision the *request* carried, and these requests are unpinned, so a warm server names no revision and reuse never fires there — the metadata is restreamed each time (small, and the stream's first-chunk commit check still applies). `docs/ARCHITECTURE.md` promises the resolved commit is always reported, which that path does not honour; fixing that server-side would restore the reuse fast path with no client change. Happy to file it as a follow-up issue.

Related: with the metadata phase now sending `ignore_weights`, the manifest excludes weights entirely, so a revision that re-uploads only weights produces a byte-identical metadata manifest — your scenario got easier to hit, which is one more reason reuse fails closed.

zhengluo-nv:
/ok to test [645d969](https://github.com/ai-dynamo/modelexpress/pull/592/commits/645d969a925cfe9f7f66915140ce2c9064218abc)

zhengluo-nv:
Approving as my understanding is the server_cache_strategy is disabled by default

zhengluo-nv (approved):

PR #225, opened 2026-04-10
feat: add safety checks for P2P weight transfer

nicolasnoble:
Port of #199 (release/0.3.0) to main, adapted for the LoadStrategyChain architecture.

- MLA feature gate in RdmaStrategy.is_available() - blocks MLA models from P2P, detected via kv_lora_rank in HF config
- ManifestMismatchError for tensor name/size mismatches during transfer
- TransferFingerprint for runtime environment logging
- Replace MLA models with Llama-3.1-405B-Instruct in P2P examples
- MLA limitation documented in README, ARCHITECTURE.md, DEPLOYMENT.md
- Bypass with MX_SKIP_FEATURE_CHECK=1

## Summary by CodeRabbit

* **New Features**
  * Added automatic P2P transfer safety checks that block Multi-head Latent Attention (MLA) models from GPU-to-GPU transfer and fallback to disk loading.
  * Introduced manifest validation to detect tensor incompatibilities during transfer.
  * Added `MX_SKIP_FEATURE_CHECK` environment variable for testing.

* **Documentation**
  * Updated deployment and architecture documentation with transfer safety details and known MLA transfer issues.
  * Updated example configurations to use supported model architectures.

* **Tests**
  * Added comprehensive transfer safety validation tests.

PR #595, opened 2026-08-06
refactor(client): remove unused TransferFingerprint

AndyDai-nv:
## Summary

`TransferFingerprint` in `transfer_safety.py` is dead code. It was added in #225 with the documented intent that the source publish it alongside tensor metadata and the target compute its own and reject the transfer on mismatch. That wiring was never built.

Verified before removing:
- Nothing outside the class and its own unit tests references it — no importer in `publish`, `rdma_strategy`, or anywhere in the client.
- No proto field carries it, and the Rust side has no counterpart.
- It is not re-exported from `modelexpress/__init__.py`, so it is not public API.

Removed: the `TransferFingerprint` dataclass, its environment getters (`get_vllm_version`, `get_torch_version`, `get_cuda_version`, `get_deep_gemm_version`, `get_gpu_capability`, `_get_attention_backend`), and `_compute_manifest_hash`, plus the now-unused `hashlib` / `json` / `dataclasses` / `torch` / `envs` imports and the tests covering only those.

## Deliberately kept

`check_transfer_allowed()` and `detect_model_features()` stay. `RdmaStrategy.is_available()` calls `check_transfer_allowed()` before attempting P2P transfer (`load_strategy/rdma_strategy.py:135`). It does always return `True`, but the feature log it emits is its documented purpose (`docs/ARCHITECTURE.md`), and it is the intended single place for future safety gates. Removing it would delete a live call site and the QA feature log, which is a behavior change rather than a dead-code cleanup.

Note that transfer-time validation is not lost by this change: `ManifestMismatchError` already covers source/target tensor name and size mismatch on the receive path.

## Test plan

Full Python suite run on this branch and on pristine `main`, failure sets diffed:

- baseline (`main`): 60 failed, 1017 passed, 7 skipped, 17 errors
- this branch: 60 failed, 1012 passed, 7 skipped, 17 errors

Identical failure and error sets — zero regressions. The pre-existing failures are the CUDA/GPU-dependent tests, unrelated to this change. The 5-test drop in passes is exactly the deleted fingerprint round-trip and manifest-hash tests. `tests/test_transfer_safety.py` passes 12/12.

`pre-commit run` passes on the staged files.

## Summary by CodeRabbit

- **Changes**
  - Simplified transfer safety checks to focus on supported model features.
  - Removed environment and transfer-fingerprint comparisons from transfer validation.
  - Transfers are no longer blocked based solely on combinations of detected runtime features.

- **Documentation**
  - Updated architecture documentation to reflect the current transfer-safety behavior.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -171,7 +171,7 @@ ModelExpress/
 │       │       ├── adapter.py          # SglangAdapter and context builder
 │       │       └── loader.py           # MxModelLoader for remote_instance backend
 │       ├── tensor_utils.py             # Tensor collection, checksums, storage views
-│       ├── transfer_safety.py          # MLA feature gate, TransferFingerprint
+│       ├── transfer_safety.py          # Model feature detection for the P2P gate
 │       ├── rank_utils.py               # Rank detection utilities
 │       ├── vllm_worker.py              # Compatibility worker for older manual registration
 │       ├── types.py                    # TensorDescriptor, WorkerMetadata dataclasses
--- a/modelexpress_client/python/modelexpress/transfer_safety.py
+++ b/modelexpress_client/python/modelexpress/transfer_safety.py
@@ -4,33 +4,16 @@
 """
 Transfer safety checks for ModelExpress GPU-to-GPU weight transfer.
 
-Two independent safety mechanisms:
-
-1. Feature detection: surfaces model features (attention, quantization, MoE)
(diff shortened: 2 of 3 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/595/commits/4a9e5e456f1c1b255bcc1938c9f229e04c71ea80

zhengluo-nv (approved):

PR #596, opened 2026-08-07
fix: tokenreview caching improvements

wseaton:


modelexpress_server/src/auth/layer.rs
```diff
@@ -82,7 +82,7 @@ where
                     inner.call(req).await
                 }
                 Err(denial) => {
-                    warn!(reason = %denial, path = %req.uri().path(), "auth denied");
+                    debug!(reason = %denial, path = %req.uri().path(), "auth denied");
```
nicolasnoble, line 85: The single log line here covers every `Denial` variant uniformly and is now `debug!` rather than `warn!`. Combined with the counters only incrementing inside the cached review closure, that removes default-level visibility into two scenarios that are not attacks: a real service account not yet on the allowlist, caught by the new prefilter before it ever reaches a review, and a token whose verified identity does not match its claimed one, which still reaches `Denial::PermissionDenied` and is still exercised by `verify_rejects_identity_outside_allowlist`. Before this change the first case logged a specific message naming which service account needed adding; now it is indistinguishable from a garbage token. Was dropping that intentional, or is it worth keeping at `warn` or as its own counter, since it is the main thing an operator onboarding a new workload will hit?

modelexpress_server/src/auth.rs
```diff
@@ -46,93 +49,153 @@ impl Denial {
...
-struct TokenCacheKey(Zeroizing<String>);
-
-impl TokenCacheKey {
-    fn new(token: &str) -> Self {
-        Self(Zeroizing::new(token.to_string()))
-    }
+#[derive(Default)]
+struct AuthCounters {
+    reviews: AtomicU64,
+    rejected: AtomicU64,
```
nicolasnoble, line 55: Minor: `rejected` counts three different things together, malformed tokens, claims that parse but are off the allowlist, and tokens that reach `TokenReview` and are rejected there. If the summary line is meant to separate "someone is flooding us" from "someone forgot to allowlist a service account", one merged counter cannot.

modelexpress_server/src/auth.rs
```diff
@@ -46,93 +49,153 @@ impl Denial {
...
     negative_cache: Cache<TokenCacheKey, ()>,
+    review_permits: Semaphore,
+    counters: AuthCounters,
 }
 
 impl AuthState {
     const TOKEN_CACHE_MAX_ENTRIES: u64 = 10_000;
+    const NEGATIVE_CACHE_MAX_ENTRIES: u64 = 1_000;
+    const MAX_CONCURRENT_REVIEWS: usize = 8;
+    const MAX_CONCURRENT_REVIEWS_PER_CALLER: usize = 4;
```
nicolasnoble, line 94: Minor: the per-claimed-caller semaphore is shared across every pod claiming the same service account, at a fixed 4. The deployment doc describes projected per-pod tokens, so a rolling restart or scale-up of one deployment could produce more than 4 concurrent first-time auth attempts for the same claimed identity and shed some with `Unavailable` under the 250ms wait, with nothing malicious happening. Any reason 4 and 8 were picked independent of expected replica counts, or is that a tradeoff you are fine with?

nicolasnoble (commented):
Three comments inline. The first is the one worth your time; the other two are minor.

PR #598, opened 2026-08-07
feat(huggingface): support pinned revisions for ModelExpress downloads

### LINKED ISSUE #586: feat(huggingface): support pinned revisions for ModelExpress downloads
## Summary

Enable callers to request a Hugging Face model at an optional revision and always receive the immutable commit SHA that ModelExpress downloaded.

Related: https://github.com/ai-dynamo/modelexpress/issues/569
Downstream context: https://github.com/ai-dynamo/dynamo/pull/11369

## Problem

Inference workers and frontend metadata must use the same model commit. ModelExpress currently accepts only a model name, so a caller cannot request a branch, tag, or SHA, nor reliably learn the exact commit selected for an unpinned request.

## Current behavior

- ModelExpress downloads the provider's default/current model revision.
- The public download and RPC APIs do not accept a revision.
- The normal download result does not return the resolved commit SHA.
- A caller requiring a specific revision must bypass ModelExpress or rely on cache-only custom logic, which breaks cold-cache startup and risks model/metadata skew.

## Proposed behavior

- Add an optional `revision` to ModelExpress model-download and file-streaming requests.
- If supplied, resolve the requested branch, tag, or SHA to an immutable commit SHA.
- If omitted, resolve the provider's default revision once to an immutable commit SHA.
- Use that resolved SHA for all Hub file requests and the local snapshot/cache identity.
- Return the downloaded snapshot path and `resolved_revision` to the client.
- Use the resolved SHA in cache, lease, deduplication, and no-shared-storage streaming identity.

## Acceptance tests

- Cold-cache download with an explicit commit SHA produces a snapshot for that SHA and returns it.
- Cold-cache download with a tag or branch resolves it to a SHA; all files come from that SHA.
- An unpinned download returns the SHA selected for the provider's default revision.
- Two revisions of the same model do not share a cache/lease entry or coalesce.
- Metadata-only and full-weight downloads are distinguished correctly.
- No-shared-storage streaming installs a valid HF-compatible snapshot and reports the same resolved SHA.
- A nonexistent revision fails clearly and never falls back to the default revision.

AndyDai-nv:
Closes #586

## Summary

Callers can request a Hugging Face model at a branch, tag, or commit SHA, and always learn the immutable commit the request resolved to, so a frontend and its workers can be held to one model revision.

- `ModelDownloadRequest` / `ModelFilesRequest` carry an optional `revision`; `ModelStatusUpdate` reports `resolved_revision`.
- `ModelProviderTrait` gains revision-aware variants alongside the existing methods. Their defaults reject a pinned revision and delegate, so only Hugging Face implements them; NGC and GCS return a clear `InvalidArgument`.
- Client and CLI expose `--revision` and return the snapshot path plus the resolved revision. The pre-existing signatures are unchanged and delegate with nothing pinned, so no downstream Rust caller breaks.

## Two bugs this fixes along the way

**Files could come from a commit other than the one reported.** `HuggingFaceProvider::download_model` called `repo.info()` to get `info.sha` but then downloaded through `repo.get()`, which resolves against `main`. A branch moving mid-download could scatter files across commits while we reported a single SHA. Every file is now fetched from the resolved commit.

**A metadata-only download satisfied a later full-weight request.** `ignore_weights` downloads marked the model `DOWNLOADED` under the same registry key as a full download, so the next full-weight request got a cache hit on a snapshot with no weights in it.

## Registry entry keys

Registry backends key every record, lease, and claim on a single string, and a model name alone is not a sufficient key for either of the above. Records, leases, and claims now key on an `EntryKey` encoding model + resolved revision + weight mode, formatted `model@rev:<sha>` with a `#metadata` suffix for `ignore_weights`.

An unpinned, full-weight entry encodes to the bare model name, so records written before this change keep parsing, and providers without a revision concept keep the keys they have always used. No Redis schema or CRD change is needed: the K8s CR-name sanitizer already hashes the original string.

Cache eviction parses the key back to delete the right snapshot, and skips the filesystem delete while another entry still references the same commit.

## Three judgment calls worth review

- **Revision resolution happens before the lease claim**, which adds one HF metadata call per download request — the lease has to be keyed on the resolved revision. When resolution fails for an *unpinned* request, the server falls back to the snapshot already in its cache; otherwise a Hub outage would break cache hits that work today. A *pinned* request always fails hard, since falling back would hand back a different model than the one asked for.
- **Streamed snapshots now get a `refs/<revision>` entry** (`ModelProviderTrait::record_local_revision`). Without it the acceptance item "installs a valid HF-compatible snapshot" was not really met: the files were all present but `huggingface_hub` could not resolve them by branch under `HF_HUB_OFFLINE=1`. This is also the piece #569's cold-start flow depends on.
- **Eviction treats a revisionless entry as sharing files with every revision of its model.** After an upgrade an old bare-name record co-exists with new revision-scoped ones; without this it would delete their snapshot on its way out, costing a re-download.

## Acceptance criteria

Each item from #586 has a test:

| Criterion | Test |
|---|---|
| Cold-cache download with an explicit commit SHA | `test_download_at_explicit_commit_produces_that_snapshot` |
| Tag or branch resolves to a SHA; all files from that SHA | `test_download_at_tag_fetches_every_file_from_the_resolved_commit` |
| Unpinned download returns the default revision's SHA | `test_unpinned_download_reports_the_default_revision_commit` |
| Two revisions do not share a cache/lease entry | `test_tracker_scopes_the_lease_to_the_resolved_revision`, `entry_key` tests |
| Metadata-only vs full-weight distinguished | `test_tracker_separates_metadata_only_from_full_weight_downloads` |
| No-shared-storage streaming installs a valid snapshot and reports the same SHA | `test_request_model_revision_installs_the_resolved_snapshot`, `test_stream_model_files_serves_the_requested_revision` |
| A nonexistent revision fails and never falls back | `test_missing_revision_fails_without_falling_back_to_the_default`, `test_stream_model_files_rejects_an_uncached_revision` |

## Testing

`cargo test --workspace` (489 tests), `cargo clippy --workspace --all-targets` clean, `pre-commit run --all-files` clean. The HF tests are wiremock-backed and need no network. Docs updated in `docs/ARCHITECTURE.md` and `docs/CLI.md`.

## Summary by CodeRabbit

- **New Features**
  - Download models using a branch, tag, or commit revision.
  - Reports the resolved immutable commit SHA and installed path.
  - Supports revision-aware downloads, file access, caching, and deletion.
  - Enables offline use of previously cached revisions.
  - Distinguishes registry entries by model, revision, and weight mode.

- **Bug Fixes**
  - Prevents pinned downloads from silently falling back to the default revision.
  - Preserves shared cached snapshots during cleanup.

- **Documentation**
  - Added guidance and examples for revision pinning in the CLI and architecture documentation.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -333,9 +333,11 @@ When a no-shared-storage client installs a snapshot over the file stream, it als
 
 #### Registry entry keys
 
-Registry backends key every record, lease, and claim on one string. That key is a [`EntryKey`](../modelexpress_server/src/registry/entry_key.rs) encoding of `model_name` + resolved revision + weight mode, formatted as `model@rev:<sha>` with a `#metadata` suffix for `ignore_weights` downloads. This is what keeps two revisions of one model from sharing a lease (and coalescing onto each other's download), and what keeps a metadata-only download from satisfying a later full-weight request with a weightless snapshot. An unpinned, full-weight entry encodes to the bare model name, so records written before revisions existed keep parsing.
+Registry backends key every record, lease, and claim on one string. That key is an [`EntryKey`](../modelexpress_server/src/registry/entry_key.rs) encoding of `model_name` + resolved revision + weight mode. This is what keeps two revisions of one model from sharing a lease (and coalescing onto each other's download), and what keeps a metadata-only download from satisfying a later full-weight request with a weightless snapshot.
 
-Cache eviction parses the key back to delete the right snapshot, and skips the filesystem delete while another entry — typically the metadata-only twin of the same commit — still references it.
+An unpinned, full-weight entry encodes to the bare model name, so records written before revisions existed keep parsing and providers without a revision concept keep the keys they have always used. Anything else encodes as `mx1:<revision>:<flags>:<model_name>`. The model name goes last because it is the only field with no character restrictions — a GCS object path accepts almost any byte, so a name containing the separator must not be able to impersonate the other fields and aim a delete at the wrong model.
+
+Cache eviction parses the key back to delete the right snapshot. It keeps the files while another entry still references the same commit, treating a revisionless entry as covering every snapshot of its model. When the evicted entry is the last one for its model the whole repository directory goes, so a snapshot kept alive by an earlier shared-file decision cannot outlive the last record pointing at it. Hugging Face snapshot entries are symlinks into `blobs/`, so a per-revision delete also reclaims the blobs no surviving snapshot references — otherwise it would free almost no disk.
 
 ### p2p.proto - P2pService
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -98,6 +98,7 @@ ModelExpress/
 │       ├── registry/
 │       │   ├── state.rs                # RegistryManager wrapper
 │       │   ├── backend.rs              # RegistryBackend trait + ModelRecord
+│       │   ├── entry_key.rs            # EntryKey: model + revision + weight mode
 │       │   ├── k8s_types.rs            # ModelCacheEntry CRD type
 │       │   └── backend/
 │       │       ├── redis.rs            # Redis registry backend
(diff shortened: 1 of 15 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/598/commits/e606cfa757f1ec5fb7dc0c614d4b923f7affbe7f

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/598/commits/d80c29231b36380463925c7c3132a87e9b91754b

zhengluo-nv (approved):

PR #600, opened 2026-08-08
fix(vmm): request FABRIC handle type so arena memory is MNNVL-exportable

chanh:
`CudaVmmBackend._allocation_prop()` never set `prop.requestedHandleTypes`, so
every `cuMemCreate` in the VMM arena produced an allocation with no exportable
handle types. `uct_cuda_ipc_mem_reg`'s
`cuMemExportToShareableHandle(CU_MEM_HANDLE_TYPE_FABRIC)` therefore fails, and
peer-to-peer over the UCX `cuda_ipc` transport is impossible for arena-backed
memory on multi-node NVLINK (MNNVL) systems.

Measured on GB300 (2 nodes on one MNNVL fabric, 512 MB tensor), replaying what
`uct_cuda_ipc_mem_reg` does:

                                     | arena (before) | default allocator
    cuPointerGetAttribute(14)        | 0x0            | 0x8 (FABRIC)
    cuMemRetainAllocationHandle      | rc=0           | rc=0
    cuMemExportToShareableHandle     | rc=1 FAILED    | rc=0 OK

With this change the arena reports `0x8` and the export returns `rc=0`.

The fabric handle type is probed once and the outcome cached: platforms that
reject it fall back to a plain allocation with a warning rather than failing
the load, so non-fabric hosts are unaffected.

Verified end to end with Kimi-K3 at TP8 across two GB300 nodes: 2712 tensors /
206.38 GB per rank transferred in ~1.0 s at 1.6-2.36 Tbps, versus a disk load
of 334-454 s. Without the change the transfer cannot start at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA virtual memory allocation compatibility on systems that do not support FABRIC handles.
  * Added automatic fallback to standard memory allocation when FABRIC-exportable memory is unavailable.
  * Preserved clear error reporting when FABRIC allocation fails after support has been detected.

* **Improvements**
  * Cached hardware capability checks to provide more reliable and efficient memory allocation behavior.

### DIFF
--- a/modelexpress_client/python/modelexpress/vmm/backend.py
+++ b/modelexpress_client/python/modelexpress/vmm/backend.py
@@ -81,6 +81,8 @@ class CudaVmmBackend:
     def __init__(self, device: int = 0) -> None:
         self._cu = _import_cuda()
         self._device = device
+        # None = not yet probed; True/False = FABRIC export usable on this host.
+        self._fabric_supported = None
         self._ensure_context()
         self._granularity = self._compute_granularity()
         logger.debug(
@@ -124,8 +126,44 @@ class CudaVmmBackend:
             raise ValueError(
                 f"size {size} is not a multiple of granularity {self._granularity}"
             )
-        prop = self._allocation_prop()
-        handle = self._call("cuMemCreate", self._cu.cuMemCreate(size, prop, 0))
+        # Probe FABRIC once; hosts that reject it fall back to a plain alloc.
+        handle = None
+        if self._fabric_supported is not False:
+            fabric = self._fabric_handle_type()
+            if fabric is not None:
+                try:
+                    handle = self._call(
+                        "cuMemCreate",
+                        self._cu.cuMemCreate(
+                            size, self._allocation_prop(fabric), 0
+                        ),
+                    )
+                    if self._fabric_supported is None:
+                        self._fabric_supported = True
+                        logger.info(
+                            "VMM arena: FABRIC-exportable allocations enabled"
+                        )
+                except CudaBackendError as exc:
+                    # Only a platform-level refusal means FABRIC is unavailable;
+                    # anything else (OOM, bad device) must not disable it.
+                    if (
+                        self._fabric_supported is None
+                        and exc.result_code in self._fabric_unsupported_codes()
+                    ):
+                        self._fabric_supported = False
+                        logger.warning(
+                            "VMM arena: FABRIC handle type unsupported (%s); "
+                            "falling back to non-exportable allocations -- "
+                            "MNNVL cuda_ipc peer-to-peer will NOT be available",
+                            exc,
+                        )
+                    else:
+                        raise
+        if handle is None:
+            prop = self._allocation_prop()
+            handle = self._call(
+                "cuMemCreate", self._cu.cuMemCreate(size, prop, 0)
+            )
         try:
             self._call("cuMemMap", self._cu.cuMemMap(va, size, 0, handle, 0))
             try:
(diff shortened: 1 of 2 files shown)


chanh:
Thanks — both findings were real bugs in the patch, fixed and force-pushed:

**Cache only FABRIC-specific failures.** The blanket `except Exception` meant a transient `CUDA_ERROR_OUT_OF_MEMORY` on the first arena allocation would latch `_fabric_supported = False` and silently disable MNNVL P2P for the rest of the process. Now catches `CudaBackendError` and only caches when `result_code` is `CUDA_ERROR_NOT_SUPPORTED` or `CUDA_ERROR_NOT_PERMITTED` (looked up via `getattr` so older bindings don't break); everything else re-raises with `_fabric_supported` untouched.

**cuda-python floor.** Raised to `>=12.4` in the `vmm` extra — it's scoped to the `CudaVmmBackend` this change touches, so the blast radius is limited to users of that allocator. Also added a warning when the enum is missing, so the degrade is visible rather than silent.

DCO sign-off added.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/600/commits/90b55caf509cee0a2891ed9c8d19b76ce6598a44

chanh:
`P2P test (sglang-mooncake)` failed here (`test_rdma_transfer_logged`: `SGLang MxModelLoader transfer_engine COMPLETE` not found in target logs). I don't think it's related to this change — evidence below, but I can't re-run it myself to confirm (needs repo admin).

- **The changed code isn't reachable in this test.** The patch only executes inside `CudaVmmBackend.allocate`, which requires `MX_VMM_ARENA=1`. That variable isn't set anywhere in `ci/`, `.github/`, or the charts — it appears only inside the library. With the arena off, `_allocation_prop` / the FABRIC probe never run.
- **`P2P test (sglang)` passed on this same commit**, with identical code. Only the mooncake variant failed.
- **Mooncake is a separate transfer engine** — no references in the Python client this patch touches.
- The target came up healthy in the log (`The server is fired up and ready to roll!`, `/generate` 200 OK). Only the completion marker was absent, which reads more like a timing/scrape race than a transfer failure.

The other change in this PR is the `cuda-python>=12.4` floor in the `vmm` extra, which likewise doesn't affect a run that never constructs the arena.

Happy to be wrong — if a maintainer re-runs the job and it fails again, I'll dig in. Worth noting #601 has the same P2P suite pending and doesn't touch `vmm/backend.py` at all, so its result on this test is a useful control.

chanh:
Follow-up: a second P2P job failed here (`P2P DP=2 test (vllm)`), and it is unambiguously infrastructure rather than code —

```
Failed to schedule pod, did not tolerate taint (taint=CriticalAddonsOnly=true:NoSchedule);
incompatible requirements, label "agentpool" does not have known values
```

Karpenter couldn't provision nodes, so `mx-source` / `mx-target` never started.

That also prompted me to check the one part of this PR that *isn't* behind the arena — the `cuda-python>=12.4` floor. It's likewise inert in CI:

- `[vmm]` appears nowhere in the repo, so the extra is never installed
- the worker images install the client with a plain `uv pip install .` (base deps only)

So neither change in this PR is exercised by these jobs: the arena code needs `MX_VMM_ARENA=1` (never set in CI), and the dependency floor needs the `vmm` extra (never installed).

Both failures look environmental — one scheduling, one a missing log marker on an otherwise healthy server. Re-runs would confirm; I don't have permission to trigger them.

chanh:
Update: #601 hit the identical `FailedScheduling` error on its P2P jobs, and that PR doesn't touch `vmm/backend.py` at all — so this is environmental rather than specific to this change.

The pod names in the logs carry the run IDs of several concurrent, unrelated workflows (`31412207831`, `31415610148`, `31418088319`, `31434814575`, …), all hitting:

```
Failed to schedule pod, did not tolerate taint (taint=CriticalAddonsOnly=true:NoSchedule);
incompatible requirements, label "agentpool" does not have known values
```

Looks like the CI cluster's node pool needs attention — it appears to be affecting PRs broadly.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/600/commits/f5fb94e0cb065c8d6800b773c38a4f845e5e74de

AndyDai-nv (approved):

PR #601, opened 2026-08-08
fix(nixl): validate arena coverage before single-MR registration

chanh:
`register_arena()` registers one MR over the arena's `[base, base+used)` bump
range, but never checks that every tensor actually lies inside it. NIXL resolves
transfer descriptors by *containment* (`nixlSecDescList::getCoveringIndex` ->
`nixlBasicDesc::covers`), so a single uncovered tensor makes the whole
`prep_xfer_dlist` fail with `NIXL_ERR_NOT_FOUND` -- not just that one
descriptor. The receive then fails with no indication of which tensor was at
fault.

This is reachable whenever a tensor is allocated outside the arena's
`use_arena` scope -- for example buffers created after load by post-processing,
which do not come from the arena's bump allocator.

Observed on GB300 with Kimi-K3 at TP8: the arena registered cleanly as
1 region / 208.40 GB and the consumer reached
`Receiving 2712 tensors from source (P2P)`, then failed in `prep_xfer_dlist`
with `NIXL_ERR_NOT_FOUND` and no further detail. Per-tensor registration
(2712 regions, 1:1 with the descriptors) transfers correctly, which is what
isolated the coverage gap.

Containment resolution itself was verified against the installed nixl wheel:
a descriptor inside a registered region resolves, 512 sub-descriptors inside a
single region resolve, and a descriptor overrunning the region end or with a
mismatched devId returns NotFound.

This adds an up-front coverage check that falls back to per-tensor registration
and names the first uncovered tensor, turning a silent whole-transfer failure
into a warning plus a working (if less compact) registration. It does not
change behaviour when the arena does cover every tensor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tensor registration reliability by validating that all tensors fit within the registered memory range.
  * Automatically falls back to individual tensor registration when tensors extend beyond that range.
  * Added diagnostic logging to help identify registration issues.

### DIFF
--- a/modelexpress_client/python/modelexpress/nixl_transfer.py
+++ b/modelexpress_client/python/modelexpress/nixl_transfer.py
@@ -280,7 +280,11 @@ class NixlTransferManager:
         self._tensor_descriptors = tensor_descriptors
         return tensor_descriptors
 
-    def register_tensors(self, tensors: dict[str, torch.Tensor]) -> bytes:
+    def register_tensors(
+        self,
+        tensors: dict[str, torch.Tensor],
+        force_per_tensor: bool = False,
+    ) -> bytes:
         """
         Register tensors with NIXL for RDMA access.
 
@@ -318,7 +322,7 @@ class NixlTransferManager:
 
         # Phase 1: Discover CUDA allocation boundaries (if pool reg enabled)
         alloc_discovery_start = time.perf_counter()
-        if _pool_reg_enabled():
+        if _pool_reg_enabled() and not force_per_tensor:
             if self._accelerator_backend.supports_pool_reg():
                 allocations = self._find_cuda_allocations(tensor_descriptors)
             else:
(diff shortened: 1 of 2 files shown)


chanh:
Thanks — good catch, the fallback was incomplete. Fixed and force-pushed.

You're right that `register_tensors()` still routes through `_find_cuda_allocations` when `MX_POOL_REG=1`, which resolves the same per-handle `cuMemGetAddressRange` bounds that made the arena range insufficient in the first place — so the fallback could re-trigger the exact `NIXL_ERR_NOT_FOUND` it exists to avoid.

Added `force_per_tensor: bool = False` to `register_tensors()`, honoured in the pool-reg branch, and set it at the coverage fallback so that path always registers 1:1 with the descriptors.

DCO sign-off added.

chanh:
CI caught a real problem with the patch — fixed and pushed.

`test_arena_registration_uses_vram_segment` was failing because its `FakeArena` returned a hardcoded `(0x1000, 0x2000)` that doesn't cover the test tensor's actual address, so the new coverage check correctly diverted it to the per-tensor fallback. The fake wasn't faithful to a real arena, which is precisely what the check is designed to catch. Updated it to derive the range from the tensor so it exercises the single-MR path it's meant to test.

Also added the two cases you asked for:
- `test_arena_registration_falls_back_when_tensor_uncovered` — an uncovered tensor falls back to per-tensor registration
- `test_arena_fallback_bypasses_pool_registration` — with `MX_POOL_REG=1`, the fallback still registers per tensor rather than going through `_find_cuda_allocations`, which is the case that would otherwise re-trigger `NIXL_ERR_NOT_FOUND`

`tests/test_pool_registration.py`: 28 passed locally.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/601/commits/6441ba9d3cad2180d68703517b3f007ed282b2b5

chanh:
The P2P job failures on this PR (and on #600) are a **cluster-wide CI scheduling problem**, not these changes.

The same error appears across many concurrent workflow runs — the run IDs are visible in the pod names (`31412207831`, `31415610148`, `31418088319`, `31423692184`, `31434814575`, …), i.e. multiple unrelated PRs:

```
Failed to schedule pod, did not tolerate taint (taint=CriticalAddonsOnly=true:NoSchedule);
incompatible requirements, label "agentpool" does not have known values
```

Karpenter isn't provisioning nodes, so `mx-source`/`mx-target` never start and the tests time out.

Independently, the change here isn't reachable in these jobs: `register_arena()` is only called when `ctx.vmm_arena is not None` (`load_strategy/base.py:201`), which requires `MX_VMM_ARENA=1` — not set anywhere in `ci/`, `.github/`, or the charts. The unit tests that do exercise it pass (`tests/test_pool_registration.py`, 28 passed).

Flagging in case the CI cluster's node pool needs attention — it looks like it's affecting everyone's PRs right now, not just these.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/601/commits/7b8de2dda1c2524566240c96de9c69d1a931b8df

AndyDai-nv (approved):

PR #602, opened 2026-08-10
fix(weight_transfer): register local destination memory on the trainer-pull path

nicolasnoble:
TrainerPullStrategy never registered its local destination memory with NIXL. A NIXL READ requires the local destination to sit inside a registered region, so prep_xfer_dlist on the local side could not resolve and the transfer failed with NIXL_ERR_NOT_FOUND. The failure is silent from the caller's point of view: the strategy falls back to rdma and then to instant_tensor, so what looks like a successful weight load has quietly gone to disk instead of over the fabric. Every other strategy in the chain (default, gds, instant_tensor, model_streamer, rdma) already makes this call in the same place; this path was the only one that never did.

The second commit closes the gap that leaves. register_tensors is deliberately best-effort and swallows failures, which is correct at its other call sites, where a miss costs only P2P serving capability. On the trainer-pull path the registered region is the destination of an inbound READ, so a skip means the same silent fall-through to disk. It now raises StrategyFailed(mutated=False) when no descriptors exist after registration. The mutated flag matters on its own: letting the condition reach PullRole.initialize would surface it as mutated=True and claim the model may be half-written when nothing has been touched.

I'd rather not mirror the registration preconditions into is_available() as well, which was the other half of the automated review suggestion. The two copies would drift, and the guard covers the case at the point where it matters.

Measured on real hardware against a live trainer holding 399 registered tensors, 16.38 GB total. Before the change the READ fails at prep on both current main and on the consolidation branch in #588. After it, main completes the transfer in 0.576s at 227.4 Gbps and #588 completes it in 0.577s at 226.9 Gbps, which is the single-QP ceiling for a 400 Gb/s link.

The original defect is not reachable from the mocked suite, which passes with and without the registration call, because it only appears once there is a real NIXL agent on the far side and memory actually has to be registered. It was found with a scratch end-to-end harness that is not in the repo, and getting that harness or an equivalent into CI is worth doing separately. The precondition guard added in the second commit is reachable, and is covered in both branches plus the healthy path; those tests were verified to fail when the guard is removed.

Signed-off-by: Nicolas 'Pixel' Noble <nicolas@nobis-crew.org>

### DIFF
--- a/modelexpress_client/python/modelexpress/load_strategy/trainer_pull_strategy.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/trainer_pull_strategy.py
@@ -31,7 +31,14 @@ from typing import TYPE_CHECKING, Any
 
 from ..adapter import EngineAdapter, StrategyFailed
 from ..nixl_transfer import is_nixl_available
-from .base import LoadContext, LoadResult, LoadStrategy, _as_load_result, _init_nixl_manager
+from .base import (
+    LoadContext,
+    LoadResult,
+    LoadStrategy,
+    _as_load_result,
+    _init_nixl_manager,
+    register_tensors,
+)
 from .. import envs
 from ..weight_transfer.protocol.serialization import decode_trainer_table
 from ..weight_transfer.roles.pull import PullRole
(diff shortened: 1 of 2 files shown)


zhengluo-nv (approved):

modelexpress_client/python/modelexpress/load_strategy/trainer_pull_strategy.py
zhengluo-nv, line 88: Mixing RL with inference load strategy chain is not the long-term path. We probably want to separate and have a RL load strategy chain

tanushriya910 (approved):

modelexpress_client/python/modelexpress/load_strategy/trainer_pull_strategy.py
nicolasnoble: agreed, this shouldn't live in the inference chain long term. keeping it out of scope here since this PR is a six line bug fix on the existing path, but I don't disagree with the direction and it's worth its own discussion.

PR #604, opened 2026-08-10
feat(refit): add Redis-backed lifecycle control plane

zhengluo-nv:
## Summary

- add a clean RL-specific `RefitService` without changing the existing `WeightSyncService`
- define external orchestrator APIs for creating, reading, and releasing immutable weight versions
- define internal worker APIs for registration, manifest publication, discovery, version leases, and safe source-manifest eviction
- add the `STAGING -> READY -> RELEASING` lifecycle
- use a fresh TTL-bound `worker_id` for each process lifetime; `RegisterWorker` is also the heartbeat API
- represent each worker's manifest publication as `WeightVersionShard`; retain multiple publications for a source slot so a replacement or peer generator can become an additional source
- use `RegisterVersionLease` and `DeleteVersionLease` as the complete lease API
- organize storage behind a backend-neutral `RefitBackend` contract, following the existing inference backend layout
- implement Redis first with one documented atomic Lua transition per domain operation; Kubernetes CRD support remains a separate backend implementation

Weight bytes and full tensor manifests remain on trainer and generator workers. Redis stores only control-plane metadata.

## API visibility

External APIs called by the RL framework orchestrator:

- `CreateWeightVersion`
- `GetWeightVersion`
- `DeleteWeightVersion`

Internal APIs called by ModelExpress worker clients:

- `RegisterWorker`
- `CreateWeightVersionShard`
- `ListWeightVersionShards`
- `DeleteWeightVersionShard`
- `RegisterVersionLease`
- `DeleteVersionLease`

## Lifecycle guarantees

- a version becomes `READY` atomically after all expected version-scoped source slots are published
- `WeightVersionShard` is a physical worker manifest, while `source_slot_id` identifies the required logical contribution; multiple worker publications may cover the same slot
- deleting a `READY` version moves it to `RELEASING`
- `RELEASING` rejects new leases and shard publications
- `RegisterVersionLease` acquires a lease only while a version is `READY`
- registering the same live lease owner renews its expiry while the version is `READY` or `RELEASING`, so an in-flight generator update can finish
- an expired lease cannot be registered again after logical release begins
- any active lease protects every shard of the version
- explicit lease release or TTL expiration allows source workers to evict their shards

## Validation

- `cargo fmt --all`
- `cargo check -p modelexpress-server -p model-express-workspace-tests`
- `cargo test -p model-express-workspace-tests --test refit_service_redis --no-run`
- `git diff --check`
- Redis 7 public gRPC E2E coverage exercises concurrent idempotent creation, source-slot readiness, replacement-source publication, lease/release/eviction, and TTL expiry. Local execution is blocked by the local Docker daemon returning HTTP 500; CI runs it against Redis 7.

### DIFF
--- a/modelexpress_common/proto/refit.proto
+++ b/modelexpress_common/proto/refit.proto
@@ -1,0 +1,165 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package model_express.refit;
+
+// Control-plane metadata for RL weight publication. Weight bytes and full
+// manifests remain on trainer and generator workers.
+service RefitService {
+  // External APIs: called by the RL framework orchestrator.
+  rpc CreateWeightVersion(CreateWeightVersionRequest) returns (WeightVersion);
+  rpc GetWeightVersion(GetWeightVersionRequest) returns (WeightVersion);
+  rpc DeleteWeightVersion(DeleteWeightVersionRequest) returns (WeightVersion);
+
+  // Internal APIs: called by ModelExpress worker clients, not by the RL
+  // framework orchestrator.
+  rpc RegisterWorker(RegisterWorkerRequest) returns (WorkerRegistration);
+  rpc CreateWeightVersionShard(CreateWeightVersionShardRequest) returns (CreateWeightVersionShardResponse);
+  rpc ListWeightVersionShards(ListWeightVersionShardsRequest) returns (ListWeightVersionShardsResponse);
+  rpc DeleteWeightVersionShard(DeleteWeightVersionShardRequest) returns (DeleteWeightVersionShardResponse);
+  // A live lease protects every shard while a generator installs the version.
+  // Re-registering the same lease owner renews its expiry.
+  rpc RegisterVersionLease(RegisterVersionLeaseRequest) returns (VersionLease);
+  rpc DeleteVersionLease(DeleteVersionLeaseRequest) returns (DeleteVersionLeaseResponse);
+}
+
+enum WorkerRole {
+  WORKER_ROLE_UNSPECIFIED = 0;
+  WORKER_ROLE_TRAINER = 1;
+  WORKER_ROLE_GENERATOR = 2;
+}
+
+enum WeightPayloadFormat {
+  WEIGHT_PAYLOAD_FORMAT_UNSPECIFIED = 0;
+  WEIGHT_PAYLOAD_FORMAT_FULL_TENSOR = 1;
+  WEIGHT_PAYLOAD_FORMAT_XOR_DELTA = 2;
+}
+
+enum WeightVersionState {
+  WEIGHT_VERSION_STATE_UNSPECIFIED = 0;
+  WEIGHT_VERSION_STATE_STAGING = 1;
+  WEIGHT_VERSION_STATE_READY = 2;
+  WEIGHT_VERSION_STATE_RELEASING = 3;
+}
+
+message WorkerRegistration {
+  // Fresh identity for this concrete process lifetime. A restarted worker
+  // receives a new worker_id rather than fencing a prior registration.
+  string worker_id = 1;
+  reserved 2, 3;
+  WorkerRole role = 4;
+  string model_name = 5;
+  string endpoint = 6;
+  uint64 expires_at_unix_ms = 7;
+}
+
+message RegisterWorkerRequest {
+  WorkerRegistration worker = 1;
+  uint32 ttl_seconds = 2;
+}
+
+message WeightVersion {
+  // Server-generated 8-character lowercase hexadecimal ID.
+  string version_id = 1;
+  string model_name = 2;
+  optional uint64 sequence = 3;
+  string idempotency_key = 4;
+  WeightPayloadFormat payload_format = 5;
+  string base_version_id = 6;
+  // Required version-scoped source contributions. Physical source workers are
+  // not baked in; the trainer coordinator supplies opaque IDs such as a
+  // selected publisher global rank.
+  repeated string expected_source_slots = 7;
+  string layout_signature = 8;
+  WeightVersionState state = 9;
+  uint64 created_at_unix_ms = 10;
+}
+
+message CreateWeightVersionRequest {
+  string model_name = 1;
+  optional uint64 sequence = 2;
+  string idempotency_key = 3;
+  WeightPayloadFormat payload_format = 4;
+  string base_version_id = 5;
+  repeated string expected_source_slots = 6;
+}
+
+message GetWeightVersionRequest {
+  string version_id = 1;
+}
+
+message DeleteWeightVersionRequest {
+  string version_id = 1;
+}
+
+message WeightVersionShard {
+  string version_id = 1;
+  // Unique physical publication of one worker's complete manifest.
+  string shard_id = 2;
+  // Required version-scoped source contribution covered by this publication.
+  string source_slot_id = 3;
+  string worker_id = 4;
+  reserved 5, 6;
+  uint64 tensor_count = 7;
+  uint64 total_bytes = 8;
+  string manifest_digest = 9;
+  string manifest_endpoint = 10;
+  string transport = 11;
+}
+
+message CreateWeightVersionShardRequest {
+  WeightVersionShard shard = 1;
+}
+
+message CreateWeightVersionShardResponse {
+  WeightVersionShard shard = 1;
+  WeightVersion version = 2;
+}
+
+message ListWeightVersionShardsRequest {
+  string version_id = 1;
+}
+
+message ListWeightVersionShardsResponse {
+  repeated WeightVersionShard shards = 1;
+}
+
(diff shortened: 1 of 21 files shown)


modelexpress_server/src/refit/backend.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[error("{0}")]
+    FailedPrecondition(String),
+    #[error("{0}")]
+    AlreadyExists(String),
+    #[error("{0}")]
+    ResourceExhausted(String),
+    #[error("{0}")]
+    Internal(String),
+    #[error("Refit metadata backend error: {0}")]
+    Unavailable(String),
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 77d30c8. `Unavailable` now carries raw backend text; `backend_status` owns the client-facing Refit prefix, while server startup adds only its contextual failure message. The prefix is emitted once.

modelexpress_server/src/server.rs
```diff
@@ -115,6 +117,17 @@ pub async fn run_server(
...
+    let refit_service = create_refit_backend(&backend)
+        .await
+        .map_err(|error| {
+            error!("Failed to connect Refit metadata backend: {error}");
+            error
+        })?
+        .map(RefitServiceImpl::new);
+    if refit_service.is_none() {
+        info!("Refit service is unavailable: this metadata backend is not implemented yet");
+    }
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 77d30c8. Refit-backend initialization now uses the same 10-second timeout policy as the registry and P2P backends. The Refit service is also reported as serving through gRPC health when its backend is available, with Redis E2E coverage for that health check.

workspace-tests/tests/refit_service_redis.rs
```diff
@@ -0,0 +1,537 @@
...
+        .expect("create version")
+        .into_inner();
+    let mut current_shard = shard(&version.version_id, "tp-0", &worker_id);
+    current_shard.worker_generation = 2;
+    client
+        .create_weight_version_shard(CreateWeightVersionShardRequest {
+            shard: Some(current_shard),
+        })
+        .await
+        .expect("refreshed registration remains live past its original TTL");
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 77d30c8. The refresh now uses a 10-second TTL, validates liveness after 5 seconds, and waits another 6 seconds before the expiry assertion. This leaves a five-second margin for the publication RPCs on a loaded runner.

zhengluo-nv:
Follow-up on the remaining review-body suggestions in 77d30c8:

- Added the atomic source-slot membership check, atomic publication-state response, enum arguments for Lua scripts, Redis error classification, Refit health registration, startup timeout, and wider TTL-test margin.
- The generation-fence suggestion no longer applies because Refit now uses a fresh worker ID per process lifetime.
- Lease IDs intentionally remain 8 hex characters, consistent with the approved short internal-ID design.
- Idempotency-key retention remains open because its replay window must be defined together with the version metadata-retention policy; this PR does not invent a TTL.

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+
+syntax = "proto3";
+
+package model_express.refit;
+
+// Control-plane metadata for RL weight publication. Weight bytes and full
+// manifests remain on trainer and generator workers.
+service RefitService {
+  // External APIs: called by the RL framework orchestrator.
+  rpc CreateWeightVersion(CreateWeightVersionRequest) returns (WeightVersion);
```
nv-hwoo, line 12: Why do we want the orchestrator to call this API before the trainers call `CreateWeightVersionShard`? Why not just go directly to the trainers?

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+  WorkerRegistration worker = 1;
+  uint32 ttl_seconds = 2;
+}
+
+message WeightVersion {
+  // Server-generated 8-character lowercase hexadecimal ID.
+  string version_id = 1;
+  string model_name = 2;
+  optional uint64 sequence = 3;
+  string idempotency_key = 4;
```
nv-hwoo, line 68: what is this for?

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+
+message GetWeightVersionRequest {
+  string version_id = 1;
+}
+
+message DeleteWeightVersionRequest {
+  string version_id = 1;
+}
+
+message WeightVersionShard {
```
nv-hwoo, line 102: do we need model ID here as well? how does MX know which `CreateWeightVersionShardRequest` corresponds to which model ID?

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+}
+
+message RegisterWorkerRequest {
+  WorkerRegistration worker = 1;
+  uint32 ttl_seconds = 2;
+}
+
+message WeightVersion {
+  // Server-generated 8-character lowercase hexadecimal ID.
+  string version_id = 1;
```
nv-hwoo: Do we need the version to be in hexadecimal? it would be hard to track/understand the version progression if they're in hex.

modelexpress_common/proto/refit.proto
zhengluo-nv: I want to expose an abstracted weight version object to external RL frameworks and human operators. With weight version as a resource, human can easily reason about what is the current active versions, what versions are deleted, etc. Otherwise, they needs to look into hundreds of WeightVersionShard to figure out source of truth.

Calling CreateWeightVersion can also initialize an global unique ID that can be used to create shards

modelexpress_common/proto/refit.proto
zhengluo-nv: `version_id` is for our system which is guaranteed unique, RL framework agnostic. `sequence` is a human-readable string, could be a training job + training step

modelexpress_common/proto/refit.proto
zhengluo-nv: To ensure idempotency of CreateWeightVersion API, i.e. one caller calling the API twice won't create 2 versions, and only create one version

modelexpress_common/proto/refit.proto
zhengluo-nv: WeightVersionShard and WeightVersion are associated by version_id, which is a MX-owned UUID. model id is a customer/RL framework provided string. 

WeightVersion1 --> N WeightVersionShards
WeightVersion2 --> N WeightVersionShards
...

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+  string model_name = 2;
+  optional uint64 sequence = 3;
+  string idempotency_key = 4;
+  WeightPayloadFormat payload_format = 5;
+  string base_version_id = 6;
```
tanushriya910: nit: wondering if this base_version_id is related to delta weight update, maybe worth adding a comment to clarify.

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+  // Required version-scoped source contributions. Physical source workers are
+  // not baked in; the trainer coordinator supplies opaque IDs such as a
+  // selected publisher global rank.
```
tanushriya910, line 75: question: whats the trainer coordinator being referred here, is this something that RL orchestrator side needs to assemble

modelexpress_common/proto/refit.proto
```diff
@@ -0,0 +1,165 @@
...
+}
+
+message DeleteWeightVersionRequest {
+  string version_id = 1;
+}
+
+message WeightVersionShard {
+  string version_id = 1;
+  // Unique physical publication of one worker's complete manifest.
+  string shard_id = 2;
```
tanushriya910: question: why do we need shard_id as well as worked_id and source_slot_id

modelexpress_common/proto/refit.proto
zhengluo-nv: Yes, only used by delta weight update. Will add comment and make it optional

modelexpress_common/proto/refit.proto
zhengluo-nv: My understanding is, each weight version consists of N shards, published by each trainer worker. Once all N shards are published, the weight version becomes READY. This expected_source_slots is for this use case: each shard has a source_slot_id provided by the trainer frameworks like Megatron, and MX will monitor if all expected_source_slots are published or not.

modelexpress_common/proto/refit.proto
zhengluo-nv: Removed shard_id. A shard is identified by (version_id, worker_id, source_slot_id)

modelexpress_common/proto/refit.proto
nv-hwoo: kinda nit but can we use different name for this unique ID? I think the frameworks often refer to weight versions as the incremental integer values when they say "version" so would be confusing if we use it differently in MX.

zhengluo-nv: Rename them to be uid and version_number

nv-hwoo (approved):

PR #605, opened 2026-08-11
fix(sglang): support model_path in GDS loading

daii-0818:
## Summary

- resolve the checkpoint path from vLLM `model` or SGLang `model_path`
- preserve revision and tqdm handling for both engines
- add a regression test using an SGLang-shaped model config

## Problem

The shared GDS strategy currently reads `ctx.model_config.model`. SGLang exposes the checkpoint as `model_path`, so GDS loading fails immediately with:

```text
'ModelConfig' object has no attribute 'model'
```

The strategy then falls back to the default disk loader, preventing SGLang from using GDS.

## Testing

```text
uv run --frozen --extra dev pytest tests/test_gds_loader.py tests/test_sglang_loader.py tests/test_vllm_loader.py -q
129 passed, 8 skipped
```

## Summary by CodeRabbit

* **Bug Fixes**
  * GDS model loading now supports configurations that provide a model path instead of a model reference.
  * Loading continues to honor the configured revision and progress-display settings.
  * Clearer errors are raised when no model or model path is available.

* **Tests**
  * Added coverage for model-path loading and loader shutdown behavior.


daii-0818:
PR #605:

@zhengluo-nv Could you please review this SGLang/GDS compatibility fix when you have a chance? It allows GDS loading to resolve SGLang's `model_path` while preserving the existing `model` precedence. All CI checks are green.

zhengluo-nv:
/ok to test [491ff9f](https://github.com/ai-dynamo/modelexpress/pull/605/commits/491ff9f46fc1af9325e3d3eb103f3309509f0a9b)

zhengluo-nv (approved):

PR #615, opened 2026-08-12
fix(vllm): derive expert-parallel world size instead of reading a missing field

nicolasnoble:
vLLM's ParallelConfig has no expert_parallel_size attribute. metadata/publish.py read one anyway with a getattr default of 0, so the expert-parallel component of every vLLM mx_source_id has been a constant zero. sglang and TRT-LLM populate it for real, so this is the vLLM path only.

It produces false matches, not failed ones. An MoE deployment with --enable-expert-parallel and one without hash identically, and after #363 both have a worker at worker_rank 0, so the transfer succeeds and delivers the wrong expert weights with nothing raised and no disk fallback.

build_source_identity now computes tp * dp * pcp when enable_expert_parallel is set and 1 otherwise, following vllm/model_executor/layers/fused_moe/config.py. The attribute is prefill_context_parallel_size; the longer name in parallel_state.py is a local parameter, and reading that one would reintroduce the same silent default.

No dp field on SourceIdentity. manifest-azure-dp2.yaml and TEST_PLAN row 5.2 describe that omission as deliberate, DP cores being interchangeable copies, and the corrected product already carries dp. It does leave pcp unhashed, which fits better with the wider topology work.

Adds the first test coverage for build_source_identity, and corrects five places in CI and TEST_PLAN claiming that enabling EP raises the value above zero. None were executable assertions, which is why the ep2 job passed with both sides publishing the same wrong number.

This changes published mx_source_id values for vLLM sources, non-MoE included, without a version change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected expert-parallel world-size reporting in source identity metadata.
  * Expert-parallel size is now derived from tensor, data, and prefill-context parallelism.
  * Disabled expert parallelism now reports a size of 1 instead of 0.
  * Excluded decode-context parallelism from the calculation.

* **Documentation**
  * Updated metadata descriptions, test plans, deployment guidance, and version-bump procedures to reflect the corrected behavior.

* **Tests**
  * Added coverage for expert-parallel sizing, source-identity propagation, defaults, and related configuration handling.


modelexpress_client/python/modelexpress/metadata/publish.py
```diff
@@ -41,6 +41,32 @@ def _get_worker_server(device_id: int) -> "WorkerGrpcServer | None":
     return _worker_servers.get(device_id)
 
 
+def _derive_expert_parallel_size(parallel) -> int:
```
zhengluo-nv: Please move vLLM specific code into the engines/vllm directory

nicolasnoble: Moved in 66db2d61: `build_source_identity` plus its two private helpers now live in `engines/vllm/source_identity.py`. Left the rest of `metadata/publish.py` where it was, since sglang and `load_strategy/base.py` both import from it. Shout if you meant a narrower slice than that.

zhengluo-nv (approved):

PR #616, opened 2026-08-12
feat(refit): add RL trainer client and Megatron adapter

zhengluo-nv:
## Summary

- add the synchronous, rank-local `ModelExpressTrainerClient` lifecycle for staging and publishing versioned trainer shards
- add a framework-neutral `TrainerEngineAdapter` contract and rank-local manifest service
- add the first Megatron trainer adapter by reusing the existing NIXL registration and reshard manifest implementation
- organize Megatron implementations under canonical `modelexpress_rl/train/...` and `modelexpress_rl/inference/...` namespaces

## Why

The Redis-backed `RefitService` provides the version and shard control plane, but RL trainer actors still need a framework-facing client that connects their native engine tensors to those APIs. This PR adds that client boundary without wiring ModelExpress into a specific RL framework.

The intended lifecycle is:

1. The RL orchestrator creates a global `WeightVersion`.
2. Each trainer rank stages its local contribution through its engine adapter.
3. Each rank publishes its `WeightVersionShard` metadata and serves its immutable manifest.
4. The server marks the version `READY` after all expected source slots are published.
5. A future version-retirement integration will complete `source_reuse_ready`; until then the Megatron IN_PLACE implementation fails that fence explicitly instead of claiming the source buffers are reusable.

## API and implementation notes

- `ModelExpressTrainerClient.initialize()` resolves RL deployment policy from `modelexpress_rl.envs` and shared connectivity settings from `modelexpress.envs`, constructs the selected trainer adapter internally, and maintains worker registration.
- RL-specific environment configuration is limited to trainer engine, staging mode, and payload format; model/server/worker/heartbeat settings continue to use the shared ModelExpress environment module.
- the engine passes its live NIXL manager and manifest publisher but does not construct or supply a `TrainerEngineAdapter`.
- the Megatron manager dependency is typed by the narrow `agent_name`, `nixl_metadata`, and metadata-listener port protocol used by publication.
- the client has a simple no-argument constructor; `initialize()` remains the supported setup path.
- `worker_endpoint` identifies the trainer-side manifest service, while `server_url` identifies the central ModelExpress control plane.
- `stage_shard()` delegates engine-specific tensor discovery and staging to `TrainerEngineAdapter`.
- `StagedWeightVersionShard.publish()` is idempotent and publishes only control metadata; weight bytes remain on trainer-owned buffers.
- all staged buffer owners for a version remain retained; publishing another shard for the same version does not release an earlier allocation.
- `MegatronTrainerAdapter` currently supports `IN_PLACE` plus `FULL_TENSOR`; it derives the source slot from the initialized Megatron global rank and the NIXL metadata endpoint from the shared worker host plus the manager's listen port.
- Canonical Megatron-specific APIs live under `modelexpress_rl`; temporary re-exports at the prior `modelexpress.refit.reshard` paths preserve NeMo-RL #3632 compatibility while the shared reshard core remains engine-neutral.

## Validation

- `200 passed` across the environment, trainer-client, Megatron, and refit/reshard Python suite on the rebased, single-commit head
- `4 passed` against a real Redis 7 backend in the `RefitService` gRPC lifecycle suite, covering automatic readiness, replacement publishers, leases, release, shard deletion, and worker expiry
- a live cross-language E2E ran the Python `ModelExpressTrainerClient` and Megatron adapter against the Rust ModelExpress server and Redis backend: the trainer registered, published its shard, served its manifest through `RefitWorkerService`, and advanced the version from `STAGING` to `READY`
- the same E2E exercised a generator-like consumer that discovered the published shard, fetched the worker-local manifest, verified its SHA-256 digest and tensor metadata, registered and deleted a version lease, released the version to `RELEASING`, and deleted the source shard
- the E2E used synthetic registered tensor addresses and NIXL metadata; it validates the API and metadata lifecycle but does not claim a real GPU/NIXL byte transfer
- Python CI with protobuf 5 and protobuf 6, amd64 and arm64 client wheel builds, and the server image build passed on the prior head; those checks have restarted for this amended head
- Ruff passes for the changed source and test files, the generated gRPC stubs pass the reported `F541` check, and `git diff --check origin/main...HEAD` passes

## Not included

- direct NeMo-RL or another RL framework call-site wiring
- generator-side planning, transfer, or weight installation
- Megatron version-retirement signaling for IN_PLACE buffer reuse
- a real GPU/NIXL data-transfer qualification; this PR reuses the existing registered-buffer and manifest path

## Summary by CodeRabbit

* **New Features**
  * Added RL trainer support for registering workers, staging and publishing weight shards, renewing leases, and retrieving manifests.
  * Added Megatron training and inference resharding support with validation, shard layouts, aliases, and publication helpers.
  * Added configuration options for trainer engines, staging modes, payload formats, source slots, and metadata endpoints.
  * Added manifest retrieval through the refit service.
* **Documentation**
  * Expanded setup guidance for trainer weight-version publication and Megatron integration.
* **Refactor**
  * Consolidated Megatron resharding APIs while retaining compatibility aliases.
* **Tests**
  * Added coverage for trainer publication, manifest retrieval, validation, and environment settings.


modelexpress_client/python/modelexpress_rl/refit_pb2_grpc.py
```diff
@@ -0,0 +1,533 @@
...
+try:
+    from grpc._utilities import first_version_is_lower
+    _version_not_supported = first_version_is_lower(GRPC_VERSION, GRPC_GENERATED_VERSION)
+except ImportError:
+    _version_not_supported = True
+
+if _version_not_supported:
+    raise RuntimeError(
+        f'The grpc package installed is at version {GRPC_VERSION},'
+        + f' but the generated code in refit_pb2_grpc.py depends on'
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in cfaab7a at the generation source. generate_proto.sh now strips the redundant f prefix for both generated gRPC stubs; the checked-in outputs were updated, and Ruff F541 passes for both files.

zhengluo-nv:
@coderabbitai review

modelexpress_client/python/tests/test_reshard_megatron.py
```diff
@@ -371,6 +377,28 @@ def test_a_missing_fused_gate_up_order_is_rejected():
         )
 
 
+def test_gated_aliases_reject_inconsistent_declared_global_shape():
+    fused = torch.arange(32, dtype=torch.bfloat16).reshape(8, 4)
+
+    with pytest.raises(ValueError, match="linear_fc1.weight: derived gate/up shape"):
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in c1d4f31. Both pytest.raises patterns now use raw regex strings with escaped literal periods, so the assertions require the exact tensor names. Validation: 18 Megatron reshard tests passed, Ruff passed for the changed test file, and git diff --check passed.

KavinKrishnan:
Overall, this is a strong refactor. Separating the version lifecycle in `ModelExpressTrainerClient` from engine-specific staging in `TrainerEngineAdapter` gives NeMo-RL and other frameworks a much cleaner integration point.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/adapter.py
```diff
@@ -0,0 +1,116 @@
...
+                total_bytes=total_bytes,
+                transport="NIXL",
+            ),
+            # IN_PLACE performs no asynchronous copy, so the shard is ready to
+            # publish as soon as its manifest has been built.
+            publish_ready=CompletionFence(lambda: None),
+            # The live tensors cannot be reused until the published version is
+            # retired. That signal is not wired yet, so fail instead of
+            # incorrectly reporting that the source is safe to mutate.
+            source_reuse_ready=CompletionFence(_source_reuse_unsupported),
```
KavinKrishnan, line 112: `source_reuse_ready` correctly raises instead of claiming the weights are safe to modify. However, the trainer could start its next optimizer step and overwrite those weights while a receiver is still reading them.

What is expected to hold the trainer back until the version is retired? Could that requirement be documented here so callers do not accidentally corrupt an in-flight transfer?

modelexpress_client/python/modelexpress_rl/train/engines/megatron/aliases.py
```diff
@@ -136,10 +168,14 @@ def _build_gated_aliases(
 
 
 def _build_qkv_aliases(
-    item: MegatronAliasInput, agent_name: str
+    item: MegatronTensorSpec, agent_name: str
 ) -> list[PublishedTensor]:
     if len(item.hf_names) != 3 or item.tensor.ndim != 2:
         raise ValueError(f"{item.name}: QKV aliasing requires 2D q/k/v weights")
+    required = ("head_dim", "num_heads_local", "num_kv_heads_local")
```
KavinKrishnan, line 175: Heads up on this one — not something you broke, the code came over as-is in the move. Worth knowing about though, since this file is where it lives now.

It breaks on models with fewer KV heads than trainer ranks. Someone outside the team hit it last week on Nemotron Ultra, going Megatron TP8/EP64 → vLLM TP8/EP1. That model has 64 query heads and only 2 KV heads, and the publisher works out the local count by dividing, so 2 over 8 gives 0 and the check just below throws.

The obvious fix is bumping that 0 up to 1, and that's the trap. Once you have fewer KV heads than ranks, Megatron stops splitting on head boundaries — it chops the fused block into equal row chunks and glues it back together with an all-gather at runtime, so it can cut anywhere. Here the block is 8704 rows cut every 1088, so six of the eight ranks end up holding no K or V rows at all, and rank 0's chunk stops halfway through a query head. The loop below walks each group as `[q..., k, v]`, which just isn't what those ranks have. Bump it to 1 and you'd get past the check and then hand receivers the wrong bytes, which is worse than the crash.

What we think it needs: stop counting in whole heads here. Take the rows this rank actually owns, work out which bits overlap each group's q, k and v sections, and publish one entry per overlap. The publisher also has to start sending the model-wide head counts, because by the time we're here the local KV count has already been rounded down to nothing.

Happy to pick it up on our end — mostly just want it written down now that the code lives here.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/aliases.py
```diff
@@ -245,4 +281,8 @@ def build_hf_aliases(
     return aliases
 
 
-__all__ = ["MegatronAliasInput", "build_hf_aliases"]
+# Compatibility name for the existing reshard API.
+MegatronAliasInput = MegatronTensorSpec
```
KavinKrishnan, line 285: This keeps the old class name, but the old import paths themselves are removed. NeMo-RL #3632 still uses those paths, so it will fail to import as soon as this lands.

Could we keep temporary re-exports at the old paths and refactor after we merge the upstream PR?

modelexpress_client/python/modelexpress_rl/train/engines/megatron/adapter.py
zhengluo-nv: It relies on caller (megatron for example) to integrate with `source_reuse_ready` correctly. For IN_PLACE  weight staging, the workflow will be sync:

Create weight version → Trigger generator update → Wait for all generators finish update → Delete weight version → Resume training

modelexpress_client/python/modelexpress_rl/train/engines/megatron/aliases.py
zhengluo-nv: I haven't looked into the actual reshard logic. This is out of the scope of this PR.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/aliases.py
zhengluo-nv: Sure

modelexpress_client/python/modelexpress_rl/train/adapter.py
```diff
@@ -0,0 +1,155 @@
...
+    WRITE_TO_STORAGE = "WRITE_TO_STORAGE"
+    IN_PLACE = "IN_PLACE"
+
+
+class WeightPayloadFormat(str, Enum):
+    """Weight representation fixed for one initialized client."""
+
+    UNSPECIFIED = "UNSPECIFIED"
+    FULL_TENSOR = "FULL_TENSOR"
+    XOR_DELTA = "XOR_DELTA"
```
nv-hwoo, line 50: 1. XOR is one way to compute delta so let's just say `DELTA`.
2. Not sure how the weight payload format is used throughout the codebase but I want to make sure is that the payload format should not determine the transfer methods. For example, we can use both P2P weight transfer and delta approach to selectively transfer the weight tensors that have changed. So the payload is still the weight tensors, not deltas, but we still computed delta to choose which weights to transfer.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/adapter.py
zhengluo-nv: Addressed in 8f9437d8. The adapter comment and README now state the required synchronous IN_PLACE lifecycle explicitly: the framework must update every generator, retire the published version, and only then resume training or start the next optimizer step. The fence still fails explicitly until rank-local retirement signaling is implemented.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/aliases.py
zhengluo-nv: Tracked as #639: https://github.com/ai-dynamo/modelexpress/issues/639. The issue records the Nemotron Ultra TP8 example, explains why max(local_kv_heads, 1) is unsafe, and defines overlap-based publication plus reconstruction tests. This algorithmic limitation predates the module move and remains intentionally outside #616.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/aliases.py
zhengluo-nv: Addressed in 8f9437d8. Temporary re-export modules restore modelexpress.refit.reshard.megatron_aliases and modelexpress.refit.reshard.megatron_publisher. I verified the exact imports used by NeMo-RL #3632, added an import-identity regression test, and the broader RL/refit/reshard suite passes (200 tests).

zhengluo-nv:
Review follow-up: commit `8f9437d8` also addresses CodeRabbit's finding that failed to post inline. `MegatronTargetSpec` and `MegatronTargetLayout` now reject booleans and non-integral `global_shape`, `tp_size`, and `tp_rank` values before partitioning. Regression coverage is included; the final environment/trainer/Megatron/refit/reshard suite passes (`200 passed`), along with Ruff and `git diff --check`.

modelexpress_client/python/modelexpress_rl/train/adapter.py
zhengluo-nv: Good point—we should separate payload encoding, tensor selection, and transport. Computing deltas to select dirty tensors does not imply sending delta bytes; that case still uses FULL_TENSOR, with only the selected tensors included, over any supported transport. XOR_DELTA currently means the transferred bytes are XOR deltas and therefore require a base version. I’ll clarify this distinction. I prefer keeping the specific XOR_DELTA name until we introduce a generic delta encoding plus an explicit codec.

tanushriya910 (approved):

PR #624, opened 2026-08-13
feat: MTP loading improvements

emricksini-h:
## Description

Draft-shard selection did not work for checkpoints stored in an object store such as S3: runai's `pull_files` matches `allow_pattern` against the full object key, so the bare `model.safetensors.index.json` never matched and the draft fell back to streaming the whole checkpoint.

When only the draft-shards were downloaded, loading was then failing because the draft's retry rebuilds it with `initialize_model()`, and that rebuild cleared the layer registries on `compilation_config`. One `compilation_config` is shared by every model built from a VllmConfig, so the clear also unregistered the already-loaded target's layers, leaving its entries missing from `static_forward_context`.

### Proposed fix

1. Anchor the index glob (`*model.safetensors.index.json`), and warn instead of falling back silently.
2. Unregister only the layers owned by the model being discarded, instead of clearing the registries the target co-owns.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safetensors index discovery using precise filename matching.
  * Added clearer warnings when safetensors indexes are unavailable.
  * Preserved shared compilation state when reinitializing model components.
  * Improved cleanup of stale model registrations without affecting active components.
  * Enhanced selection of draft model shards from object storage.

* **Tests**
  * Added coverage for index matching, missing-index warnings, shard selection, and compilation-state cleanup.


zhengluo-nv:
/ok to test [9639786](https://github.com/ai-dynamo/modelexpress/pull/624/commits/9639786f7160aa8f2588515e607cb03ad5023cc3)

zhengluo-nv (approved):

PR #636, opened 2026-08-13
refactor(reshard): route device work through the backend

yafshar:
## Summary

`ReshardReceiver` assumed CUDA in two ways: it called `torch.cuda` for every
stage synchronize, and it scoped receive and staging buffer allocations into the
classic `cudaMalloc` pool unconditionally. Both now resolve from the
constructor's `device` via `AcceleratorBackend`, so the receiver names no
accelerator directly.

Adds `requires_classic_alloc_pool()` to the backend protocol. It states a
requirement, not a capability: CUDA needs registered buffers scoped into a
classic pool because its caching allocator under `expandable_segments` can
return VMM ranges that register successfully but fail during RDMA WRITE when
`nvidia_peermem` cannot pin the underlying pages. No equivalent hazard is known
or observed on XPU, so XPU uses the default allocator — successful XPU
registration does not prove the WRITE-time hazard absent. `torch.xpu` does
expose `MemPool` and `XPUPluggableAllocator`, so an alternate XPU pool could be
implemented if one is ever needed.

`registered_buffer_alloc_scope()` makes that selection from its own module,
leaving `cuda_pool.py` as purely the CUDA implementation. It raises
`NotImplementedError` if a backend requires a pool that isn't implemented,
rather than silently running CUDA code.

Also fixes a latent bug: the receiver built its `NixlTransferManager` without
passing a backend, so the manager fell back to its CUDA default. Harmless while
every target was CUDA, immediately fatal otherwise — an XPU target dies in
`torch.cuda.set_device` during `initialize()`, before it registers anything.
Building the test harness for this PR omitted the same argument on the publish
side, which confirmed the default-backend failure mode on hardware.

## Behavior change to sign off on

Before this change an XPU reshard target would eventually call
`torch.cuda.synchronize` and could not complete correctly. The receiver now
routes synchronization, allocation, and NIXL registration through the XPU
backend. That makes this more than a refactor: it changes operational behavior
for a non-CUDA target. The change is target-side, and it was exercised with both
tested publisher families. A CUDA receiver never encountered those target-side
assumptions.

An XPU receiver was then exercised end to end against a live synthetic
publisher, so this is validated rather than inferred.

Separately, and unchanged by this PR, no accelerator compatibility check governs
this path. The gap is documented with a `TODO(publisher-accelerator)`: the
rendezvous identity and shard table carry no publisher family, so
`accelerators_compatible` has nothing to compare. No pairing is rejected on
accelerator-family grounds; other NIXL, fabric, or model-geometry constraints
may still prevent transfer. Closing it means publishing the source family in the
shard table and comparing both endpoints, since compatibility is a property of
the source-target pair and a target-only check cannot express it.

## Compatibility

`requires_classic_alloc_pool()` is a new member of the `AcceleratorBackend`
protocol. It is not `@runtime_checkable`, so nothing isinstance-checks it; all
in-tree implementers (`cuda.py`, `xpu.py`, the test mock) are updated. An
out-of-tree implementer would raise `AttributeError` on the reshard path until
it adds the method.

## Testing

| Check | Result |
|---|---|
| Client suite, local | 1031 passed, 22 skipped |
| Client suite, CUDA node (torch 2.13.0+cu130) | 1052 passed, 1 skipped |
| Client suite, XPU node (torch 2.13.0+xpu) | 1031 passed, 22 skipped |
| `cargo check --workspace --tests` | clean |
| `cargo clippy --workspace --all-targets` | clean, no warnings |

The suite-count difference reflects optional dependencies and platform-specific
skips, not different test selection.

New `_prepare` regression test asserts all three allocation sites — convert
staging, full-pull staging, and receive buffers — go through the
backend-selected allocation scope. Mutation-checked: dropping any one of the
three call sites makes it fail.

### End-to-end refit on hardware

A live synthetic publisher plus a minimal `ReshardReceiver` subclass, over all
four accelerator pairings (publisher→receiver):

| pairing | result |
|---|---|
| cuda → cuda | all destinations exact |
| cuda → xpu | all destinations exact |
| xpu → cuda | all destinations exact |
| xpu → xpu | all destinations exact |

One payload covers all three transfer classes in a single plan — an exact
whole-tensor segment, a column narrow promoted to a full pull and re-sliced
locally, and a bf16-served source cast into an fp32 destination (widening, so
lossless, making exact equality the correct assertion). Every destination
matched the expected source-derived values exactly. Two refits per run, so the
cached-plan path is covered. Coverage gate required at 1.0. Plan shape was
identical on every run: 
`segments=3, exact_descriptors=10, descriptor_savings=7, full_pull_sources=1, converts=1, fallback=0`. 

Publisher digests were byte-identical from both families.

### Not covered

The engine-specific hooks. The harness supplies its own `_capture` and
`_install`, which are simpler than `VllmReshardReceiver`'s meta-twin capture and
PWAL install. The payload is also unquantized.

## Follow-ups (not in this PR)

- Publish the source accelerator in the shard table and gate on both endpoints
- Drop `classic_cuda_alloc` from `reshard/__init__.py` and `__all__` so the
  lazy import actually yields a CUDA-free import path on XPU
- Move the allocation scope onto the backend itself, which needs `cuda_pool.py`
  relocated out of `refit/reshard/`
- Publisher-side ergonomics: nothing steers a non-CUDA publisher toward passing
  `accelerator_backend` to `NixlTransferManager`, and omitting it is immediately
  fatal
- Land the end-to-end harness in-tree so the result is reproducible

## Summary by CodeRabbit

* **New Features**
  * Improved refit and reshard support across CUDA and XPU accelerators.
  * Backend-specific buffer allocation and synchronization are now applied automatically.
  * XPU uses its standard allocator, while CUDA uses the required classic allocation pool.
  * Added public support for selecting the appropriate registered-buffer allocation context.

* **Documentation**
  * Updated architecture and refit documentation with accelerator-specific behavior and current compatibility limitations.

* **Tests**
  * Expanded coverage for CUDA/XPU backend selection, allocation, synchronization, and transfer workflows.


yafshar:
@coderabbitai Trimmed `xpu.py` — it duplicated `cuda_pool.py`'s module docstring and the
allocator paragraph in `ARCHITECTURE.md`. Kept the "not a proof of absence" line:
the CUDA failure mode passes `ibv_reg_mr` and only fails at WRITE, so successful
XPU registration doesn't rule it out, and without that line the next reader
reasonably concludes XPU is proven safe.

Keeping the other two. The cited guideline is about not explaining *removed* code;
both of these explain code that is present:

- `alloc_scope.py` — justifies a function-level import, and records that the lazy
  import does *not* keep `cuda_pool` out of a non-CUDA process, since
  `reshard/__init__.py` re-exports `classic_cuda_alloc`. An earlier review round
  proposed that import specifically for XPU isolation; the note prevents someone
  re-deriving that it doesn't provide any.
- `test_reshard_refit_fused_wire.py` — marks a stub backend that reads as inert
  setup but is load-bearing; deleting it fails with an `AttributeError` several
  frames from the cause.

yafshar:
@zhengluo-nv could you take a look when you get a chance?

Merged with `main` (`aa2a6fc`), two conflicts in code #635 touched. Three reshard tests are updated mechanically: `batch_install` and `ceiling` needed `_backend` on their harnesses plus had dead `torch.cuda.synchronize` patches,
and `descriptor_cache` patched `classic_cuda_alloc`, which this PR replaces with `registered_buffer_alloc_scope`.

Re-validated post-merge on L40S/B200 + Arc B60: client suite 1453/4 skipped on CUDA, 1431/26 on XPU; reshard refit e2e `cuda→cuda` and `cuda→xpu` exact over two refits. All 22 CI checks green.

zhengluo-nv:
reshard feature is under refactoring.

yafshar:
@zhengluo-nv you noted on 2026-08-26 that the reshard feature is under refactoring. Following up now that this is green again, and to check whether it should still wait.

Current state: merged `main` through 79bf7b5, head 6c553fef, all 25 checks green. 

On the overlap question, this PR touches `accelerators/{base,cuda,xpu}.py`, `refit/reshard/{__init__,alloc_scope,receiver}.py`. Neither NCCL M2N branch collides with it: #497 adds `refit/reshard/transport/nccl_m2n/` and edits `transport/__init__.py`; #748 adds `modelexpress_rl/collective/` plus the new proto and Rust module and touches nothing under `refit/reshard/`. `transport/base.py` + `nixl.py` have been there since #496, so routing device work through the backend seems orthogonal to which transport lands.

Three questions:

1. Does the refactoring you mentioned refer to the M2N work (#748 vs #497), or to something separate that isn't on a branch yet?
2. Is `refit/reshard/transport/` expected to gain the M2N transport (#497's placement), or stay NIXL-only with M2N as a sibling stack (#748's)?
3. Either way, can this land now, or would you rather it wait on that decision?

Happy to rebase onto whichever placement wins if it's close.

PR #645, opened 2026-08-17
feat(metrics): Prometheus foundation for new metrics

yixinh-nv:
## Summary

Makes the Prometheus pipeline correct so that new metric families can be added on it. It
deliberately adds no new families itself, beyond `mx_build_info`, which later families join
against.

This covers the first two of six rollout steps — the eight issues below, which are the
prerequisite for everything after them.

| Step | Scope | Exit criterion |
| --- | --- | --- |
| **1. Server exposition** — *this PR* | `/metrics` listener on its own port, Helm repoint, `prometheus-client` into the images, `mx_build_info` (D1, D5) | `up == 1` on a pod that has served no traffic |
| **2. Client pipeline** — *this PR* | multiprocess mode, eager `enable()`, pod-scoped push, bounded labels, the selection funnel (D2–D4, D6–D8) | one port exposes merged series from every rank at TP=8, and survives an OOM-kill |
| 3. Request and backend coverage | one gRPC interceptor with in-handler outcomes plus one backend-operation wrapper — 19 server RPCs, 5 worker RPCs, 3 metadata backends in two changes | a simulated Redis outage surfaces as `outcome="backend_error"`, not as silence |
| 4. Export what is already computed | refit stage timings, cache statistics, registry status counts, NIXL agent health — all measured today and thrown away | refit stage sums match the existing JSON log record |
| 5. Load and transfer timing tiers | `mx_load_seconds` down to `mx_transfer_phase_seconds`, plus the paths that bypass the strategy chain | Σ L4 ≤ L3 ≤ L2 ≤ L1 ≤ L0 holds on a live multi-strategy run |
| 6. Deployment surface | gauges, PodMonitor, dashboards, recording rules, `absent()` alerts | a dashboard answers "why was this load slow" end to end |

Steps 3 and 4 are the cheap ones — 4 is almost free, since the code already computes those
numbers and discards them. Step 5 is the largest, and is what finally makes "why was this
model load slow" answerable.

## Problem

Model Express ships six client metrics and no server endpoint at all. Eight issues, each
of which fails **silently** — plausible-looking output or a clean `up == 0` rather than an
error.

| # | Issue |
| --- | --- |
| D1 | `prometheus.io/port` points at the tonic gRPC listener. tonic is HTTP/2 only; Prometheus scrapes HTTP/1.1, so every server pod reports `up == 0` forever — indistinguishable from a crashed pod. |
| D2 | All N ranks call `start_http_server` on one port; one binds, N-1 give up. The endpoint serves one rank's data as if it were the pod's, and vanishes if that rank exits first. |
| D3 | The Pushgateway grouping key is the hostname, and a push is a `PUT` that replaces the group — ranks on a host erase each other, and the result looks complete. |
| D4 | Init is reachable only from a recording call, so a run that skips P2P produces output byte-identical to `MX_METRICS_ENABLED=0` — the run you most need to diagnose. |
| D5 | No container image installs `prometheus-client` explicitly. Whether the collector works depends on the engine in the base image happening to pull it in as one of *its* dependencies — unpinned, and one base bump or one `--no-deps` rebuild away from silently disabling itself on the `ImportError` path. |
| D6 | The flush runs from `atexit`, which does not run on SIGKILL — an OOM-killed rank contributes nothing. |
| D7 | `source_worker_id` is a per-process `uuid4`, so its label domain grows with process count over time, not cluster size. |
| D8 | `mx_p2p_candidates{stage="listed"}` can never observe zero, and a `ListSources` RPC failure records nothing — a backend outage and an empty cluster are the same absence. |

## Fix

- **Server gets a `/metrics` listener** (`axum`, default port **9401**) plus `mx_build_info`.
  The chart *generates* the scrape annotation from `.Values.metrics.port` so it cannot drift
  back onto the gRPC port. (D1)
- **Client moves to `prometheus_client` multiprocess mode**: one endpoint serves the merged
  union of every rank, losing the bind is the normal case, and mmap writes are durable at
  increment time so no exit hook is in the path. (D2, D6)
- **Pushgateway is demoted** to an opt-in batch escape hatch: one push per pod, keyed on pod
  UID, from the merged registry, mutually exclusive with the scrape endpoint. (D3)
- **`enable()` is called unconditionally** from all three engine loaders, ahead of the
  eligibility filter. (D4)
- **`prometheus-client` installed explicitly in all ten client images** — the `[metrics]`
  extra, or its own install line where `--no-deps` protects the engine stack — with a
  `>=0.20.0` floor, and added to `[dev]` since CI installs only that. The collector no
  longer depends on which engine happens to be in the base image. (D5)
- **`source_worker_id` dropped** by default; `MX_METRICS_SOURCE_ID_LABEL=1` restores it for
  benchmark runs. (D7)
- **Selection funnel records zero**, plus a bounded
  `mx_p2p_list_sources_total{result=ok|empty|error}`, on the RDMA **and** SGLang
  `transfer_engine` discovery paths. (D8)

Two config traps fail silently:

- `MODEL_EXPRESS_SERVER_METRICS_PORT` reaches the server **only** via the clap
  `--metrics-port` override. `load_layered_config` reads env as
  `Environment::with_prefix("MODEL_EXPRESS").separator("_")`, so the name resolves to the key
  path `server.metrics.port`, matches no field, and serde drops it without a warning.
- `metrics_port` is a `u16`, not `NonZeroU16`, and carries `#[serde(default)]`.
  `load_layered_config` swallows deserialization errors and returns `T::default()` with no log
  line, so anything this field rejects costs *the whole config file* — gRPC port, cache
  directory, eviction policy and auth settings included. `0` is the documented disable value.

## Compatibility

- **Server** — additive; there was no endpoint before. `MODEL_EXPRESS_SERVER_METRICS_PORT=0`
  or `--set metrics.enabled=false` disables it, and a bind failure logs and continues rather
  than crashlooping the cache service.
- **`/metrics` is not behind `AuthLayer`.** That layer wraps the gRPC services; the metrics
  listener is a separate HTTP server, so it answers anyone who can reach the pod IP even under
  `security.mode=enforce`. Normal for an exporter, and pod-IP-only by default
  (`metrics.service` is `false`) — but worth knowing if you enabled SA auth.
- **Client** — still opt-in, still no-raise. With `MX_METRICS_ENABLED` unset nothing changes.
- **Breaking for existing dashboards:** `source_worker_id` is gone from
  `mx_p2p_source_selections_total` unless `MX_METRICS_SOURCE_ID_LABEL=1`. It does *not* error
  — PromQL groups everything under `source_worker_id=""`, so `sum by (source_worker_id) (...)`
  silently collapses N lines into one and reads as perfectly balanced. Group by `policy`.
- **Additive:** `mx_p2p_list_sources_total` and `mx_build_info`.
- **One example manifest is wired end to end** —
  `examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml` (TP=8) carries the scrape
  annotations, the memory-backed `emptyDir`, `PROMETHEUS_MULTIPROC_DIR`, `MX_METRICS_PORT`,
  the containerPort and an initContainer that clears stale files. The **other** example
  manifests and a PodMonitor are follow-up, so pods built from those still take the
  one-rank-per-endpoint fallback (with a warning saying so).

## Validation

- Ran `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt
  --all --check`, `cargo test --workspace --all-features`, and `cargo test -p
  modelexpress-server --features integration-tests`.
- Ran `pytest tests/` — 131 passing across the metrics, source-selection and SGLang suites.
- Ran `helm lint` and `helm template` against all five values files.
- Ran `pre-commit run --all-files`.
- One regression test per issue, asserting on the *exposition* rather than on calls. A
  multi-rank test forks real ranks at TP=2 and TP=8 and kills every one without running an
  exit hook (SIGKILL, and `os._exit` for rank 0), then asserts the merged total, that no
  `pid` label leaked, and that `mx_build_info` merges to 1 rather than N.
- A Rust integration test boots a real server and scrapes it over raw HTTP/1.1, and asserts
  the gRPC port does **not** answer one, so merging the ports regresses loudly.

### On H200 cluster

- **Phase 0** — a server pod that has served no traffic answers a cross-pod HTTP/1.1 scrape
  with `200`, the OpenMetrics content type and `mx_build_info`. The same scrape against the
  gRPC port fails with `curl: (1) Received HTTP/0.9 when not allowed` — D1, demonstrated
  rather than asserted.
- **Phase 1** — eight ranks in one pod behind one port scrape `360/360`, and still `360/360`
  after four of them are SIGKILLed mid-flight. `mx_build_info` merges to one series at 1
  rather than to 8, and no `pid` label leaks.

Measured by reconstructing the old mechanism and driving both through the same workload, in
the real `vllm-runtime` client image at TP=8. The harness is not part of the change — it did
its job at review time, and keeping code that models a mechanism no longer in the repo is a
liability:

| Scenario | Before | After |
| --- | --- | --- |
| Events reaching a scrape of the pod's one endpoint | 12.5 % mean (2.8–22.2 %) | 100 % |
| Events retained after the pod pushes to a gateway | 10–22 %, varying by run | 100 % |
| Events surviving when every rank is SIGKILLed | 0 % | 100 % |

Both "before" figures are ranges rather than points, and that *is* the issue: the survivor
was whichever rank won the race to `bind()`, or happened to `PUT` last.

Enabling metrics costs the load path 1.7 µs per recording call (195 ns disabled), against a
load that makes thousands of them, not millions.

Scrape cost grows with the mmap files left behind by exited ranks; the trade-off that
governs and the measured curve are in `docs/METRICS.md`, since it is an operator choice
rather than a review one.

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for server and client components.
  * Added a dedicated HTTP `/metrics` endpoint, configurable by port and scheme.
  * Added build, source-selection, and multiprocess aggregation metrics.
  * Added Helm and container configuration for metrics scraping and service exposure.
* **Documentation**
  * Added comprehensive metrics deployment, configuration, troubleshooting, and verification guidance.
  * Updated the README roadmap and documentation links.
* **Bug Fixes**
  * Improved metrics reliability during retries, shutdowns, worker failures, and hard-kill scenarios.


benchmarks/metrics/metrics_pipeline.py
nv-hwoo, line 1: what is this `benchmarks` directory for? do we need this under our python client package?

modelexpress_server/src/server.rs
```diff
@@ -53,6 +62,32 @@ pub async fn run_server(
...
+    })?;
+
+    // Start the Prometheus listener before anything else can fail, so a scrape
+    // proves the exporter came up even on a server that never reaches a healthy
+    // state. It is a separate HTTP/1.1 listener on its own port because tonic
+    // serves HTTP/2 only — a scrape aimed at the gRPC port can never succeed.
+    // Its shutdown is signalled at the very end of this function, after the gRPC
+    // server has drained, so the drain window stays scrapeable.
+    let (metrics_shutdown_tx, metrics_shutdown_rx) = tokio::sync::oneshot::channel();
+    let metrics_handle = metrics_addr.map(|metrics_addr| {
```
KavinKrishnan: I think there’s one more listener cleanup path to handle. If registry, refit, or P2P setup fails after the metrics listener starts, `run_server` returns early and drops the `JoinHandle`, which leaves the listener running. A later retry can then fail to bind port 9401. Could we use a small cleanup guard so every return path stops the listener?

KavinKrishnan:
One scope question: this PR adds the multi-rank metrics path, but none of the shipped manifests set `PROMETHEUS_MULTIPROC_DIR` or `MX_METRICS_PORT`. Without those, users still get the single-rank fallback rather than merged pod metrics. Could we wire these into an example manifest here, or make it clearer that deployment wiring is still a follow-up?

modelexpress_client/python/modelexpress/metrics.py
```diff
@@ -178,26 +486,183 @@ def observe_transfer_seconds(self, policy: str, outcome: str, seconds: float) ->
...
+            "PROMETHEUS_MULTIPROC_DIR=%s is not usable (%s); metrics will be "
+            "empty. Mount it as an emptyDir in the pod manifest.",
+            multiproc_dir,
+            e,
+        )
+        return
+    try:
+        from prometheus_client import values
+
+        if not getattr(values.ValueClass, "_multiprocess", False):
```
KavinKrishnan: This check depends on the private `prometheus_client.values.ValueClass._multiprocess` field, while the dependency has no upper version bound. If that internal field changes, healthy setups could produce a misleading warning. Could we check for an actual `.db` file after the first metric write instead, or otherwise guard the supported versions?

modelexpress_server/src/server.rs
yixinh-nv: done. Added a `MetricsListener` guard — `shutdown()` drains on the happy path, `Drop` aborts on every other — plus a test driving an early backend failure.

One correction: the bind failure does not reproduce. Dropping the guard also drops its oneshot sender, which fires the graceful shutdown and frees the socket. What actually leaked was the detached task, so that is what the guard bounds; the doc comment and test say so rather than claiming a bind failure.

modelexpress_client/python/modelexpress/metrics.py
yixinh-nv: done. Swapped the private attribute for an on-disk check: after the families are built, `_warn_if_multiprocess_inactive` looks for `.db` files. Empty means the value class latched single-process, which is the thing that actually matters and survives any upstream rename. Tests now cover both directions.

yixinh-nv:
done. Wired into `examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml` — TP=8, so the case the fallback costs most: scrape annotations, memory-backed `emptyDir`, `PROMETHEUS_MULTIPROC_DIR`, `MX_METRICS_PORT`, containerPort, and an initContainer for the reset. `docs/METRICS.md` links to it.

Two caveats noted inline: the initContainer runs on pod start, not on an app-only container restart, and the other manifests plus a PodMonitor are still follow-up.

examples/model_streamer_k8s/client/sglang/Dockerfile
```diff
@@ -12,6 +12,12 @@ RUN pip uninstall -y modelexpress 2>/dev/null || true && \
     rm -rf "$(python3 -c 'import site; print(site.getsitepackages()[0])')"/modelexpress* && \
     pip install --no-cache-dir --no-deps .
 
+# --no-deps above excludes the [metrics] extra, so prometheus-client is
+# installed explicitly; the collector otherwise catches the ImportError and
+# disables itself. It has no dependencies of its own, so --no-deps keeps the
+# engine-provided stack untouched.
+RUN pip install --no-cache-dir --no-deps "prometheus-client>=0.20.0"
```
nv-hwoo: I don't think we need to install prometheus by default to these dockerfiles since it's an opt-in feature and the example dockerfiles are installing modelexpress with `--no-deps`.

examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml
nv-hwoo, line 1: I would suggest creating a separate yaml for enabling prometheus and add a section in the readme about it, instead of changing the existing example.

examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml
```diff
@@ -45,7 +45,31 @@ spec:
...
+        - name: reset-metrics-dir
+          image: <your-registry>/modelexpress-client:latest
+          imagePullPolicy: IfNotPresent
+          command: ["python3", "-m", "modelexpress.metrics", "--reset"]
+          env:
+            - name: PROMETHEUS_MULTIPROC_DIR
+              value: /var/run/mx-metrics
+          volumeMounts:
+            - name: mx-metrics
+              mountPath: /var/run/mx-metrics
```
nv-hwoo: This looks like a bad devX. Is there any alternatives than enforcing the users to specify this when they enable prometheus in MX?

modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -87,6 +88,13 @@ def load_model(
         """
         load_start = time.perf_counter()
 
+        # Unconditionally, and before the strategy eligibility filter below: a
+        # run that skips P2P and falls back to a local or HuggingFace path must
+        # still bring the exporter up, or it produces output byte-identical to
+        # MX_METRICS_ENABLED=0 -- and that is the run you most need to diagnose.
+        # No-op unless MX_METRICS_ENABLED=1; never raises.
+        enable_metrics()
```
nv-hwoo: should we move this out of the model loading path? probably not a big deal but wouldn't hurt to do these in the init instead

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -277,6 +280,7 @@ def _find_source_instances(
         ``random``). The retry slice (MAX_SOURCE_RETRIES) is applied by the
         caller in load(), so the selector controls ordering only.
         """
+        policy = configured_policy_label()
```
nv-hwoo, line 283: is this relevant to this PR?

modelexpress_client/python/modelexpress/metrics.py
```diff
@@ -178,26 +491,204 @@ def observe_transfer_seconds(self, policy: str, outcome: str, seconds: float) ->
...
+        try:
+            return Gauge(
+                name, documentation, labels, registry=registry, multiprocess_mode=mode
+            )
+        except (ValueError, TypeError):
+            continue
+    return Gauge(name, documentation, labels, registry=registry)
+
+
+def enable() -> bool:
```
nv-hwoo: nit: I don't think `enable` is a great function name. Can we rename it to be more useful/specific.

examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml
yixinh-nv: done. `vllm-single-node.yaml` is reverted back, and `vllm-single-node-metrics.yaml` is its sibling with each addition marked `# metrics:`. Added a section to the examples README.

modelexpress_client/python/modelexpress/engines/vllm/loader.py
yixinh-nv: done. Moved to `MxModelLoader.__init__` in all three loaders — still before any strategy decision, but once per loader instead of per load.

modelexpress_client/python/modelexpress/metrics.py
yixinh-nv: done. Renamed to `enable_metrics()`.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
yixinh-nv: Yes — it's what lets the `except` branch record. `configured_policy_label()` used to be computed inside the `try`, so a `ListSources` RPC failure had no policy label in scope and recorded nothing. That silence is half of D8: a backend outage and an empty cluster looked identical.

benchmarks/metrics/metrics_pipeline.py
yixinh-nv: done. Moved to a top-level `benchmarks/metrics/`. It was never packaged (`packages.find` matches `modelexpress*` only) but you're right it didn't belong in the client tree.

examples/model_streamer_k8s/client/sglang/Dockerfile
yixinh-nv: done — reverted all of them.

examples/p2p_transfer_k8s/client/vllm/vllm-single-node.yaml
yixinh-nv: done — cut it to three env vars and a port. The volume, mount and initContainer were unnecessary: the ranks are processes in one container, so a plain path like `/tmp/mx-metrics` is already shared, and a container restart recreates it empty (which is what the initContainer was for). The `emptyDir` is now documented as optional, for tmpfs and a size cap.

The variable itself has to stay in the manifest — `prometheus_client` fixes its value class at import, so setting it from Python lands after the engine already imported it.

benchmarks/metrics/metrics_pipeline.py
nv-hwoo: But what is this benchmark script for? I'm curious why we should maintain this benchmark script for prometheus in our repo.

yixinh-nv: Sorry I think this can be removed since it was used to produce the benchmarking results in the PR description.

yixinh-nv: Just deleted.

nv-hwoo (approved):

PR #646, opened 2026-08-18
chore(deps): bump h2 to 0.4.16 for RUSTSEC-2026-0258

nicolasnoble:
h2 0.4.12 is affected by RUSTSEC-2026-0258, unbounded memory growth from a peer sending empty DATA frames. The advisory requires 0.4.16 or later.

This is a lockfile-only change. h2 is not declared in any workspace manifest; it comes in transitively through hyper, reqwest and tonic, so `cargo update -p h2` resolves it without touching a Cargo.toml. Exactly one package moves.

cargo audit fires the advisory on 0.4.12 and is silent on 0.4.16. `cargo check --workspace --all-targets` is clean.

One thing worth knowing for anyone looking at a green Security Audit on another open PR right now: cargo-audit pulls the advisory database at run time, so any run predating this advisory reports clean against the same vulnerable lockfile. Those greens say nothing about the current advisory set, and each of those PRs picks this fix up on its next merge from main.


tanushriya910:
> Exactly one package moves
seeing some other packages like window-sys getting downgraded as well

nicolasnoble:
yeah, good catch. cargo update retargeted six references it had no business touching, three consumers to socket2 0.5.10 and three to windows-sys 0.52.0. reverted those, so the diff is now just the h2 version and checksum. cargo check --workspace --locked is clean on it, so the lock doesn't need that rewrite to build.

zhengluo-nv (approved):

tanushriya910 (approved):

PR #647, opened 2026-08-18
feat(refit): add FSDP trainer publisher for RL reshard

tanushriya910:
Add an FSDP/DTensor trainer-engine adapter so FSDP trainers can publish weight versions through MX's reshard backend, alongside Megatron.

## What's new

- `engines/fsdp/publisher.py` extracts each rank's local shards from an FSDP `state_dict`. Unsharded and replicated tensors publish in full from every rank; sharded DTensors publish their per-dim local box via `compute_local_shape_and_global_offset`. HF-name conversion stays out of this layer, the receiver owns that mapping.
- `engines/fsdp/adapter.py` wires that extraction into the trainer contract, with two staging modes:
  - COPY_TO_DEVICE (default): allocates one persistent wire-dtype arena per shard, registers it once, and snapshots the live weights into it each step. Safe against a moving/re-materialized source since the registered address never changes.
  - IN_PLACE: registers the DTensor's local storage directly and serves it with no copy. Requires the source to already be contiguous and in the wire dtype, and fails fast toward COPY_TO_DEVICE if the source storage moves between steps.
- `train/adapter.py`: `NixlMetadataProvider` gains `register_tensors`, since NIXL can only transfer memory that's been registered, and an adapter needs to register its own staging arenas or in-place buffers before publishing.
- `rendezvous.build_sources`: unsharded/replicated tensors get published redundantly by every rank, so the same box can show up more than once. `build_sources` now dedups to one shard per distinct box, picked at random, since the copies are byte-identical and randomizing spreads reads across the publishing ranks' NICs.

## Known gaps (follow-ups, not blocking)

- Version retirement isn't wired up yet, so `source_reuse_ready` fails rather than claiming the source is safe to mutate (mirrors the Megatron adapter's current state).
- Arena registration is one dmabuf MR per shard, not a single pooled `VmmArena`.

## Testing

- `test_refit_fsdp_publisher.py`: shard extraction (skips non-float, handles unsharded/replicated/sharded), manifest building, contiguity and endpoint validation.
- `test_refit_fsdp_adapter.py`: both staging modes, moved-source rejection for IN_PLACE, staging-mode/tensor-set immutability after `initialize`, unsupported mode/format rejection.
- `test_reshard_refit_rendezvous.py`: `build_sources` dedup behavior for redundant boxes vs. genuinely distinct boxes.

## Summary by CodeRabbit

* **New Features**
  * Added support for publishing and staging FSDP tensor shards for weight transfer.
  * Added manifest generation with shard, tensor, device, and memory metadata.
  * Added tensor registration for refreshing transfer metadata.
  * Improved replicated-shard handling by selecting one source for duplicate regions and distributing reads across available sources.

* **Bug Fixes**
  * Added validation for unsupported formats, invalid layouts, changed shard sets, and unsafe source reuse.

* **Tests**
  * Added coverage for FSDP publishing, staging, validation, manifests, and replicated-shard selection.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 import base64
 import json
 import logging
+import random
 import time
 import uuid
 from dataclasses import dataclass
@@ -176,14 +177,25 @@ def build_sources(tensors: list) -> tuple:
     Returns ``(sources, session_to_agent, session_to_device)`` where ``sources``
     is ``{src_name: SourceInfo}`` for ``plan_transfer`` and the two maps drive
     ``NixlReshardTransport``. Each shard's ``session`` is its owning agent name.
+
+    Unsharded/replicated sources are published by every trainer rank, so a source
+    can arrive with several shards covering the SAME box. The planners require a
+    single non-overlapping cover, so keep one shard per distinct box, chosen at
+    random -- the copies are identical, and randomizing spreads a replicated
+    source's reads across the publishing ranks' NICs without assuming any
+    rank/agent ordering.
     """
     sources = {}
     session_to_agent = {}
     session_to_device = {}
     for t in tensors:
         dtype = _torch_dtype(t.dtype)
-        shards = []
+        by_box: dict = {}
         for s in t.shards:
+            by_box.setdefault((tuple(s.shard_offset), tuple(s.shape)), []).append(s)
+        shards = []
+        for group in by_box.values():
+            s = random.choice(group)
             session = s.agent_name
             shards.append(
                 Shard(
--- a/modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
@@ -36,7 +36,6 @@ from __future__ import annotations
 import base64
 import json
 import logging
-import random
 import time
 import uuid
 from dataclasses import dataclass
@@ -177,25 +176,14 @@ def build_sources(tensors: list) -> tuple:
     Returns ``(sources, session_to_agent, session_to_device)`` where ``sources``
     is ``{src_name: SourceInfo}`` for ``plan_transfer`` and the two maps drive
     ``NixlReshardTransport``. Each shard's ``session`` is its owning agent name.
-
-    Unsharded/replicated sources are published by every trainer rank, so a source
-    can arrive with several shards covering the SAME box. The planners require a
-    single non-overlapping cover, so keep one shard per distinct box, chosen at
-    random -- the copies are identical, and randomizing spreads a replicated
-    source's reads across the publishing ranks' NICs without assuming any
-    rank/agent ordering.
     """
     sources = {}
     session_to_agent = {}
     session_to_device = {}
     for t in tensors:
         dtype = _torch_dtype(t.dtype)
-        by_box: dict = {}
-        for s in t.shards:
-            by_box.setdefault((tuple(s.shard_offset), tuple(s.shape)), []).append(s)
         shards = []
-        for group in by_box.values():
-            s = random.choice(group)
+        for s in t.shards:
             session = s.agent_name
             shards.append(
                 Shard(
--- a/modelexpress_client/python/modelexpress_rl/train/engines/fsdp/adapter.py
+++ b/modelexpress_client/python/modelexpress_rl/train/engines/fsdp/adapter.py
@@ -74,8 +74,10 @@ class FSDPTrainerAdapter(TrainerEngineAdapter):
         self._source_slot_id = f"publisher:global-rank:{dist.get_rank()}"
         self._initialized = False
         self._staging_mode: TrainerStagingMode | None = None
-        # Populated once by initialize(); keyed by tensor name.
-        self._expected_names: frozenset[str] = frozenset()
+        # name -> (global_shape, shard_offset, local_shape) fixed at initialize().
+        self._expected_layout: dict[
+            str, tuple[tuple[int, ...], tuple[int, ...], tuple[int, ...]]
+        ] = {}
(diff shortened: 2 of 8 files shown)


modelexpress_client/python/modelexpress_rl/train/engines/fsdp/adapter.py
```diff
@@ -0,0 +1,278 @@
...
+        """Fix the shard layout and register the source buffers (idempotent)."""
+        if self._initialized:
+            return
+        if staging_mode not in self.supported_staging_modes:
+            raise NotImplementedError(
+                f"FSDPTrainerAdapter does not support {staging_mode.value} staging"
+            )
+        self._expected_names = frozenset(s.name for s in shards)
+        if len(self._expected_names) != len(shards):
+            raise ValueError("FSDP shard names are not unique within this rank")
```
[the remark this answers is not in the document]
tanushriya910: updated in new commit

modelexpress_client/python/modelexpress_rl/train/engines/fsdp/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+    ranks; the receiver dedups them by box (keeping one per distinct box, chosen
+    at random) so the extra copies only spread reads across NICs, they do not
+    corrupt fan-in.
+    """
+    shards: list[LocalTensorShard] = []
+    for name, value in state_dict.items():
+        if not value.is_floating_point():
+            continue
+        full_shape = tuple(value.shape)
+        zero_offset = tuple(0 for _ in full_shape) or (0,)
```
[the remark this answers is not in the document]
tanushriya910: addressed in new commit

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
```diff
@@ -176,14 +177,25 @@ def build_sources(tensors: list) -> tuple:
...
     session_to_device = {}
     for t in tensors:
         dtype = _torch_dtype(t.dtype)
-        shards = []
+        by_box: dict = {}
         for s in t.shards:
+            by_box.setdefault((tuple(s.shard_offset), tuple(s.shape)), []).append(s)
+        shards = []
+        for group in by_box.values():
+            s = random.choice(group)
```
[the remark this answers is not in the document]
tanushriya910: discover_trainers already filters to SOURCE_STATUS_READY, and handshake_with_peers retries each selected peer with backoff for the whole refit timeout. The remaining gap is narrower i.e. a peer that's truly dead but still READY (stale until the heartbeat reaper catches up) being the one candidate kept for a box, which then times out the whole handshake.
The replanning code for addressing dead trainers is going to be updated as part of the refactoring API work so this might be redundant work at this point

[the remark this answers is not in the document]
tanushriya910: already being tracked as part of refactoring work

KavinKrishnan: One heads-up: #635 already dedups these same `(offset, shape)` boxes in `merge_shard_tables`, which production calls right before this function. Once both land, every group here has one item, so the random pick does not spread reads across NICs. Could we leave this out of the FSDP PR and handle balancing in `merge_shard_tables` separately? A stable hash of receiver rank and tensor name would spread reads without changing randomly between runs.

KavinKrishnan (commented):
The FSDP adapter direction looks good. I checked out the branch and all 30 focused tests pass; the scalar path also makes it through the planner correctly. The main thing to sort out is the replica dedup overlap with #635 below. The address check is small but worth fixing here; step stamping and configurable wire dtype can be follow-ups.

modelexpress_client/python/modelexpress_rl/train/engines/fsdp/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+
+    by_name: dict[str, PublishedTensor] = {}
+    for shard in shards:
+        served = shard.served_tensor
+        if not served.is_contiguous():
+            raise ValueError(f"{shard.name}: served tensor must be contiguous for RDMA")
+        published_shard = PublishedShard(
+            agent_name=agent_name,
+            device_id=served.device.index if served.device.type == "cuda" else 0,
+            addr=served.data_ptr(),
```
KavinKrishnan: Could we reject `served.data_ptr() <= 0` here, like the Megatron publisher does? A meta or unregistered tensor currently publishes address 0 and only fails later in NIXL.

modelexpress_client/python/modelexpress_rl/train/engines/fsdp/publisher.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Every rank publishes what it holds, including the tensors it holds in full
+    (unsharded and replicated). Those redundant full copies are identical across
+    ranks; the receiver dedups them by box (keeping one per distinct box, chosen
+    at random) so the extra copies only spread reads across NICs, they do not
+    corrupt fan-in.
+    """
+    shards: list[LocalTensorShard] = []
+    for name, value in state_dict.items():
+        if not value.is_floating_point():
```
KavinKrishnan, line 84: Can we at least report which non-float entries were skipped? Right now int and bool buffers disappear silently and show up later as a coverage miss, which is hard to trace.

modelexpress_client/python/modelexpress/refit/reshard/rendezvous.py
tanushriya910: removed it in the new commit, will rely on #635 to address the overlaps

modelexpress_client/python/modelexpress_rl/train/engines/fsdp/publisher.py
tanushriya910: addressed in the new commit

modelexpress_client/python/modelexpress_rl/train/engines/fsdp/publisher.py
tanushriya910: added a debug log

zhengluo-nv (approved):

PR #648, opened 2026-08-18
feat(refit): add RL generator client and staged vLLM installation

zhengluo-nv:
## Summary

- add synchronous, framework-facing `ModelExpressControlClient`, `ModelExpressTrainerClient`, and `ModelExpressGeneratorClient` APIs for immutable RL weight versions
- add exact-version generator discovery, leases, transfer-plan validation/rebuild, staged NIXL transfer, full-payload digest verification, and framework-controlled apply
- add the first generator engine implementation for vLLM, preserving vLLM's load-time/PWAL behavior and CUDA-graph-bound storage
- keep the new RL path independent from the legacy `ReshardReceiver`: transfer and verification live in one private engine-neutral module, while vLLM capture and installation live in one private engine-specific module

## Companion integration

- NeMo-RL integration: [NVIDIA-NeMo/RL#3704](https://github.com/NVIDIA-NeMo/RL/pull/3704)
- the GPU E2E below validated this MX head `3cb98ac7` together with NeMo-RL PR head `d26b92c0`

## Architecture

The RL orchestrator uses the public control client. Framework-native RPCs invoke rank-local trainer and generator clients inside their existing actors. The MX server stores control metadata only; manifests stay on workers and weight bytes move directly over NIXL.

```mermaid
flowchart LR
    subgraph Framework["RL framework"]
        O["RL orchestrator<br/>ModelExpressControlClient"]
        T["Trainer actor<br/>ModelExpressTrainerClient"]
        G["Generator actor<br/>ModelExpressGeneratorClient"]
    end

    S["MX server<br/>RefitService<br/>Redis metadata"]

    subgraph Trainer["Trainer rank"]
        TA["Trainer engine adapter"]
        B["Registered source buffers"]
        M["RefitWorkerService<br/>manifest endpoint"]
    end

    subgraph Generator["Generator rank"]
        GA["Generator engine adapter"]
        X["NIXL staged transfer<br/>plan, pull, verify"]
        I["vLLM installer<br/>capture, apply"]
    end

    O -->|"Create / Get / Delete version"| S
    O -->|"Framework-native RPC"| T
    O -->|"Framework-native RPC"| G
    T --> TA --> B
    T -->|"Register worker; publish shard metadata"| S
    T --> M
    G -->|"Register worker; discover shards; lease version"| S
    G --> GA --> X --> I
    G -->|"Fetch exact-version manifest"| M
    B -->|"NIXL reads"| X
```

The generator adapter composes:

- `modelexpress_rl/inference/nixl_staged_transfer.py`: exact-manifest decoding, planning, reusable registered buffers, NIXL reads, transforms, and verification
- `modelexpress_rl/inference/engines/vllm/installer.py`: unquantized meta-twin geometry capture and graph-safe vLLM installation through layerwise reload and PWAL

This avoids adding RL state and policy to the legacy discovery-oriented receiver. The current worker manifest is still the existing opaque rendezvous envelope, but decoding is isolated inside the staged-transfer module; replacing it with a typed RL manifest is a follow-up rather than another concern added to the vLLM adapter.

## Version lifecycle

1. The orchestrator creates a `STAGING` weight version with all expected trainer source slots.
2. Each trainer rank stages its native buffers and publishes one worker-local manifest. The final required slot makes the version `READY`.
3. Each generator rank registers a lease, fetches the exact version's shard publications and manifests, then validates its cached physical fingerprint.
4. If the fingerprint changed, the generator rebuilds its transfer plan. It pulls into separate registered buffers and verifies complete full-tensor coverage and source digests before apply.
5. At the framework-approved safe point, the vLLM installer applies the verified buffers and refreshes graph-bound or derived state.
6. Staged handles and leases are released independently; the orchestrator can retire the version after the update.

### Automated checks

- full Python suite: **1,356 passed, 33 skipped**
- focused RL refit, Megatron publication, staged-transfer, generator-client, and vLLM adapter suite after the final digest fix: **59 passed, 1 skipped**
- targeted Ruff and formatting checks passed
- `git diff --check` passed
- commit is GPG-signed and DCO-signed off

### NeMo-RL GPU E2E

Validated the final PR head `3cb98ac7` with NeMo-RL `d26b92c0` using image `nvcr.io/0980761089281446/model-express-dev-containers:nemo-rl-mx-3cb98ac7-d26b92c02-b200-20260819-r3` (`sha256:fc14b6f451e4668ca51ef0b4e267e4d714290cb2855fd825906c50cd856fb2a3`).

- topology: two B200 nodes, one GPU per node; non-colocated Megatron TP1 trainer and vLLM TP1 generator
- workload: one synchronous Llama 3.1 8B GRPO step with four generated responses
- transport: direct trainer-to-generator NIXL over UCX (`rc,cuda_copy`)
- strict capture: **291 copies, 0 unsupported sources, 0 unattributed copies**
- staged transfer completed with full source-digest verification before vLLM apply
- training completed successfully: total step **27.73 s**, transfer/update **12.51 s**, average reward **1.0**
- lifecycle audit: the completed version reached `RELEASING` with no remaining shard or lease records
- run-specific Kubernetes resources were removed; the shared Refit server and Redis were preserved

A non-fatal NeMo-RL interpreter-shutdown warning remains after successful completion; the driver exited `0` and the warning is outside the refit data path.

## Summary by CodeRabbit

* **New Features**
  * Added RL weight-version lifecycle management for creating, viewing, and deleting versions.
  * Added generator support for staged weight transfers, version validation, lease management, and safe application.
  * Added vLLM support for staged tensor installation and configurable generator settings.
  * Simplified trainer setup with automatic transport and manifest-service management.
* **Documentation**
  * Updated architecture and usage documentation for the RL refit workflow and package structure.
* **Tests**
  * Added coverage for lifecycle operations, staged transfers, environment settings, and vLLM integration.


modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -0,0 +1,498 @@
...
+        raise ValueError("model_config is required for the vLLM generator adapter")
+
+    from .engines.vllm import VllmGeneratorAdapter
+
+    return VllmGeneratorAdapter(
+        model=model,
+        vllm_config=vllm_config,
+        model_config=model_config,
+        worker_id=worker_id,
+    )
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in `5a023595`. The public client now accepts a typed engine context through `ModelExpressGeneratorConfig`. `VllmGeneratorContext` and vLLM adapter construction/validation live under `inference/engines/vllm/`; the generic layer dispatches by the concrete context type and imports no vLLM runtime APIs.

modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -0,0 +1,498 @@
...
+            )
+        digest = hashlib.sha256(response.manifest).hexdigest()
+        if (
+            response.manifest_digest != shard.manifest_digest
+            or digest != shard.manifest_digest
+        ):
+            raise RuntimeError(
+                f"manifest digest mismatch for source slot {shard.source_slot_id!r}"
+            )
+        return response.manifest
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed by making the current security boundary explicit in `docs/ARCHITECTURE.md`. The initial worker manifest protocol remains plaintext and unauthenticated, so it is supported only on a trusted, network-isolated cluster where untrusted clients cannot reach the MX server or worker gRPC endpoints. Manifest digest validation remains in place; TLS and caller authentication are documented as a follow-up protocol/deployment change rather than implied by the digest.

modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -0,0 +1,498 @@
...
+            stop.set()
+            renewal.join()
+            self._service.DeleteVersionLease(
+                refit_pb2.DeleteVersionLeaseRequest(
+                    version_id=version.uid,
+                    lease_id=lease.lease_id,
+                    worker_id=self.worker_id,
+                ),
+                timeout=self._rpc_timeout_seconds,
+            )
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in the current client lifecycle. A version lease now remains owned by the staged handle through apply/release. Staging, apply, and release cleanup paths preserve an in-flight primary exception when `DeleteVersionLease` fails and log the cleanup failure; a lease deletion error is surfaced only when there is no primary error to mask.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/installer.py
```diff
@@ -0,0 +1,256 @@
...
+            name
+            for name, parameter in self._model.named_parameters()
+            if parameter.device.type == "meta"
+        ]
+        if meta_parameters:
+            logger.error(
+                "vLLM refit left %d parameters on the meta device: %s",
+                len(meta_parameters),
+                meta_parameters[:10],
+            )
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed. The post-install meta-device check now raises `IncompleteRefit` with the count and parameter names, so the update fails closed instead of reporting success with unbacked parameters.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/installer.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Resolve captured parameter names through the live module hierarchy;
+        # PWAL is a layer operation rather than a whole-model operation.
+        groups: dict[Module, list[tuple[str, str]]] = {}
+        names = set(tensors)
+        for module_name, module in self._model.named_modules():
+            for leaf, _parameter in module.named_parameters(recurse=False):
+                full_name = f"{module_name}.{leaf}" if module_name else leaf
+                if full_name in names:
+                    groups.setdefault(module, []).append((full_name, leaf))
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed. `_VllmInstaller._process_and_commit()` now calls `initialize_layerwise_reload()` before resolving parameter groups, resolves against the restored load-time hierarchy, and raises `IncompleteRefit` if any staged tensor name is unmatched.

modelexpress_client/python/modelexpress_rl/inference/nixl_staged_transfer.py
```diff
@@ -0,0 +1,564 @@
...
+        return _StagedNixlWeights(
+            tensors=self._recv_buffers,
+            metrics={
+                "bytes_received": sum(d.nbytes for d in prepared.descriptors),
+                "segments": len(prepared.descriptors),
+                "wire_s": round(wire_seconds, 6),
+                "full_pull_sources": len(prepared.plan.full_pulls),
+                "converts": len(prepared.plan.converts),
+            },
+        )
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed at both layers. The NIXL transfer assigns a monotonically increasing generation to prepared and staged state. The vLLM adapter rejects plan replacement while a staged weight is active and validates generation plus active-handle identity before apply or release, preventing stale handles from accessing reused buffers.

modelexpress_client/python/modelexpress_rl/train/client.py
```diff
@@ -5,18 +5,19 @@
 
 from __future__ import annotations
 
+import os
 import threading
 import uuid
-from dataclasses import dataclass
-from typing import Any
+from typing import Any, Self
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed. Public return annotations now use their enclosing class names, so these modules no longer import or depend on `typing.Self` and remain importable on Python 3.10.

modelexpress_client/python/modelexpress_rl/train/client.py
```diff
@@ -195,18 +208,54 @@ def initialize(
...
+        try:
+            client._register_worker()
+            client._registration_thread = threading.Thread(
+                target=client._renew_worker_registration,
+                name=f"modelexpress-refit-renew-{worker_id}",
+                daemon=True,
+            )
+            client._registration_thread.start()
+        except Exception:
+            client.close()
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed. If registration-thread startup fails, `_registration_thread` is cleared before the exception reaches `close()`, allowing channel and owned trainer resources to be cleaned up without joining an unstarted thread. The failure path has focused test coverage.

modelexpress_client/python/tests/test_refit_generator_client.py
```diff
@@ -0,0 +1,303 @@
...
+    ]
+
+
+def test_generator_releases_lease_when_manifest_is_invalid(monkeypatch):
+    server, endpoint, service = _start_server(manifest_digest="bad-digest")
+    adapter = _Adapter(service)
+    generator = _initialize(monkeypatch, endpoint, adapter)
+
+    try:
+        with pytest.raises(RuntimeError, match="no usable source.*digest mismatch"):
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed. The `pytest.raises(..., match=...)` pattern is now a raw string, and Ruff passes.

modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -0,0 +1,567 @@
...
+        if lease_ttl_seconds is None:
+            lease_ttl_seconds = registration_ttl_seconds
+        if registration_ttl_seconds <= 0:
+            raise ValueError("registration_ttl_seconds must be positive")
+        if lease_ttl_seconds <= 0:
+            raise ValueError("lease_ttl_seconds must be positive")
+        if config.max_transfer_attempts <= 0:
+            raise ValueError("max_transfer_attempts must be positive")
+        if config.rpc_timeout_seconds <= 0:
+            raise ValueError("rpc_timeout_seconds must be positive")
```
nv-hwoo: We can run the positive number check functions defined in the envs, instead of checking it here.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/adapter.py
```diff
@@ -0,0 +1,155 @@
...
+if TYPE_CHECKING:
+    from torch.nn import Module
+    from vllm.config import ModelConfig, VllmConfig
+
+
+@dataclass(frozen=True)
+class _VllmTransferPlan:
+    transfer: _PreparedNixlTransfer
+    fingerprint: tuple
+    generation: int
```
nv-hwoo: does transfer plan need to be engine specific?

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/adapter.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@dataclass(frozen=True)
+class _VllmStagedWeight:
+    transfer: _StagedNixlWeights
+    generation: int
```
nv-hwoo: what is generation? is it version?

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/adapter.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        plan = _VllmTransferPlan(
+            transfer=transfer,
+            fingerprint=inputs.physical_fingerprint,
+            generation=transfer.generation,
+        )
+        self._active_plan = plan
+        return plan
+
+    def validate_transfer_plan(
```
nv-hwoo, line 86: I feel like a lot of the functions here can be moved to the base class to be shared by both engines if we know the sglang will also define the same transfer plan and staged weights with same fields. Is this true?

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/context.py
```diff
@@ -0,0 +1,26 @@
...
+    from torch.nn import Module
+    from vllm.config import VllmConfig
+
+
+@dataclass(frozen=True)
+class VllmGeneratorContext(GeneratorEngineContext):
+    """Live vLLM objects required to install weights on one generator rank."""
+
+    model: Module
+    vllm_config: VllmConfig
```
nv-hwoo, line 23: are these all we need to install weights? what about information about the staged weights?

modelexpress_client/python/modelexpress_rl/inference/engines/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+def _create_generator_adapter(
+    *,
+    engine_context: GeneratorEngineContext,
+    worker_id: str,
+) -> GeneratorEngineAdapter:
+    """Construct the adapter selected by the concrete engine context."""
+    if isinstance(engine_context, VllmGeneratorContext):
+        return _create_vllm_adapter(engine_context, worker_id)
+    raise TypeError(f"unsupported generator context {type(engine_context).__name__}")
```
nv-hwoo, line 20: if this is going to be like a factory for engine adapters, do we need some kind of flag to distinguish the engine types?

modelexpress_client/python/modelexpress_rl/train/engines/megatron/selection.py
```diff
@@ -0,0 +1,172 @@
...
+
+def model_express_role_for_mapping(
+    *, mapping: Any, megatron_module: Any, tensor_ndim: int
+) -> str:
+    """Return the MX role for one supported Megatron-Bridge mapping."""
+    mapping_name = type(mapping).__name__
+    if getattr(mapping, "is_expert", False):
+        raise NotImplementedError(
+            f"ModelExpress does not yet support expert mapping {mapping_name}"
+        )
```
nv-hwoo, line 25: does our current resharding not support MoE models?

modelexpress_client/python/modelexpress_rl/inference/client.py
zhengluo-nv: Updated. Moved to ModelExpressGeneratorConfig

modelexpress_client/python/modelexpress_rl/inference/engines/__init__.py
zhengluo-nv: It uses the type of engine_context to distinguish the engine types.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/adapter.py
zhengluo-nv: Not engine specific. Updated

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/context.py
zhengluo-nv: Yes, these are the rank-local construction inputs. Staged weights are runtime output from stage_weight(), not initialization context, so upstream frameworks should not construct or pass them.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/adapter.py
zhengluo-nv: I renamed it to plan_revision.  It is process-local and changes whenever `prepare` rebuilds
the plan. It is not a control-plane weight-version number.

modelexpress_client/python/modelexpress_rl/train/engines/megatron/selection.py
zhengluo-nv: Will support MoE models as follow-up. My agent says "A prior development branch implemented related MoE support, but it was never merged into current main" so I think Kavin has already figured it out, but not merged

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/adapter.py
zhengluo-nv: Yeah it's true. Right now don't know what to share. Will do after sglang adapter is implemented

tanushriya910 (commented):
did we check for correctness or also perf differences, wondering if we are seeing any differences in update/transfer time with the refactor

tanushriya910 (approved):

PR #649, opened 2026-08-18
fix(nixl): deregister memory before destroying the agent

KavinKrishnan:
## What this fixes

Destroying a NIXL agent while its memory is still registered aborts the process inside `ucp_worker_destroy`. That takes down the whole Ray worker rather than failing just the teardown.

In practice this hit at the end of **every** GRPO run that used the reshard refit path — after all training work had completed. A run that had actually succeeded still exited looking like a crash, which makes it hard to tell a real failure from a cosmetic one in CI logs.

## The fix

Retain the registered memory descriptors and release them in reverse registration order before the agent is destroyed.

```mermaid
flowchart LR
    S[shutdown] --> D[deregister retained descriptors<br/>reverse order]
    D --> A[destroy agent]
    A --> O[clean exit]
```

Two edge cases are covered because both actually occur:

- **Partial setup** — registration failed midway, so only some descriptors exist. Releasing must not assume a complete set.
- **Repeated shutdown** — teardown can be reached twice, so it has to stay idempotent rather than double-releasing.

## Review guide

Small and self-contained: `modelexpress/nixl_transfer.py` (+38/-9) and `tests/test_nixl_peer_lifecycle.py` (+24).

The thing worth checking is ordering: every registered region must be released *before* agent destruction, in all three paths (normal, partial, repeated).

## Test plan

- `PYTHONPATH=. python3 -m pytest tests/test_nixl_peer_lifecycle.py -q` — 28 passed
- Observed on 32 GPUs: GRPO runs on the reshard path now exit cleanly instead of aborting in `ucp_worker_destroy`

## Context

Split out of #635, which review asked to land gradually rather than in one piece. This part is independent of the rest — it touches only `nixl_transfer.py` and its test, and it does not depend on the other split PRs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource cleanup during shutdown by deregistering retained memory allocations in the correct order.
  * Teardown now continues safely if an individual cleanup operation fails.
  * Ensured memory tracking is cleared and the underlying agent is released consistently.

* **Tests**
  * Added coverage for shutdown ordering, cleanup, and agent release behavior.

### DIFF
--- a/modelexpress_client/python/modelexpress/nixl_transfer.py
+++ b/modelexpress_client/python/modelexpress/nixl_transfer.py
@@ -140,6 +140,10 @@ class NixlTransferManager:
         self._metadata: bytes = b""
         self._tensor_descriptors: list[TensorDescriptor] = []
         self._tensors: dict[str, torch.Tensor] = {}
+        # Registration descriptors must be deregistered before destroying the
+        # UCX-backed NIXL agent. Dropping an agent with live GPU registrations
+        # can abort inside ucp_worker_destroy during framework teardown.
+        self._registered_memory: list[Any] = []
         # Remote agents this manager has loaded, so shutdown can disconnect them.
         # Maps agent name -> (ip, port) for agents reached over the P2P socket, or
         # None for agents loaded from a metadata blob.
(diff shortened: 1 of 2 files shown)


KavinKrishnan:
Closing — folding this back into #635 rather than reviewing it separately.

The split would have needed three separate approvals, and reviewer availability is the real constraint here, not diff size. One PR with a clear review guide is the faster path.

Nothing is lost: the change is in #635 unchanged, and the readability concern that prompted the split is addressed there by giving the fused QKV row layout a name (`_QkvLayout` / `bands()`) instead of leaving it as index arithmetic inside a nested loop.

PR #651, opened 2026-08-18
feat(metrics): per-RPC and storage-backend metrics on the server

yixinh-nv:
## Summary

Adds per-RPC and storage-backend metrics to the server. This is step 3 of the six in #645,
and the first step that adds real families rather than plumbing.

The point of it is one thing: **make a backend outage visible.** Today it is not.

| Step | Scope | Exit criterion |
| --- | --- | --- |
| 1–2. Pipeline (#645) | exposition path, `mx_build_info`, client multiprocess mode | `up == 1`; merged series at TP=8 |
| **3. Request and backend coverage** — *this PR* | one tower layer over 21 routed methods, decorators over 3 backend traits | a backend error surfaces as `outcome="backend_error"`, not as silence |
| 4. Export what is already computed | refit stage timings, cache statistics, registry status counts, NIXL health | refit stage sums match the existing JSON log record |
| 5. Load and transfer timing tiers | `mx_load_seconds` down to `mx_transfer_phase_seconds` | Σ L4 ≤ L3 ≤ L2 ≤ L1 ≤ L0 on a live run |
| 6. Deployment surface | gauges, PodMonitor, dashboards, recording rules | a dashboard answers "why was this load slow" |

## Problem

Four `P2pService` handlers report failure **in band**: they catch the error, log it, and
return `Ok` carrying `success: false` or an empty list. That is deliberate and clients
depend on it — but it means the gRPC status code says `Ok` throughout.

`ListSources` is the clearest case. These two are byte-identical on the wire:

- Redis is down.
- No peers have published yet.

Both are `Ok(ListSourcesResponse { instances: [] })`. So the obvious implementation — a
status-code-derived interceptor — would report **100% success through a total backend
outage**, while the client independently misfiles the same outage as `metadata_miss`. The
metric would be worse than no metric: confidently wrong during the incident it exists to
explain.

Nothing else on the server is covered either. There are no RPC counts, no handler latency,
and no visibility into Redis or the kube API, so "is the server slow, or is the store slow"
is unanswerable.

## Fix

**One tower layer**, applied once at `Server::builder()`, covering all 21 routed methods.
It resolves the outcome in three steps:

1. an `RpcOutcome` the handler tagged onto its own response;
2. otherwise a `tonic::Status` in the response extensions;
3. otherwise success.

Step 2 is why this is cheap. `Status::into_http` already does
`response.extensions_mut().insert(self)`, so **every** `Err(Status)` path self-reports —
including auth rejections and tonic's `unimplemented` fallback. All nine `RefitService`
handlers, which fail honestly, needed no edit. Only handlers that would otherwise misreport
were touched.

**Three decorators** for the backend traits, implementing the same trait and delegating.
The Redis, Kubernetes and in-memory implementations are untouched, and a backend added later
is instrumented by construction rather than by remembering to add timing code.

To be explicit about scope, since the problem above is told through P2P: this half is
**server-wide, not P2P-only**. `mx_backend_ops_total` covers all three storage subsystems --
P2P source metadata, the model registry, and the refit weight-version store -- distinguished
by the `store` label. P2P leads the narrative only because that is where the in-band-failure
problem is sharpest.

**The registry is hoisted out of `MetricsListener::spawn`**, which built it and froze it
into an `Arc` in one expression. `Registry::register` takes `&mut self` with no interior
mutability, so nothing could register a family after that point.

| Family | Type | Labels |
| --- | --- | --- |
| `mx_grpc_requests_total` | Counter | `method`, `outcome` |
| `mx_grpc_request_seconds` | Histogram | `method`, `outcome` |
| `mx_grpc_requests_in_flight` | Gauge | `method` |
| `mx_backend_ops_total` | Counter | `store`, `op`, `result` |
| `mx_backend_op_seconds` | Histogram | `store`, `op`, `result` |
| `mx_backend_ops_in_flight` | Gauge | `store`, `op` |

`method` is a closed set of the 21 known routes plus `other`, so an unrecognised path cannot
mint a series.

**The two streaming RPCs are deliberately not recorded.** `EnsureModelDownloaded` and
`StreamModelFiles` return their response head as soon as the stream is set up, and report
failure as a stream item -- `EnsureModelDownloaded` sends `ModelStatus::ERROR` as an item.
The layer's future resolves at the head, so recording there would report `outcome="ok"` and a
sub-millisecond duration for a download that ran for forty minutes and failed. That is the
same confidently-wrong number this PR exists to remove, so they are absent rather than wrong.
`Health/Watch` is excluded for the same reason. Covering them means instrumenting the
response body and reading the trailer, which is its own change. `store` is the subsystem (`p2p`, `registry`, `refit`), not the storage engine
— only one engine is live per pod, so the engine stays on `mx_build_info{backend=...}` and
joins from there.

## Two deltas from the plan in #645

Worth calling out, because #645's roadmap table said otherwise:

- **The 5 Python `WorkerService` RPCs are not here.** They are a client-side family
  (`mx_worker_grpc_requests_total`) and belong with the client work, so this PR is
  Rust-server-only.
- **The backend half is three decorators, not one wrapper.** `MetadataBackend` (9 methods),
  `RegistryBackend` (12) and `RefitBackend` (9) share no supertrait and have three different
  error types, so no blanket impl is possible. That is 30 explicit forwards and the bulk of
  the diff. A macro would shrink it to roughly a third and be harder to review than the code
  it replaced.

## Testing

- **Exit criterion, end to end** — `an_in_band_failure_is_not_recorded_as_a_success` drives a
  real client against a real server, and asserts the scrape shows `outcome="invalid_argument"`
  for a call whose gRPC status was `Ok`. This is also the only test that can cover the
  propagation step: `tonic::Response::into_http` is `pub(crate)`, so no unit test outside
  tonic can drive it.
- **The forwarding trap** — `list_workers_filtered` is the one trait method with a default
  body, and only Redis overrides it. A decorator that inherited the default would silently
  bypass the pipelined implementation and double-count every call. Asserted with a mock that
  requires `list_workers` to be called exactly zero times.
- **Encoded names** — counters are registered *without* `_total` because the encoder appends
  it; the tests assert the exported names, and that nothing ends in `_total_total`.
- **Cardinality** — unknown paths collapse to `other`.
- **Cancellation** — a request dropped mid-flight records `outcome="cancelled"` and writes no
  latency sample, so the histogram is not silently conditioned on completion; the in-flight
  gauges are asserted to return to zero on every exit path.
- Both load-bearing assertions were mutation-tested: flipping the `list_sources` tag to `Ok`,
  and disabling the tag insertion entirely, each fail the relevant test.

- **Route coverage** — a test parses the protos and fails the build if any routed RPC lacks a
  `method_label` arm. Note what this does *not* cover: a new **streaming** RPC that does get a
  `method_label` arm but is not added to `resolves_at_head` would still be recorded at the
  head. The limitation is stated on `resolves_at_head` itself; closing it needs
  streaming-ness derived from the protos rather than from a hand-maintained list.

280 server unit tests (`--features integration-tests`), 5 integration tests, `cargo clippy
--workspace --all-targets --all-features -- -D warnings` clean.

## Verified against a real backend

The unit tests prove the mechanism; only a real outage proves the exit criterion. Run against
a Redis on an internal dev cluster, with the server driven by the generated Python stubs.

**The exit criterion holds.** With Redis healthy, 200 `ListSources` calls recorded
`outcome="ok"`. Redis was then force-deleted and the calls repeated. On the wire those calls
still returned `Ok` with an empty `instances` list -- indistinguishable from a healthy empty
cluster -- and the metric said otherwise:

```
mx_grpc_requests_total{method="P2pService/ListSources",outcome="backend_error"} 2
mx_backend_ops_total{store="p2p",op="list_workers_filtered",result="error"} 2
```

The gRPC layer and the backend decorator reached that verdict independently, by different
routes, and agree. A status-code-derived metric would have recorded both calls as successes.

**The forwarding trap is real and is avoided.** During the healthy phase the counters read
`op="list_workers_filtered"` 200 and `op="list_workers"` 4 -- the latter only from the
reaper, which genuinely calls it. Had the decorator inherited the default body, the 200 would
have appeared under both names and Redis's pipelined implementation would never have run.

**Overhead is below this harness's noise floor.** Four interleaved rounds of 3000 RPCs
against the pre-change binary and this one, on the in-memory backend so nothing but the layer
differs, on a path that returns before touching any backend:

| | p50 | p95 | mean |
| --- | --- | --- | --- |
| `main` (29e2f7cd) | 0.095-0.096 ms | 0.150-0.158 ms | 0.101-0.104 ms |
| this PR | 0.095-0.101 ms | 0.149-0.179 ms | 0.102-0.113 ms |

p50 is identical. Run-to-run variance within each arm exceeds the difference between them,
and the newer binary was faster on mean in two of the four rounds, so the honest statement is
that the layer's cost is not resolvable at this sample size rather than any specific figure.

**Scrape cost** at full label saturation for the exercised paths: 140 series, 13.7 KB,
0.9-1.2 ms per scrape.

## Compatibility

Additive. No existing family, label or endpoint changes, and no configuration is required —
the families register on the same registry and appear on the same `/metrics` port added in
 #645.

Recording is on whenever the server runs, including when the metrics listener is disabled
(`--metrics-port 0`); only the listener is conditional. The alternative — an `Option` at
every instrumentation site — costs a second, untested code path to save four `Arc`
allocations.

Handler behaviour is unchanged. The outcome tag is metrics-only: it does not touch the
response body or the status code, and dropping it degrades the metric to `outcome="ok"`
rather than breaking the call.

### DIFF
--- a/modelexpress_server/src/metrics/backend.rs
+++ b/modelexpress_server/src/metrics/backend.rs
@@ -1,0 +1,355 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the storage backends behind the server.
+//!
+//! Three unrelated traits sit between the server and its storage --
+//! `MetadataBackend` (P2P source metadata), `RegistryBackend` (model registry)
+//! and `RefitBackend` (weight versions) -- each with its own error type and no
+//! common supertrait. They are instrumented by decorators that implement the
+//! same trait and delegate, so the concrete Redis, Kubernetes and in-memory
+//! implementations are untouched and a new backend is instrumented by
+//! construction rather than by remembering to add timing code.
+//!
+//! `store` names the **subsystem**, not the storage engine. Only one engine is
+//! live per pod, so `store="redis"` would be a constant; the engine is already
+//! carried by `mx_build_info{backend=...}` and joins on from there. Naming the
+//! subsystem also keeps `op="connect"` unambiguous, which it would not be if two
+//! subsystems both reported it against the same engine.
+
+use std::future::Future;
+use std::time::Instant;
+
+use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue, LabelValueEncoder};
+use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::metrics::gauge::Gauge;
+use prometheus_client::metrics::histogram::Histogram;
+use prometheus_client::registry::Registry;
+
+use super::buckets;
+
+/// Which storage subsystem performed the operation.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub enum Store {
+    /// P2P source metadata (`MetadataBackend`).
+    P2p,
+    /// Model registry (`RegistryBackend`).
+    Registry,
+    /// Weight-version store for RL refit (`RefitBackend`).
+    Refit,
+}
+
+impl Store {
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::P2p => "p2p",
+            Self::Registry => "registry",
+            Self::Refit => "refit",
+        }
+    }
+}
+
+// Hand-written for the same reason as `RpcOutcome`: the derive would encode the
+// variant identifier, giving `P2p` rather than `p2p`.
+impl EncodeLabelValue for Store {
+    fn encode(&self, encoder: &mut LabelValueEncoder<'_>) -> Result<(), std::fmt::Error> {
+        EncodeLabelValue::encode(&self.as_str(), encoder)
+    }
+}
+
+/// Whether the operation succeeded.
+///
+/// Two values only. The error *kind* is deliberately not a label: the three
+/// backends have three unrelated error types, two of them boxed
+/// `dyn Error`, so any faithful mapping would be an unbounded string.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub enum OpResult {
+    /// The operation returned `Ok`.
+    Ok,
+    /// The operation returned `Err`.
+    Error,
+    /// The caller was dropped before the operation finished. Recorded from a
+    /// drop guard, so a store that hangs is visible rather than absent.
+    Cancelled,
+}
+
+impl OpResult {
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::Ok => "ok",
+            Self::Error => "error",
+            Self::Cancelled => "cancelled",
+        }
+    }
+}
+
+impl EncodeLabelValue for OpResult {
+    fn encode(&self, encoder: &mut LabelValueEncoder<'_>) -> Result<(), std::fmt::Error> {
+        EncodeLabelValue::encode(&self.as_str(), encoder)
+    }
+}
+
+/// Label set for the backend families.
+#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct BackendLabels {
+    /// Storage subsystem.
+    pub store: Store,
+    /// Trait method name. Bounded because every call site passes a literal.
+    pub op: &'static str,
+    /// Success or failure.
+    pub result: OpResult,
+}
+
+/// Label set for in-flight tracking. No `result`: an operation still running
+/// does not have one yet.
+#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct BackendOpLabels {
+    /// Storage subsystem.
+    pub store: Store,
+    /// Trait method name.
+    pub op: &'static str,
+}
+
+fn fast_histogram() -> Histogram {
+    Histogram::new(buckets::FAST)
+}
+
+type LatencyFamily = Family<BackendLabels, Histogram, fn() -> Histogram>;
+
+/// Handles for the backend families. Cloning shares the underlying storage.
+#[derive(Clone)]
+pub struct BackendMetrics {
+    ops: Family<BackendLabels, Counter>,
+    latency: LatencyFamily,
+    in_flight: Family<BackendOpLabels, Gauge>,
+}
+
+impl BackendMetrics {
+    /// Register the families on `registry` and return handles to them.
+    ///
+    /// Registered without the `_total` suffix; the encoder appends it.
+    pub fn register(registry: &mut Registry) -> Self {
+        let ops = Family::<BackendLabels, Counter>::default();
+        let latency: LatencyFamily =
+            Family::new_with_constructor(fast_histogram as fn() -> Histogram);
+        let in_flight = Family::<BackendOpLabels, Gauge>::default();
+
+        let backend = registry.sub_registry_with_prefix("backend");
+        backend.register(
+            "ops",
+            "Storage backend operations by subsystem, method and result",
+            ops.clone(),
+        );
+        backend.register(
+            "op_seconds",
+            "Storage backend operation latency by subsystem, method and result",
+            latency.clone(),
+        );
+
+        // See the note in `metrics::grpc`: a plain Gauge is safe here because the
+        // server is one process with an in-process registry, so there is no
+        // SIGKILLed rank whose gauge could wedge.
+        backend.register(
+            "ops_in_flight",
+            "Storage backend operations currently running, by subsystem and method",
+            in_flight.clone(),
+        );
+
+        Self {
+            ops,
+            latency,
+            in_flight,
+        }
+    }
+
+    /// Mark one operation as started.
+    fn in_flight_inc(&self, store: Store, op: &'static str) {
+        self.in_flight
+            .get_or_create(&BackendOpLabels { store, op })
+            .inc();
+    }
+
+    /// Mark one operation as finished, however it finished.
+    fn in_flight_dec(&self, store: Store, op: &'static str) {
+        self.in_flight
+            .get_or_create(&BackendOpLabels { store, op })
+            .dec();
+    }
+
+    /// Count an operation whose future was dropped before it completed.
+    ///
+    /// The counter only, never the histogram: a partial duration would enter the
+    /// distribution as a fast sample for an operation that was in fact still
+    /// running, which biases the tail in exactly the wrong direction.
+    fn record_cancelled(&self, store: Store, op: &'static str) {
+        self.ops
+            .get_or_create(&BackendLabels {
+                store,
+                op,
+                result: OpResult::Cancelled,
+            })
+            .inc();
+    }
+
+    /// Await `future`, then record its latency and result.
+    ///
+    /// The recording happens strictly after the await. That ordering is
+    /// load-bearing: `Family::get_or_create` returns a `!Send` guard, so holding
+    /// one across an await would make the caller's future non-`Send` and break
+    /// the `#[async_trait]` bound with an error pointing at lock internals.
+    pub async fn time<T, E, F>(&self, store: Store, op: &'static str, future: F) -> Result<T, E>
+    where
+        F: Future<Output = Result<T, E>>,
+    {
+        // A hung store is precisely what this histogram exists to reveal, and a
+        // hang is precisely the case that produces no sample: whoever is waiting
+        // gives up, this future is dropped, and the recording below never runs.
+        // The guard turns that into a counter increment, and holds the in-flight
+        // gauge so a wedged store is visible while it is still wedged rather than
+        // only afterwards.
+        let mut guard = InFlightGuard::enter(self.clone(), store, op);
+        let started = Instant::now();
+        let outcome = future.await;
+        guard.pending = false;
+        let labels = BackendLabels {
+            store,
+            op,
+            result: if outcome.is_ok() {
+                OpResult::Ok
+            } else {
+                OpResult::Error
+            },
+        };
+        // `as_secs_f64` is float division; `Instant - Instant` would trip
+        // `arithmetic_side_effects`.
+        let elapsed = started.elapsed().as_secs_f64();
+        self.ops.get_or_create(&labels).inc();
+        self.latency.get_or_create(&labels).observe(elapsed);
+        outcome
+    }
+}
+
+/// Tracks one in-flight operation, and records a cancellation if the future is
+/// dropped before it completed. See the equivalent in [`crate::metrics::grpc`].
+struct InFlightGuard {
+    metrics: BackendMetrics,
+    store: Store,
+    op: &'static str,
+    /// Cleared once the operation resolves. Still set at drop means cancelled.
+    pending: bool,
+}
+
+impl InFlightGuard {
+    fn enter(metrics: BackendMetrics, store: Store, op: &'static str) -> Self {
+        metrics.in_flight_inc(store, op);
+        Self {
+            metrics,
+            store,
+            op,
+            pending: true,
+        }
+    }
+}
+
+impl Drop for InFlightGuard {
+    fn drop(&mut self) {
+        // Runs on every exit path, so the gauge cannot drift.
+        self.metrics.in_flight_dec(self.store, self.op);
+        if self.pending {
+            self.metrics.record_cancelled(self.store, self.op);
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::metrics::{encode_text, new_registry};
+
+    #[tokio::test]
+    async fn both_results_encode_under_the_documented_names() {
+        let mut registry = new_registry();
+        let metrics = BackendMetrics::register(&mut registry);
+
+        let ok: Result<(), ()> = metrics
+            .time(Store::P2p, "list_workers", async { Ok(()) })
+            .await;
+        assert!(ok.is_ok());
+
+        let failed: Result<(), ()> = metrics
+            .time(Store::Registry, "connect", async { Err(()) })
+            .await;
+        assert!(failed.is_err());
+
+        let encoded = encode_text(&registry).unwrap_or_else(|_| String::from("<encode failed>"));
+
+        assert!(
+            encoded
+                .contains(r#"mx_backend_ops_total{store="p2p",op="list_workers",result="ok"} 1"#),
+            "{encoded}"
+        );
+        assert!(
+            encoded.contains(
+                r#"mx_backend_ops_total{store="registry",op="connect",result="error"} 1"#
+            ),
+            "{encoded}"
+        );
+        assert!(
+            encoded.contains("mx_backend_op_seconds_bucket"),
+            "{encoded}"
+        );
+        assert!(!encoded.contains("_total_total"), "{encoded}");
+    }
+
+    /// A dropped operation is counted, and contributes no latency sample.
+    #[tokio::test]
+    async fn a_dropped_operation_is_recorded_as_cancelled() {
+        let mut registry = new_registry();
+        let metrics = BackendMetrics::register(&mut registry);
+
+        // Polled, then dropped mid-flight -- the guard is created inside the
+        // async body, so it exists only once the future has been polled at least
+        // once. That is the real shape of a cancellation: a caller gives up on a
+        // store that stopped answering, exactly as the startup timeouts in
+        // `run_server` do.
+        let timed_out = tokio::time::timeout(
+            std::time::Duration::from_millis(10),
+            metrics.time(
+                Store::P2p,
+                "get_metadata",
+                std::future::pending::<Result<(), ()>>(),
+            ),
+        )
+        .await;
+        assert!(timed_out.is_err(), "the inner future should never resolve");
+
+        let encoded = encode_text(&registry).unwrap_or_else(|_| String::from("<encode failed>"));
+        assert!(
+            encoded.contains(
+                r#"mx_backend_ops_total{store="p2p",op="get_metadata",result="cancelled"} 1"#
+            ),
+            "{encoded}"
+        );
+        assert!(
+            !encoded.contains(r#"mx_backend_op_seconds_count{store="p2p",op="get_metadata"#),
+            "a cancellation wrote a latency sample: {encoded}"
+        );
+        // The guard decrements on every exit path, so a cancelled operation must
+        // not leave the gauge stuck above zero -- that is the failure mode a
+        // gauge has and a counter pair does not.
+        assert!(
+            encoded.contains(r#"mx_backend_ops_in_flight{store="p2p",op="get_metadata"} 0"#),
+            "in-flight gauge did not return to zero: {encoded}"
+        );
+    }
+
+    #[tokio::test]
+    async fn the_inner_value_is_passed_through_untouched() {
+        let mut registry = new_registry();
+        let metrics = BackendMetrics::register(&mut registry);
+
+        let value: Result<u32, ()> = metrics.time(Store::Refit, "get", async { Ok(7) }).await;
+        assert_eq!(value, Ok(7));
+    }
+}
--- a/modelexpress_server/src/metrics/grpc.rs
+++ b/modelexpress_server/src/metrics/grpc.rs
@@ -1,0 +1,706 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Per-RPC metrics for every gRPC service the server exposes.
+//!
+//! # Why the outcome is not the gRPC status code
+//!
+//! Several handlers report failure *in band*: they catch a backend error, log
+//! it, and return `Ok` with an empty or defaulted response body. `ListSources`
+//! is the clearest case — a Redis outage and "no peers have published yet" are
+//! the same `Ok(ListSourcesResponse { instances: [] })` on the wire. A metric
+//! derived from the status code would therefore read 100% success straight
+//! through a total backend outage, which is precisely the incident it exists to
+//! surface.
+//!
+//! So the outcome is resolved in three steps, in order:
+//!
+//! 1. An [`RpcOutcome`] the handler itself inserted into the response
+//!    extensions. This is the authoritative value and the only one that can see
+//!    an in-band failure.
+//! 2. Otherwise a [`tonic::Status`] in the response extensions. Every
+//!    `Err(Status)` path puts itself there — `Status::into_http` does
+//!    `response.extensions_mut().insert(self)` — so handlers that fail honestly
+//!    need no edit at all, and neither do auth rejections or tonic's own
+//!    `unimplemented` fallback for an unrouted path.
+//! 3. Otherwise success. A successful unary response carries `grpc-status` in
+//!    the HTTP/2 trailers, not the head, so its absence here is the success
+//!    signal.
+//!
+//! Step 2 is why instrumenting 21 routed paths costs a handful of one-line tags
+//! rather than 21 rewrites: only handlers that lie about their outcome need
+//! touching.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::time::Instant;
+
+use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue, LabelValueEncoder};
+use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::metrics::gauge::Gauge;
+use prometheus_client::metrics::histogram::Histogram;
+use prometheus_client::registry::Registry;
+use tower::{Layer, Service};
+
+use super::buckets;
+
+/// Outcome of one RPC.
+///
+/// A closed enum, deliberately: `method` is already a 22-value label, so an
+/// open-ended outcome would multiply an unbounded domain into it. Anything that
+/// does not map onto a named variant becomes [`RpcOutcome::Error`].
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub enum RpcOutcome {
+    /// The handler completed and reported no failure.
+    Ok,
+    /// The caller's request was malformed.
+    InvalidArgument,
+    /// The requested model, source, or version does not exist. Distinct from
+    /// [`RpcOutcome::BackendError`] because it is a normal, expected answer.
+    NotFound,
+    /// A backend the handler depends on failed. This is the variant a
+    /// status-code-derived metric cannot see, and the reason this module exists.
+    ///
+    /// Wider than the metadata store: every `Code::Unavailable` maps here, which
+    /// includes the auth layer's `Status::unavailable` when the Kubernetes
+    /// TokenReview API is down. Because this layer sits outside `AuthLayer`, that
+    /// case appears on every method at once and can read like a total
+    /// metadata-store outage. `mx_backend_ops_total` disambiguates: a real store
+    /// outage lights that family up too.
+    BackendError,
+    /// Rejected by the ServiceAccount auth layer.
+    Unauthenticated,
+    /// The caller went away before the handler finished: client disconnect,
+    /// `RST_STREAM`, or a deadline. Recorded from a drop guard, never from a
+    /// response, because in this case there is no response.
+    Cancelled,
+    /// Any other failure.
+    Error,
+}
+
+impl RpcOutcome {
+    /// Wire form. `snake_case`, matching every other label value in the schema.
+    const fn as_str(self) -> &'static str {
+        match self {
+            Self::Ok => "ok",
+            Self::InvalidArgument => "invalid_argument",
+            Self::NotFound => "not_found",
+            Self::BackendError => "backend_error",
+            Self::Unauthenticated => "unauthenticated",
+            Self::Cancelled => "cancelled",
+            Self::Error => "error",
+        }
+    }
+
+    /// Map a gRPC status code onto the closed domain.
+    fn from_code(code: tonic::Code) -> Self {
+        match code {
+            tonic::Code::Ok => Self::Ok,
+            tonic::Code::InvalidArgument => Self::InvalidArgument,
+            tonic::Code::NotFound => Self::NotFound,
+            tonic::Code::Unauthenticated | tonic::Code::PermissionDenied => Self::Unauthenticated,
+            tonic::Code::Unavailable => Self::BackendError,
+            _ => Self::Error,
+        }
+    }
+}
+
+// Hand-written rather than derived: the derive encodes `stringify!(Variant)`,
+// which would emit `BackendError`. Renaming the variants to match the wire form
+// would trip `non_camel_case_types` under `-D warnings`, so the mapping lives in
+// `as_str` instead.
+impl EncodeLabelValue for RpcOutcome {
+    fn encode(&self, encoder: &mut LabelValueEncoder<'_>) -> Result<(), std::fmt::Error> {
+        EncodeLabelValue::encode(&self.as_str(), encoder)
+    }
+}
+
+/// Label set for the per-RPC families.
+#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct RpcLabels {
+    /// `Service/Method`, drawn from a closed set by [`method_label`].
+    pub method: &'static str,
+    /// Resolved outcome; see the module docs for the resolution order.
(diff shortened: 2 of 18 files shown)


modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,433 @@
...
+//!    `Err(Status)` path puts itself there — `Status::into_http` does
+//!    `response.extensions_mut().insert(self)` — so handlers that fail honestly
+//!    need no edit at all, and neither do auth rejections or tonic's own
+//!    `unimplemented` fallback for an unrouted path.
+//! 3. Otherwise success. A successful unary response carries `grpc-status` in
+//!    the HTTP/2 trailers, not the head, so its absence here is the success
+//!    signal.
+//!
+//! Step 2 is why instrumenting 19 RPCs costs a handful of one-line tags rather
+//! than 19 rewrites: only handlers that lie about their outcome need touching.
```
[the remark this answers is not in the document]
yixinh-nv: Fixed. The layer covers 21 routed paths -- 19 proto RPCs plus the two `grpc.health.v1` methods -- so the module doc now says 21, matching the label domain and METRICS.md.

modelexpress_server/src/p2p/service.rs
```diff
@@ -260,36 +304,45 @@ impl P2pService for P2pServiceImpl {
...
-                }))
+                tagged(
+                    GetMetadataResponse {
+                        found,
+                        worker,
+                        mx_source_id: req.mx_source_id,
+                        worker_id: req.worker_id,
+                        identity,
+                    },
+                    RpcOutcome::Ok,
```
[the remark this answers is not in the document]
yixinh-nv: Fixed. The outcome now derives from `found`, so a record with no workers reports `not_found` like the `Ok(None)` branch it is indistinguishable from on the wire. Asserted in `test_get_metadata_record_with_empty_workers`.

modelexpress_server/src/p2p/state.rs
```diff
@@ -34,15 +37,44 @@ impl P2pStateManager {
...
+    /// instead would also instrument the direct-construction integration tests.
+    async fn build_backend(
+        &self,
+        config: BackendConfig,
+    ) -> MetadataResult<Arc<dyn MetadataBackend>> {
+        let backend = create_backend(config).await?;
+        Ok(match &self.metrics {
+            Some(metrics) => InstrumentedMetadataBackend::wrap(backend, metrics.clone()),
+            None => backend,
+        })
```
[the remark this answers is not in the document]
yixinh-nv: Fixed, and the cause is narrower than a factory restructure: `create_backend` calls `connect()` on the concrete backend before returning it, so by the time there is anything to wrap the connection has already happened and the decorator's `connect` forward is never reached. The factory call is now timed as `op="connect"` in both state managers instead, which needs no change to `create_backend` -- the direct-construction integration tests call it too. Worth noting it is not startup-only: `get_backend` reconnects lazily, so a flapping backend now shows as repeated `connect` failures.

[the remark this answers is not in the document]
yixinh-nv: Done -- the Refit factory future is now timed with `BackendMetrics::time(Store::Refit, "connect", ...)` inside the timeout, so all three managers are consistent.

One caveat worth recording rather than leaving implicit: a connect that exceeds the 10s startup timeout records nothing in any of the three, because `tokio::time::timeout` drops the inner future before the recording runs. That path is loud rather than silent -- the server logs the timeout and refuses to start -- so it does not need a metric to be noticed. Noted in a comment at the call site.

ajcasagrande (commented):
## Review — verified against a local build

I read the full diff, then checked the load-bearing claims against the vendored tonic 0.13.1 source rather than taking them on trust. My sandbox had no `libssl-dev`, so I sideloaded it into a private sysroot and built for real. The testing claims in the description all reproduce:

| Command | Result |
| --- | --- |
| `cargo check --workspace --all-targets --all-features` | clean |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | clean |
| `cargo test -p modelexpress-server --features integration-tests` | 280 unit (2 ignored), 5 `in_process_server`, 1 `auth_e2e`, 0 failures |

**Claims that hold**, checked in the tonic source rather than assumed:

- **Extension propagation is real.** `Response::into_http` does `*res.extensions_mut() = self.extensions` (`response.rs:82`), reached from `Grpc::map_response` (`server/grpc.rs:431`). The handler-set `RpcOutcome` does arrive on the `http::Response` the layer sees.
- **`Err(Status)` self-reports.** `Status::into_http` does `extensions_mut().insert(self)` (`status.rs:578`), reached via the `t!` macro at `server/grpc.rs:20`. Auth denials take the same path (`auth/layer.rs:86`). So "the honest handlers need no edit" checks out.
- **The layer wraps the router once.** `service_builder.service(svc)` at `transport/server/mod.rs:660` applies the stack around `Routes`, so unrouted paths and both health services are covered, and `method_label` is matching the right URI.
- **The streaming exclusion list is complete.** The only `stream` RPCs across all five protos are the two named.
- **The forwarding trap is handled.** `list_workers_filtered` is the only default-bodied method across the three traits, and all 30 methods are forwarded explicitly under distinct op names.

The core judgement call is the one most implementations get wrong, and this one gets it right: a status-code-derived metric would have been the obvious build and a liability.

Findings are inline. One substantive (the dropped-future gap), two smaller correctness gaps, four nits. Nothing blocks; I'd want an answer on the first one before merge, even if the answer is "step 5".

*Edited after running these findings past an adversarial reviewer, which was worth doing.* Two changed: the dropped-future fix is now a `cancelled` counter rather than a `Drop` guard writing partial durations into the histogram — the latter would skew the distribution worse than the gap it closes — and the `MAX_MESSAGE_SIZE` finding is now traced through the tonic source and explicitly downgraded to plausible-but-undemonstrated. The in-band-failure finding survived the challenge unchanged.

<sub>— PR Reviewer · 3 MB of libssl-dev, smuggled in, to catch one dropped future.</sub>

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,475 @@
...
+            let started = Instant::now();
+            let result = inner.call(request).await;
+            if record && let Ok(response) = &result {
+                let labels = RpcLabels {
+                    method,
+                    outcome: resolve_outcome(response),
+                };
+                // `Duration::as_secs_f64` is float division, which
+                // `arithmetic_side_effects` permits; `Instant - Instant` is not.
+                metrics.record(&labels, started.elapsed().as_secs_f64(), is_p2p);
```
ajcasagrande: **Cancelled futures truncate the latency tail and mask partial hangs — confirmed, not inferred.**

Recording happens strictly after `inner.call(request).await` resolves. When hyper drops the handler future (client disconnect, RST_STREAM, deadline), this block never runs. I verified it with a throwaway probe that dropped the future mid-flight and asserted the counter stayed at zero.

**Impact — two distinct effects, the second is the one that bites:**

1. *Partial hangs stay green.* If one store wedges while the rest serve, overall throughput stays visibly non-zero. The wedged slice records nothing, so the graph looks healthy while a chunk of traffic is failing.
2. *Survivorship bias in the histogram.* Every timed-out request is a sample that never lands, so `mx_grpc_request_seconds` p99 looks artificially healthy **because** the slowest requests are the ones excluded. No throughput signal repairs this — it is the metric you'd consult, reporting the opposite of reality.

**Suggested fix.** A `Drop` guard, but *not* one that writes partial durations into the histogram — that manufactures synthetic short-latency samples and skews the distribution worse than the gap it closes. Increment the counter only, under a distinct `outcome="cancelled"`, and leave the histogram untouched. An in-flight gauge alongside it would give immediate visibility into a wedged backend, which is the signal actually missing here.

(Same shape applies to `metrics/backend.rs:147`.)

modelexpress_server/src/metrics/backend.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// The recording happens strictly after the await. That ordering is
+    /// load-bearing: `Family::get_or_create` returns a `!Send` guard, so holding
+    /// one across an await would make the caller's future non-`Send` and break
+    /// the `#[async_trait]` bound with an error pointing at lock internals.
+    pub async fn time<T, E, F>(&self, store: Store, op: &'static str, future: F) -> Result<T, E>
+    where
+        F: Future<Output = Result<T, E>>,
+    {
+        let started = Instant::now();
+        let outcome = future.await;
```
ajcasagrande, line 213: The backend half of the cancellation gap (see my note on `metrics/grpc.rs:358`). `let outcome = future.await;` then record — so a dropped future records neither the counter nor the histogram sample.

This is the more consequential of the two: a hung store is exactly what `mx_backend_op_seconds` exists to reveal, and a hang is precisely the case that produces no sample. The tail of this histogram is conditioned on the operation having completed.

Same suggested fix: a `Drop` guard that increments a distinct `outcome="cancelled"` counter, with nothing written to the histogram.

modelexpress_server/src/p2p/service.rs
```diff
@@ -18,6 +19,25 @@ use std::sync::Arc;
...
+/// That is deliberate and stays as it is -- clients depend on it -- but it means
+/// the status code reads `Ok` straight through a backend outage, so nothing
+/// outside the handler can tell an outage from an empty-but-healthy answer.
+///
+/// This publishes the real outcome into the response extensions, where
+/// [`crate::metrics::grpc`] reads it. The tag is metrics-only: it does not touch
+/// the response body or the status code, and losing it degrades the metric to
+/// `outcome="ok"` rather than breaking the call.
+#[allow(clippy::result_large_err)] // Returns the handlers' own tonic::Status result type.
+fn tagged<T>(body: T, outcome: RpcOutcome) -> Result<Response<T>, Status> {
```
ajcasagrande, line 35: This helper is the right pattern, and it's applied consistently here — every in-band failure in this file is tagged. That consistency is what makes one omission elsewhere stand out.

`modelexpress_server/src/services.rs:209` returns:

```rust
Ok(Response::new(ApiResponse {
    success: false,
    data: None,
    error: Some(format!("Unknown action: {}", api_request.action)),
}))
```

No `RpcOutcome` extension, so `resolve_outcome` falls through to `Ok` and the RPC records `outcome="ok"`. `ApiService/SendRequest` is a first-class arm in `method_label` (`metrics/grpc.rs:433`), so this lands under its own method label, not `other`.

To be clear about why this is a bug rather than a design choice: one could argue domain-level `success: false` shouldn't pollute transport-level telemetry — the gRPC call did succeed. But this PR has already made the opposite decision. `p2p/service.rs:92` returns `Ok(Response::new(PublishMetadataResponse { success: false, .. }))` and deliberately tags it `RpcOutcome::InvalidArgument`, and sixteen other `tagged()` sites do the same. `services.rs:209` is the one that was missed.

(Commented here because `services.rs` isn't in this diff.)

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,475 @@
...
+/// `ModelStatus::ERROR` as an item. `Health/Watch` is long-lived by design. For
+/// all three, this layer's future resolves at stream setup, so it would record
+/// `outcome="ok"` and a sub-millisecond duration for work that can run for forty
+/// minutes and fail.
+///
+/// They are therefore not recorded at all. A confidently wrong number is worse
+/// than an absent one, and this module exists to remove exactly that class of
+/// lie. Covering them means timing the response body and reading the trailer,
+/// which is a separate change.
+fn records_at_head(method: &str) -> bool {
```
ajcasagrande: This reasoning is right, and there's a third case in the same family worth naming: **a unary response that exceeds `max_encoding_message_size` records as `outcome="ok"`.**

Mechanism, from tonic 0.13.1 rather than from inference:

- `server/grpc.rs::unary()` awaits the handler, then `.map(|r| r.map(|m| tokio_stream::once(Ok(m))))` — the message is wrapped, not encoded.
- `map_response()` is **not** async: it wraps the body in `EncodeBody::new_server(..)` and returns `http::Response` immediately. That is where the future this layer awaits resolves.
- The size check lives in `finish_encoding` (`codec/encode.rs:183-190`), reached from `encode_item` inside `poll_next` (`encode.rs:104-114`) — i.e. while hyper drains the body, after this layer has already resolved, read the extensions, and recorded `ok`.

The failure then surfaces in the HTTP/2 trailers. The client sees an error; the server records a success.

Scope, stated honestly: for unary the body is `tokio_stream::once(Ok(m))`, which can never yield `Err`, so the `Poll::Ready(Some(Err(status)))` arm at `encode.rs:121` is unreachable — and streaming methods aren't recorded here anyway. That leaves exactly one trigger: a unary response over the 100 MiB set at `server.rs:357`. `p2p.proto:607` documents `WorkerMetadata` as "MB-scale", which is two orders of magnitude short, so I'd call this plausible-but-undemonstrated rather than a live bug — a heavily sharded model or a large `nixl_metadata` payload is the shape that would get there. Worth a comment at minimum so the next person doesn't have to re-derive the trailer boundary.

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    fn call(&mut self, request: http::Request<ReqBody>) -> Self::Future {
+        let path = request.uri().path();
+        let method = method_label(path);
+        let is_p2p = path.starts_with(P2P_PREFIX) && method != "other";
+        // See `records_at_head`: for streaming methods this future resolves before
+        // the work does, so anything recorded here would describe stream setup
+        // while claiming to describe the call.
+        let record = !records_at_head(method);
```
ajcasagrande: Nit: `records_at_head()` returns `true` for exactly the methods where the layer *declines* to record, and the only call site immediately negates it: `let record = !records_at_head(method)`.

The name describes the future's behaviour ("the response head is produced before the work happens"), but reads as a statement about the metrics layer ("this method is recorded at the head"), which is its inverse. `resolves_at_head()` would say the true thing and let the call site read straight.

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,475 @@
...
+            "request_seconds",
+            "gRPC handler latency by method and resolved outcome",
+            latency.clone(),
+        );
+
+        // A separate family rather than a filter over the one above: this is the
+        // surface P2P alerting keys on, and it should not shift when the generic
+        // family's label set changes.
+        let p2p = registry.sub_registry_with_prefix("p2p");
+        p2p.register(
```
ajcasagrande: Nit: `mx_p2p_rpc_total` is exactly a row-subset of `mx_grpc_requests_total` — the same events, selectable by method prefix.

The comment justifies it as insulating P2P alerts from label-set changes in the generic family, but both families are keyed by the *same* `RpcLabels` struct (lines 131 and 133). A label change hits both simultaneously, so the insulation isn't there. If the intent is a stable alerting surface, a recording rule over the generic family gives that without doubling the write path and the cardinality.

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,475 @@
...
+
+        let response = tonic::Status::unavailable("redis down").into_http::<()>();
+        assert_eq!(resolve_outcome(&response), RpcOutcome::BackendError);
+    }
+
+    /// The two `stream` RPCs in model.proto, plus the long-lived health watch,
+    /// must be excluded. If a new streaming RPC is added to a proto without being
+    /// listed here, it will silently report `ok` with a stream-setup duration.
+    #[test]
+    fn streaming_methods_are_not_recorded() {
```
ajcasagrande, line 689: This test is a good guard, and its doc comment slightly oversells what's guarded — worth tightening, because the gap is real.

`method_label()` falls through to `"other"`, and `records_at_head("other")` returns `false`, so `record` is `true`. A streaming RPC added to a proto without a `method_label` arm is therefore not merely mislabelled — it is treated as unary and recorded with a bogus `ok` and a stream-setup duration, silently polluting the latency histogram. The doc comment says it "will silently report `ok` with a stream-setup duration", which is right; the test below it doesn't cover that path.

Cheapest guard: make the fallback explicit about streaming-ness rather than inferring it from a closed list — or assert in a test that every RPC in the protos has a `method_label` arm, so a new one fails the build instead of quietly landing in `other`.

modelexpress_server/src/server.rs
```diff
@@ -162,14 +168,26 @@ pub async fn run_server(
...
+
+    let metrics_listener = MetricsListener::spawn(metrics_addr, Arc::clone(&registry));
     if metrics_listener.is_none() {
         info!("Metrics endpoint is disabled");
     }
 
     // Initialize the model registry manager (Redis or Kubernetes CRDs). Shares the
     // injected backend with the P2P state manager below.
-    let registry = Arc::new(RegistryManager::with_config(backend.clone()));
+    let registry = Arc::new(
```
ajcasagrande, line 188: Nit: this shadows the `Arc<prometheus_client::registry::Registry>` bound nine lines up with a `RegistryManager`. Two unrelated things, one name, in the same visual block. `metrics_registry` for the first would cost nothing.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Added a drop guard recording `outcome="cancelled"` on the counter only, histogram untouched as you suggested.

modelexpress_server/src/metrics/backend.rs
yixinh-nv: Same guard added around `BackendMetrics::time`, counter only, with a test that drops a polled future mid-flight.

modelexpress_server/src/p2p/service.rs
yixinh-nv: Good catch, tagged -- `SendRequest` now reports `invalid_argument` for an unknown action instead of `ok`.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Documented the trailer boundary on `resolve_outcome` rather than fixed.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Renamed to `resolves_at_head` so the call site reads straight.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Removed the family -- you are right that both are keyed by the same `RpcLabels`, so the insulation I claimed never existed.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Replaced the guard with a test that parses the protos and fails the build if any routed RPC lacks a `method_label` arm.

modelexpress_server/src/server.rs
yixinh-nv: Renamed to `metrics_registry`.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Added the in-flight gauge too -- `mx_grpc_requests_in_flight{method}` and `mx_backend_ops_in_flight{store,op}`, decremented from the same drop guard so they cannot drift.

modelexpress_server/src/services.rs
```diff
@@ -182,6 +183,17 @@ impl HealthService for HealthServiceImpl {
 #[derive(Debug, Default)]
 pub struct ApiServiceImpl;
 
+/// Return `body` with the handler's own verdict attached.
+///
+/// See [`crate::metrics::grpc`]. Mirrors the helper in `p2p/service.rs`; kept
+/// local to each module so a handler's outcome is stated next to the handler.
+#[allow(clippy::result_large_err)] // Returns the handlers' own tonic::Status result type.
+fn tagged<T>(body: T, outcome: RpcOutcome) -> Result<Response<T>, Status> {
```
ajcasagrande, line 192: **[LOW] `ModelService/DeleteModel` is the one in-band failure this helper never gets applied to.**

You added `tagged` here and used it for every in-band failure in `P2pService` and for the unknown-action path in `ApiService`. `ModelServiceImpl::delete_model` is the one that was missed.

Chain (all verified at `59f24cd`):

- `services.rs:663-692` — `delete_model` returns `DeleteModelResponse { success: true, .. }` unconditionally; after the provider/name validation there is no error exit.
- `services.rs:848-859` — `delete_model_entries` swallows a failed `get_models_by_last_used`: `error!(...)` then `return 0`.
- `services.rs:833-836` — `delete_status` swallows the per-record `registry.delete_model` failure the same way.
- `metrics/grpc.rs:246` — the method arm exists, so the RPC *is* recorded; with no `RpcOutcome` extension and an OK head, `resolve_outcome` lands on `ok`.

So during a Redis or Kubernetes registry outage, `modelexpress model clear` gets `success: true, "0 record(s)"` back and the dashboard gets `mx_grpc_requests_total{method="ModelService/DeleteModel",outcome="ok"}`. Reachable through the shipped CLI — `delete_model_on_server` at `modelexpress_client/src/lib.rs:238-248`.

**Scope, checked before raising:** `git diff --stat origin/main...HEAD -- modelexpress_server/src/services.rs` is 31 insertions / 10 deletions, and the hunks are the imports, this `tagged` helper, and `ApiService::send_request`. The PR does **not** touch `delete_model`, and the error-swallowing predates it. I am raising it anyway because the *wrong number* is new: the metric that reports the swallow as a success is what this PR adds, and "an in-band failure is not recorded as a success" is this PR's own stated guarantee and the name of its own integration test.

**Why LOW and not MEDIUM:** the swallow itself is pre-existing and unchanged, and `mx_backend_ops_total{store="registry",result="error"}` still shows the outage — an operator is not blind, just misled on this one series. The fix is one `tagged(..., RpcOutcome::BackendError)` on the helper you already wrote.

**Disclosure:** I originally listed ModelService as clean of untagged in-band failures. The contest skeptic falsified that, and it was the most useful objection of the exchange.

modelexpress_server/src/server.rs
```diff
@@ -228,9 +246,21 @@ pub async fn run_server(
         .set_serving::<P2pServiceServer<P2pServiceImpl>>()
         .await;
 
+    // Timed inside the timeout, matching the P2P and registry managers: the
+    // factory connects the concrete backend before it can be wrapped, so the
+    // decorator's own `connect` forward is never reached.
+    //
+    // A connect that exceeds the timeout records nothing, here and in the other
```
ajcasagrande: **[NIT] This comment is contradicted by the guard the same push added.**

> A connect that exceeds the timeout records nothing, here and in the other two managers -- `tokio::time::timeout` drops the inner future, so the recording never runs.

It does record. `BackendMetrics::time` installs `InFlightGuard` **before** the await (`metrics/backend.rs:211`), clears `pending` only after the inner future resolves, and `Drop` calls `record_cancelled` while `pending` is still set (`metrics/backend.rs:255-262`). A timed-out connect therefore emits `mx_backend_ops_total{store=...,op="connect",result="cancelled"}`.

Evidence — your own test drives exactly this shape (`tokio::time::timeout` over `metrics.time(...)`) and asserts the `cancelled` series is present:

```
$ cargo test -p modelexpress-server --lib a_dropped_operation_is_recorded_as_cancelled
test metrics::backend::tests::a_dropped_operation_is_recorded_as_cancelled ... ok
```

Base check: `git show bd0dc9a:modelexpress_server/src/metrics/backend.rs` has no `InFlightGuard` — the guard is new at HEAD, so this is a comment that was true when written and went stale within the same PR, not a pre-existing wart.

**Why NIT:** comment-only, zero runtime effect. It understates a good property; worst case someone later adds redundant instrumentation for a gap that isn't there. (I had this at LOW; the contest correctly argued a comment that understates working instrumentation cannot put a wrong number on a dashboard.)

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,683 @@
...
+            checked, 19,
+            "expected 19 routed server RPCs, parsed {checked}"
+        );
+    }
+
+    /// The two `stream` RPCs in model.proto, plus the long-lived health watch,
+    /// must be excluded. If a new streaming RPC is added to a proto without being
+    /// listed here, it will silently report `ok` with a stream-setup duration.
+    #[test]
+    fn streaming_methods_are_not_recorded() {
```
ajcasagrande, line 689: **[NIT] Follow-up to my earlier comment here — the arm test closed half the gap, this is the other half.**

You added `every_routed_proto_rpc_has_a_method_label_arm` (line 630) in response to the last round, and it does force a `method_label` arm for every routed proto RPC, with `checked == 19` guarding the parser itself. That closes "streaming RPC added, no arm, lands in `other`". Good fix.

What it does not close: arm added, `resolves_at_head` forgotten. This test asserts three hardcoded names, so add `rpc Foo(Req) returns (stream Resp)` to a served proto, add the `method_label` arm (you must — the arm test fails otherwise), bump 19 → 20, and the suite is green while the new stream is recorded with `outcome="ok"` and a stream-setup duration. Your own doc comment three lines above says exactly this: *"If a new streaming RPC is added to a proto without being listed here, it will silently report `ok`."*

The fix is cheap because the parser already exists: `routed_paths` reads the `rpc` lines and streaming is visible as `returns (stream ` on the same line (`model.proto:11,14`). Asserting `resolves_at_head(method_label(&path)) == line_is_streaming` over the same loop closes it, and drops the hardcoded list.

**Why NIT and not LOW:** no reachable broken state exists today. `grep -rn "returns (stream" modelexpress_common/proto/` returns exactly `model.proto:11` and `model.proto:14`, and `resolves_at_head` (line 299) lists both plus `Health/Watch` — the exclusion set is complete for every RPC this binary serves. No CLI flag, env var, config file or request payload can reach the bad state; it needs a future proto edit. I had this at LOW and the contest was right to knock it down.

Related, same shape, also NIT: a brand-new *service* needs edits to `SERVED_SERVICES` (588), `PROTOS` (631), `method_label` and `resolves_at_head`, and nothing fires if you miss them — its RPCs land in `method="other"`.

modelexpress_server/src/metrics/grpc.rs
```diff
@@ -0,0 +1,683 @@
...
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub enum RpcOutcome {
+    /// The handler completed and reported no failure.
+    Ok,
+    /// The caller's request was malformed.
+    InvalidArgument,
+    /// The requested model, source, or version does not exist. Distinct from
+    /// [`RpcOutcome::BackendError`] because it is a normal, expected answer.
+    NotFound,
+    /// The metadata backend failed. This is the variant a status-code-derived
```
ajcasagrande: **[NIT] `BackendError`'s doc is narrower than what actually lands in it.**

The doc says "The metadata backend failed", and `docs/METRICS.md` offers `sum by (method) (rate(mx_grpc_requests_total{outcome="backend_error"}[5m]))` as the backend-outage query. But `from_code` maps *every* `Code::Unavailable` (line 97), and `AuthLayer` returns `Status::unavailable("authentication backend unavailable")` when the Kubernetes TokenReview API is down (`auth.rs:44`, returned at `auth.rs:136`, converted at `auth/layer.rs:84`).

So a kube API outage shows up as `backend_error` on methods that never touched the metadata store — and because the metrics layer sits outside `AuthLayer`, it shows up on *all* of them at once, which reads like a total metadata-store outage.

**Why NIT:** it is still a genuine backend failure, not a false positive, and `mx_backend_ops_total` disambiguates immediately — if the store were really down, that family would light up too. One line of doc ("or the auth backend"), or a distinct outcome variant if you want the query to mean what it says.

ajcasagrande (commented):
**Nothing here is blocking.** One LOW and three NITs; all four are safe to land as follow-ups if you prefer.

I reviewed at `59f24cd` and ran things rather than reading only. Then I put the findings through an adversarial pass with a cross-family skeptic whose job was to kill weak ones. That pass earned its keep: two findings I had graded LOW were correctly knocked down to NIT for being unreachable/comment-only, and one finding below (`DeleteModel`) is something I had actively **cleared** as clean and was wrong about. Severities below are post-contest, not my first instinct.

### Verified correct — you do not have to defend these

- **All 30 backend-trait forwards** across the three `instrumented.rs` decorators call the identically-named inner method with identically-ordered arguments (scripted check over p2p/registry/refit). `list_workers_filtered` is explicitly overridden rather than left to the trait default body — the exact trap this pattern sets — and there is a mock test asserting `expect_list_workers().times(0)`.
- **Extension propagation is real**, not assumed. In vendored tonic 0.13.1: `Response::into_http` does `*res.extensions_mut() = self.extensions;` (`response.rs:82`) and `Status::into_http` does `response.extensions_mut().insert(self)` (`status.rs:578`). The `unimplemented` fallback also goes through `into_http` (`service/router.rs:139`), so an unknown path resolves to `method="other", outcome="error"` rather than a false `ok`.
- **Layer placement is right.** `Server::layer` composes into the `service_builder` applied to `Routes` (`transport/server/mod.rs:660`) — outside routing and outside the per-service `AuthLayer`. Auth denials return through `into_status().into_http()` (`auth/layer.rs:84`), so `outcome="unauthenticated"` genuinely gets recorded instead of being swallowed before the layer sees it.
- **The `poll_ready`/`call` clone-swap** in `GrpcMetricsService` matches the tower contract and the existing `AuthService` precedent in this repo. The `Err(_)` arm from the inner service is dead rather than a miscount: tonic's `Routes` has `Error = Infallible`.
- **Refit handlers fail honestly** with `Err(Status)`, and `refit/backend/redis.rs:96` maps io/cluster errors to `Unavailable`, so they grade `backend_error` and not the catch-all `error`.
- **Tests pass:** `cargo test -p modelexpress-server --features integration-tests` → 283 lib + 5 integration. The default-feature build also passes, so the new tests do not require the feature.
- I could not run clippy to completion locally — `modelexpress-common`, untouched by this PR, fails `result_large_err` under my rustc 1.98. Pre-existing and not counted against this PR.

### Description-only, no inline anchor

The family table in the PR description lists **`mx_p2p_rpc_total`**. No such family exists in the code — `grep -rn "mx_p2p_rpc" modelexpress_server/` is empty, and the P2P RPCs are covered by `mx_grpc_requests_total{method="P2pService/..."}`. Worth fixing in the description so nobody writes a dashboard query against it.

Also in the description: *"Streaming exclusion — asserted, so a streaming RPC added later without being listed fails a test rather than silently reporting `ok`."* That is stronger than what the test does; see the `grpc.rs` NIT below. Your in-code doc comment on `resolves_at_head` states the limitation honestly, so it is the description that overshoots.

---

Nice work overall. The in-band-failure tagging via response extensions is the right call — a status-code-derived metric genuinely cannot see these, and the `an_in_band_failure_is_not_recorded_as_a_success` integration test driving a real client against a real server is the right way to prove it.

modelexpress_server/src/refit/backend/instrumented.rs
nv-hwoo, line 1: Isn't the metrics for P2P services?

modelexpress_server/src/services.rs
yixinh-nv: Tagged -- `delete_model_entries` now reports whether the listing degraded, so a registry outage records `backend_error` instead of `ok`.

modelexpress_server/src/server.rs
yixinh-nv: Corrected -- the comment predated the guard I added later in the same PR, and a timed-out connect does land as `result="cancelled"`.

nv-hwoo (commented):
Forgot to click submit review 😓 Overall looks good. One comment left.

yixinh-nv:
Fixed both description-only items: dropped the stale `mx_p2p_rpc_total` row (and added the two in-flight gauges), and rewrote the streaming-exclusion bullet to state what the test actually guards plus the gap it leaves.

modelexpress_server/src/refit/backend/instrumented.rs
yixinh-nv: `mx_backend_ops_total` is server-wide rather than P2P-only -- it covers all three storage subsystems via the `store` label (`p2p`, `registry`, `refit`), and the PR narrative just leans on P2P because that is where the in-band-failure problem is sharpest.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Closed the other half -- the proto parser now reads `returns (stream ` off the same line and asserts `resolves_at_head(method) == is_streaming`, so an arm added without the exclusion fails the test; mutation-verified by dropping `EnsureModelDownloaded` from the set.

modelexpress_server/src/metrics/grpc.rs
yixinh-nv: Widened both the `BackendError` doc and the METRICS.md query note to say every `Unavailable` lands here, including the auth backend, and to confirm against `mx_backend_ops_total`.

nv-hwoo (approved):

PR #654, opened 2026-08-18
fix(sglang): adopt container-held tensors, and survive SGLang's object graph

AndyDai-nv:
This is the code that was run on hardware and got Kimi-K3 transferring
correctly over SGLang. Both files match the build that produced the evidence
below byte-for-byte (`adapter.py` sha256 `113b3791f927…`, `tensor_utils.py`
`1d47358946d1…`).

## Problem

SGLang's adapter never adopted hidden tensors. `capture_tensor_attrs` only
promotes bare `Tensor` attributes, so weight-derived tensors held inside
containers or on non-`nn.Module` objects stay out of the RDMA manifest:
SGLang's MXFP4 FusedMoE `quant_method` holds swizzled weights after deleting
the original params, and Kimi-K3 keeps `_attn_res_cw_cache` in a dict and
`_k3_fused_decode_args` in a tuple.

Those arrive at dummy values on an RDMA target. Every manifest check passes,
and the model then produces garbage.

Simply calling `adopt_hidden_tensors` is not enough: the traversal it relies on
cannot survive SGLang's object graph. Three shapes took the whole registration
down instead of being skipped, all reached through the same probe,
`hasattr(obj, "__dict__")` — on an object without a real instance dict that
lookup runs `__getattr__`, which is where lazy-import machinery lives.

| shape | what happened |
|---|---|
| a lookup mutates the container being walked | `RuntimeError: dictionary changed size during iteration` |
| a proxy raises for an attribute it cannot resolve | `RuntimeError: Could not find VoxtralRealtimeTextModel neither in <module 'transformers.models'>` |
| a module object reached as a plain attribute | walking it triggers the lazy-import machinery above |

## Fix

**`engines/sglang/adapter.py`** — call `adopt_hidden_tensors`, matching the vLLM
adapter. The same tensor objects are registered, so the container entries alias
the transferred memory and the receive path stays "transfer the final state,
derive nothing afterwards".

**`tensor_utils.py`** — make the traversal survive that graph:

- snapshot containers (`list(obj.items())`, `list(vars(obj).items())`,
  `enumerate(list(obj))`) so a lazy attribute cannot invalidate the iterator
- skip `types.ModuleType`: a module is an import registry, not a tensor
  container
- per-attribute `try/except` so one bad attribute costs that attribute, not the
  scan
- probe for the instance dict through `_has_own_dict()`, which treats a raising
  probe as "not a container" instead of letting it escape

The last one is subtle: an earlier version wrapped the per-attribute recursion
but not the `hasattr` probe itself, which sits in the `elif` condition, so a
raising proxy still aborted everything. The regression tests caught that.

## Verification on hardware

Kimi-K3, TP16 across 4+4 GB200 nodes on GCP, SGLang, weights received over
RDMA, source loads from disk and publishes:

| | manifest entries | target output |
|---|---|---|
| without adoption | 3042 | garbage |
| with this change | 3417 (+375) | SHA1-identical to the source: `881189c3e5e2`, `0f9720e40314`, `d4313e09c000` |

Without the `tensor_utils.py` fixes, enabling adoption on that stack does not
produce garbage — it crashes during registration with the two errors above.

## Tests

Three regression tests, one per shape, reproducing the two production error
strings. Each **fails on `main`**:

```
FAILED test_tensor_utils.py::TestTraversalRobustness::test_raising_proxy_does_not_abort_the_scan
FAILED test_tensor_utils.py::TestTraversalRobustness::test_container_mutated_during_scan
FAILED test_tensor_utils.py::TestTraversalRobustness::test_module_objects_are_not_walked
3 failed, 29 deselected
```

and passes here (32 passed in that file, 68 with `test_pool_registration.py`).

Full client suite: 87 failed / 1230 passed on this branch versus 87 failed /
1227 passed on `main` — same failures (pre-existing and environmental), plus
the three new tests.

A first attempt at these tests passed on unfixed `main`, i.e. proved nothing:
the helpers hung `@property` and `__getattr__` on names the walk never touches
(a property lives on the class, so it is not in `vars(obj)`, and the walk reads
`__dict__` values rather than doing attribute lookups). They were rewritten
around the `hasattr` probe, which is the real trigger.

## Note

Supersedes #589, which enabled adoption without the traversal fixes it needs.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/sglang/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/sglang/adapter.py
@@ -20,6 +20,7 @@ from ...accelerators import accelerator_backend_for
 from ...load_strategy.context import LoadContext, LoadResult
 from ...metadata.client_factory import create_metadata_client
 from ...tensor_utils import (
+    adopt_hidden_tensors,
     capture_tensor_attrs,
     collect_module_tensors,
 )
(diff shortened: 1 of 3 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/654/commits/cd516a44df5d2bfb51362cdb4528d61eeca4866e

zhengluo-nv (approved):

PR #656, opened 2026-08-19
feat(client): support the engine's pinned revision

scydas:
## Summary

Part of #569 — adds pinned-revision support to the server-cache metadata fallback.

The engine's `revision` previously reached the prefetch hook but stopped there. The server returned its default revision, while the engine later looked up the revision it had requested and failed offline with `LocalEntryNotFoundError`.

This change preserves the requested revision through the client pipeline:

- `EnsureModelDownloaded` receives the engine's revision.
- `ListModelFiles` and `StreamModelFiles` receive the immutable commit resolved by the server, preventing a moving branch or tag from mixing commits mid-phase.
- A server that does not confirm a requested revision fails the install rather than silently installing its default.

### Cache layout

| Engine requested | Result |
|---|---|
| no revision | `snapshots/<commit>/` + `refs/main` |
| lowercase 40-hex SHA | `snapshots/<commit>/`, no ref |
| branch, tag, or uppercase SHA | `snapshots/<commit>/` + `refs/<requested revision>` |

This follows `huggingface_hub`'s cache rule: a lowercase full commit SHA resolves directly by snapshot directory; other revision strings resolve through `refs/`.

The ref belongs to the request, not the snapshot. Therefore it is also written when an existing snapshot is reused: a commit previously installed under its own SHA has no ref, but a later branch or uppercase-SHA request resolving to that commit still needs one for offline resolution.

Pins other than `main` leave `refs/main` untouched, so installing an older revision cannot redirect later unpinned resolution. An explicit `revision=main` correctly updates `refs/main`, because that is the ref Hugging Face uses to resolve that branch.

The in-process prefetch record is now keyed by `(repo_id, revision)`, so requests for two revisions of the same model do not reuse each other's snapshot.

`_warn_on_revision_mismatch` is removed: the client can now request the revision instead of only warning that it cannot.

Not in scope: `MX_MODEL_REVISION` remains a P2P source-identity label. It accepts arbitrary strings and does not select or pin a Hugging Face revision; use the engine's own revision setting for that.

## Validation

- `uv run --no-sync pytest tests/test_hf_snapshot_prefetch_patch.py tests/test_model_prefetch.py tests/test_model_client.py tests/test_model_snapshot.py tests/test_server_cache_strategy.py`
  - `198 passed`
- Added coverage for revision forwarding, branch/tag and uppercase-SHA refs, offline resolution, reuse-path ref creation, preservation of `refs/main`, rejection of an unconfirmed or mismatched commit pin, and separate in-process installs for separate revisions.


modelexpress_client/python/modelexpress/model_client.py
```diff
@@ -220,8 +258,17 @@ def install_metadata_snapshot(
...
+                existing = cache.resolve_pinned_snapshot(expected, revision)
+            else:
+                existing = cache.resolve_snapshot(expected, revision)
             if existing is not None:
+                # The ref belongs to the request, not to the snapshot: an
+                # earlier install of this same commit under a different
+                # revision left no ref this one can be found by. Reuse skips
+                # publish(), so record it here or the engine's lookup fails
+                # against a directory that is sitting right there.
+                cache.write_revision_ref(revision, requested_revision)
```
nicolasnoble, line 271: on the unpinned path this ends up rewriting a ref that's already correct. resolve_snapshot only returns non-None when read_main_ref() == expected_commit, so by the time we're here main already holds this value, and write_revision_ref -> write_main_ref -> write_ref does a temp file, an fsync, a rename and a directory fsync to land identical bytes.

the pinned path isn't the same - write_revision_ref does skip when requested_revision == commit_hash, and a genuinely new alias is a real write. so this is only about the unpinned reuse case. probably worth an early return when the ref already reads the same hash?

modelexpress_client/python/modelexpress/model_snapshot.py
```diff
@@ -471,20 +483,50 @@ def read_main_ref(self) -> str | None:
...
+        A ref name may contain slashes -- ``refs/pr/1`` is a legal revision --
+        so the parent directories are created here. The name comes from the
+        engine's own configuration, so it is validated against the same rules
+        as a streamed file path before it becomes part of one.
+        """
         commit_hash = safe_commit_hash(commit_hash)
         refs_root = self.repo_root / "refs"
-        _ensure_directory(refs_root, self.cache_root)
-        temp_ref = refs_root / f"{_TEMP_PREFIX}{uuid.uuid4().hex}"
+        ref_path = refs_root / safe_relative_path(ref_name)
```
nicolasnoble: now that refs are published under refs/<revision>, a slash-bearing revision makes this a nested path, and safe_relative_path deliberately allows slashes so refs/pr/1 works. that leaves file-vs-directory collisions unhandled: if refs/foo already exists as a file and something later writes foo/bar, _ensure_directory's mkdir raises FileExistsError, and in the other order os.replace raises "Is a directory" on refs/foo.

both propagate as raw OSError out of install_metadata_snapshot rather than ModelSnapshotError, so a caller catching the library's own error type won't see it. test_rejects_unsafe_ref_names covers ../escape, /abs, "" and a/../b but not this case.

not sure how reachable it is in practice - depends whether a repo can have a branch foo and a branch foo/bar at once, which git does allow. up to you whether it's worth guarding.

modelexpress_client/python/modelexpress/model_prefetch.py
```diff
@@ -25,23 +25,21 @@
...
-COMMIT_HASH_PATTERN = re.compile(r"^[0-9a-fA-F]{40}$")
-
 _REPO_DIR_PREFIX = "models--"
 
 # Reentrant: ensure_metadata holds this across the whole install and calls
 # helpers that take it again.
 _lock = threading.RLock()
 _snapshot_to_repo_id: dict[str, str] = {}
-_installed: set[str] = set()
+_installed: set[tuple[str, str | None]] = set()
```
nicolasnoble: _installed and _revision_snapshots have the same key domain, are added together at 99-101, cleared together at 148-149, and there's no path that touches one without the other. the only read of _installed at line 89 is equivalent to testing membership in _revision_snapshots, so _installed looks droppable.

_snapshot_to_repo_id is keyed differently and serves repo_id_for, so that one earns its keep.

nicolasnoble (commented):
three small things inline, none of them blocking.

modelexpress_client/python/modelexpress/model_client.py
scydas: Fixed in 5c3cc60, at the shared write_ref layer rather than the call site. Whether the ref already contains the requested hash is a property of the write itself, so callers do not need to duplicate that check. I also split the read helper so the comparison uses the same validation semantics as read_main_ref. Tests verify that writing the same hash leaves the ref untouched, while a different hash still updates it.

modelexpress_client/python/modelexpress/model_snapshot.py
scydas: Guarded in 5c3cc60. I also checked reachability: Git does not allow branches foo and foo/bar to coexist within refs/heads, but the cache collision is still reachable across namespaces. For example, a tag foo and a branch foo/bar can coexist upstream but map to refs/foo and refs/foo/bar in the cache. A full ref such as refs/pr/1 can create the same conflict with a branch named refs. The guard checks for the conflicting layout before writing, so unrelated I/O failures such as ENOSPC are not converted into ModelSnapshotError. Both collision orders are covered by tests.

modelexpress_client/python/modelexpress/model_prefetch.py
scydas: Dropped in 5c3cc60. The membership check now uses the equivalent _known_snapshot lookup, while _snapshot_to_repo_id remains because it has a distinct key and is still used by repo_id_for.

nicolasnoble (approved):
all three addressed, nothing further from me.

PR #657, opened 2026-08-19
ci: bump GHA version, cargo matrix and add upstream guards

zdtsw:
# Changes

- bump various GHA versions to the latest one
- fix cargo toolchain matrix:  was hardcoded to `--default-toolchain stable` only; now installs the matrix channel AND sets `RUSTUP_TOOLCHAIN` to override `rust-toolchain.toml` so beta/nightly legs genuinely run those channels
- make failed integration test visible in UI instead of swallowed the error
- add missing restore-keys on integration-test cache
- skip workflow that require upstream(ai-dynamo) org resource: image, runner , secrets etc
- switch CI from pinned `1.24.x` to `go-version-file` sourced from `go.mod`, which moves CI to Go 1.25.0(from current go.mod)

cc @wseaton 

## CI & Build Improvements

* Updated automated build, test, checkout, caching, and coverage tooling.
* Improved Rust toolchain handling, including beta-test resilience and stable-only release builds.
* Aligned Go setup with the project’s configured version.
* Added improved caching for integration tests.
* Restricted release, wheel, copyright, and repository-mirroring jobs to supported upstream environments.
* Improved handling of integration-test failures in CI.


.github/workflows/ci.yml
```diff
@@ -41,22 +41,22 @@ jobs:
...
+      uses: actions/checkout@v7
 
     - name: Check Helm CRD manifest sync
       if: matrix.rust == 'stable'
       run: cmp examples/crds.yaml helm/crds/modelexpress-crds.yaml
 
     - name: Install Rust toolchain
       run: |
-        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${{ matrix.rust }}
```
nicolasnoble: i think this ends up being a no-op. rust-toolchain.toml pins channel = "stable" and an override file wins over rustup's default for every cargo and rustc call inside the checkout, so the beta and nightly matrix legs still run stable.

visible in the beta job's own log on run 32253540476: beta installs, is set default, rustup syncs stable back, and rustc --version reports 1.97.1. so the matrix is testing stable three times and this adds two toolchain downloads that don't get used.

if the intent is to genuinely test beta and nightly, RUSTUP_TOOLCHAIN in the job env would override the file, or cargo +${{ matrix.rust }} at the call sites. happy to be wrong if you've seen it pick up beta somewhere.

.github/workflows/ci.yml
```diff
@@ -209,12 +210,12 @@ jobs:
...
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v7
 
     - name: Set up Go
-      uses: actions/setup-go@v5
+      uses: actions/setup-go@v7
       with:
-        go-version: "1.24.x"
+        go-version-file: modelexpress_client/go/go.mod
```
nicolasnoble, line 273: this moves go from a pinned 1.24.x to whatever go.mod says, and go.mod currently reads go 1.25.0, so CI shifts from the 1.24 series to 1.25 as part of this. probably the right change, but it isn't in the description and it's the kind of thing that's easier to find now than after something breaks.

nicolasnoble (commented):
one thing that i think stops the rust matrix change from doing what it's meant to do, plus a version bump in here that isn't mentioned in the description. inline.

the split wseaton offered would probably help, if only to separate the go bump from the actions work.

.github/workflows/ci.yml
zdtsw: you are right. I missed this in the PR description, updated

.github/workflows/ci.yml
zdtsw: I've added `RUSTUP_TOOLCHAIN: ${{ matrix.rust }}` to the job env so beta/nightly actually run. 
I am  also fine to drop nightly and beta and just keeping stable. 
Let me know your preference

nicolasnoble: that does it - beta reports 1.99.0-beta.1 and nightly 1.100.0-nightly in run 32753326112, so the three legs are three different toolchains now.

i'd keep both, but could you flip beta to can-fail: true while you're in there? it's can-fail: false today, which was harmless while beta was really stable and couldn't fail on its own. now that it genuinely runs beta it can fail on its own, and with can-fail: false that blocks the merge. the early warning is worth having, just not as a blocker.

separate thing you probably want to know: ci is red on DCO, not on anything in the matrix. c57ceb68 is missing a sign-off, the other two commits have one, so it looks like it just got skipped on the quick fix. that's what's holding the merge up right now.

zdtsw: Le me fix both: 
- flip beta to `can-fail: true` 
- amend commit with the missing sign-off

zhengluo-nv:
Hi, could you also address the CodeRabbit comments?

zdtsw:
@coderabbitai review

zdtsw:
> Hi, could you also address the CodeRabbit comments?

I've addressed most of concerns fromo coderabbitai, thanks!

zhengluo-nv:
/ok to test [408de1f](https://github.com/ai-dynamo/modelexpress/pull/657/commits/408de1fa2f8cef043407236e5edcac4492b162e5)

zhengluo-nv (approved):

PR #658, opened 2026-08-19
fix(client): re-register source when server rejects heartbeats

h-avsha:
## Problem

When the server's metadata store restarts and loses its data (hit in production
when Valkey reached its memory limit), workers keep heartbeating with a cached
`mx_source_id` the server no longer knows. Every `UpdateStatus` is rejected and
logged as an error on both sides, and since `PublisherThread` ignores the result,
the worker never re-publishes: the source stays orphaned until the process
restarts. Related: #607/#608 reduced the heartbeat write amplification that
contributed to the memory limit being hit; this PR fixes recovery after the reset.

## Fix

`PublisherThread` now checks the `UpdateStatus` result. After 2 rejected
heartbeats in a row it drops its cached `mx_source_id` and re-runs the existing
publish path on the next tick, restoring the registration.

- Only answered rejections (`success=false`) count. Errors reaching the server
  keep the existing retry behavior — re-publishing couldn't succeed either.
- Heartbeat-only threads (no `publish_fn`: `publish_artifact_source`,
  `RefitRendezvous`) can't recover on their own; they now log one warning
  instead of an error per interval. A follow-up can give both a re-publish
  callback — they only lack one because their initial publish happens inline,
  not because re-publishing is unsafe.
- No proto or server changes.

## Testing

7 new tests in `TestHeartbeatReRegistration` (re-publish on threshold, counter
reset, no re-publish on transport errors, warn-once mode, clean stop).

    cd modelexpress_client/python && python -m pytest tests/ -q
    1299 passed, 32 skipped

AI assistance was used (Claude); every line reviewed by the submitter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection recovery when heartbeat updates are rejected.
  * Automatically re-registers the publisher and republishes when needed.
  * Prevented misleading shutdown messages when status updates fail.
  * Added graceful handling when republishing is unavailable or interrupted by shutdown.


zhengluo-nv:
/ok to test [66a18fb](https://github.com/ai-dynamo/modelexpress/pull/658/commits/66a18fb9b278b181110cacc311b19eb94fca662c)

zhengluo-nv (approved):

PR #672, opened 2026-08-22
feat(refit): view-ops slice planning for reshard

tanushriya910:
## What
Extend reshard geometry capture and the slice planner to support **rank-reducing view ops**, so a trainer that publishes a *stacked* source (e.g. MoE experts stacked into one tensor) can be resharded into its per-slice destination weights.

## Changes
- **geometry.py** — add `select` and `split` to `_SUPPORTED_OPS`. `select` (rank-reducing) and `split` (multi-return view, like `chunk`) are how a stacked source is unstacked into per-expert / per-slice tensors. Without them, every such weight was classified unsupported and the receiver failed closed.
- **slice_plan.py** — `_view_to_box_perm` now derives the source box for **rank-reducing** views (select/unbind drop a dim): a dropped source dim is not iterated by any view dim, so it contributes a single fixed index (`[idx, idx+1]`) decoded from the offset. Rank *increase* (unsqueeze) and dim-merging reshapes still raise `UnsupportedReshard`.
- Tests: `test_reshard_refit_geometry.py`, `test_reshard_refit_slice_plan.py`.

## Why
Stacked-expert MoE layouts (and similar stacked sources) need select/split to map each slice onto its own destination box — closing that gap in view-op coverage.

Stacked on the receiver-mapping PR (base `tanushriyas/reshard-native-convert`); retarget to `main` once that merges.

### DIFF
--- a/modelexpress_client/python/modelexpress/refit/reshard/geometry.py
+++ b/modelexpress_client/python/modelexpress/refit/reshard/geometry.py
@@ -64,6 +64,10 @@ __all__ = [
 # ``__torch_dispatch__`` and raises UnsupportedReshard.
 _SUPPORTED_OPS: dict[Callable, str] = {
     torch.Tensor.narrow: "narrow",
+    # select/split unstack a stacked source (e.g. a trainer that publishes its
+    # experts stacked -> per-expert HF weights). select is rank-reducing; split is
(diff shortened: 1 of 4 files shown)


tanushriya910:
cancelling in favor of #680

PR #677, opened 2026-08-24
feat(oci): materialize GBuild artifacts

MihailoMilenkovic:
Add digest-pinned OCI artifacts as a Modelexpress source, including the GBuild full-compile format.

## GBuild artifact contract

Modelexpress accepts `repository@sha256:...` references and requires exactly four layers:

- Manifest V2 JSON;
- Manifest V2 Cap'n Proto;
- `compile-metadata.tar.zst`, containing the preset and optional LPUSim result named by Manifest V2; and
- `payload.tar.zst`, containing runtime files for all partitions.

The consumer validates the artifact type, standard empty config, layer names, media types, and Manifest V2 revision before materialization. It checks the compile metadata archive against the paths named by Manifest V2. Its shared archive extractor rejects path traversal, links, special files, and duplicate outputs.

The OCI manifest owns blob descriptors. Manifest V2 owns the runtime layout, so the format has no second payload config.

## Registry access

Explicit OCI credentials take precedence. GAR can use Application Default Credentials. Other registries use the normal OCI credential environment.

This builds on the generic OCI provider work from #268.

## Live proof

A 14-partition GPT-OSS 20B artifact from development GAR materialized through the packaged CLI in 2m25s:

`us-docker.pkg.dev/nv-hwlpu-20260103060253/compile-artifacts-dev/full-compiles@sha256:10a79b52434cb3797d2d813cf4c0b7736a88ca8578f8486840d24936c8ed80a2`

## Verification

- `cargo fmt --all --check`
- `cargo test -p modelexpress-common providers::oci --no-default-features --features tls-rustls`
- `cargo clippy -p modelexpress-common --all-targets --no-default-features --features tls-rustls -- -D warnings`
- Digest-pinned full-artifact integration test

*🤖 Codex*


AndyDai-nv:
@MihailoMilenkovic hi, is this WIP or we can close this one?

PR #679, opened 2026-08-24
fix(artifacts): skip symlinks when sealing cache artifacts

AndyDai-nv:
## Problem

Artifact packaging raised on the first symlink found under a cache root:

```python
def _reject_symlinked_source_entries(source_root: Path) -> None:
    for path in source_root.rglob("*"):
        if path.is_symlink():
            raise ValueError(f"tarred artifact source contains symlink: {path}")
```

A user running Nemotron-3-Ultra on B200 (vLLM 0.27.1, disagg) hit this and read it as a harmless warning. It is not: `prepare_source()` throws, `PublisherThread` retries every 5s until `MX_ARTIFACT_READY_TIMEOUT_SECS` (default 1800s) and gives up, so **that worker never publishes its FlashInfer artifact at all** — the JIT-compiled modules the artifact exists to share go with it, and every decode worker recompiles them itself. Weight transfer is unaffected (140 GB in 8.84s in their logs); this is only the artifact path.

```
WARNING [publisher.py:278] [Worker 2] Source publish attempt failed (121s elapsed,
timeout=1800s), will retry next tick: tarred artifact source contains symlink:
~/.cache/flashinfer/0.6.16.post3/100a/generated/trtllm_export/fused_moe_trtllm_sm100/
flashinfer/trtllm/batched_gemm/trtllmGen_bmm_export
```

This is not an edge case on Blackwell. FlashInfer's trtllm-gen kernels ship as
prebuilt cubins plus headers stored content-addressed under a hash directory,
while generated `.cu` files include them from a fixed layout, so
`jit/fused_moe.py` bridges the two with a symlink on every JIT module lookup.
Any SM100 worker using trtllm-gen fused MoE hits it.

## Fix

Enumerate archive members explicitly and hand tar that list, instead of letting
tar walk the tree:

```
tar -cf <tar> -C <src> --null --no-recursion -T <member list>
```

Symlinks are left out of the list. They are derived state that engines rebuild:
FlashInfer's `ensure_symlink()` removes whatever sits at the path (including a
real directory, via `shutil.rmtree`) and recreates the link, and it runs before
`build_and_load()`'s cache-hit check, so it executes in every process that
touches the op. A link carried across pods would be deleted and recreated
anyway.

Skipping a symlink costs the link entry alone. Neither `os.walk(followlinks=False)`
nor tar descends through a symlinked directory, so no subtree that tar would
otherwise have archived is lost.

Logging is graded and never fatal: a link resolving inside the cache root logs
at debug (its target is archived under its real path); a link that leaves the
root or dangles logs one warning naming the paths.

### Why an explicit member list rather than `--exclude`

tar reads exclude patterns as globs, so an entry whose name contains glob
metacharacters slips past its own exclusion:

```
$ ln -s /nowhere 'src/gen/tricky[a]link'
$ tar -cf o.tar -C src --exclude='./gen/tricky[a]link' .
$ tar -tf o.tar
./gen/tricky[a]link          # not excluded
```

That symlink would then reach the target and fail `_validate_tar_members`
*after* a full transfer, turning a source-side problem into a target-side one.
`--exclude-from` fixes the metacharacter case but still cannot express a name
containing a newline (`--null` only applies to `-T`). An explicit member list
is an allowlist: nothing that is not listed can enter the archive, whatever it
is named. That keeps the file-or-directory invariant on the source side, where
it can be enforced, and leaves the extraction path untouched — no new
validation code, no new attack surface.

### Other approaches considered

- **`tar -h` (dereference).** The bytes would be deleted by the target's
  `ensure_symlink()`, so it is pure waste. It also breaks on dangling links
  (tar exits 1) and on symlink cycles — GNU tar 1.35 segfaults on
  `a/loop -> ../a` after expanding 5625 levels.
- **Storing symlinks as symlinks.** Requires new linkname validation on the
  extraction side for zero benefit, since the target rebuilds the link itself.

Non-regular files (fifo, socket, device nodes) are skipped the same way. They
serve the same invariant: previously the source accepted them and the target
rejected them after a full transfer.

## Testing

Replaces `test_tarred_p2p_artifact_transfer_rejects_symlink` with five cases:
internal symlink, external symlink, broken symlink, awkward member names
(`config[sm100].inc` and a name containing a newline, alongside symlinks with
the same shapes), and empty-directory preservation.

`test_artifact_transfer.py`, `test_vllm_artifacts.py`, `test_sglang_artifacts.py`,
`test_artifact_health_url.py`: 119 passed.

Also verified end to end against a reproduction of the reported layout (cubin
include tree under site-packages, cache under
`0.6.16.post3/100a/generated/trtllm_export/fused_moe_trtllm_sm100/...`): publish
succeeds with one warning naming the skipped link, and the target receives
`cached_ops/fused_moe_trtllm_sm100/module.so`, the empty `cached_ops/tmp`, and
the link's parent directory.

Pre-existing unrelated failures in `test_vmm_*` (the `modelexpress.vmm._alloc_ext`
C extension is not built in this environment) reproduce identically with these
changes stashed.

## Behavior change

Symlinks under a cache root are no longer an error for any artifact type. The
worst case is a target cache missing one derived entry, which degrades to a
recompile — these are all regenerable JIT caches. The current worst case is the
entire cache failing to publish, every time.

## Summary by CodeRabbit

- **New Features**
  - Artifact packaging now safely excludes symlinks and unsupported entries while preserving regular files and empty directories.
  - Archive creation supports filenames containing brackets or newlines.

- **Bug Fixes**
  - Artifact publication no longer fails when source directories contain symlinks.
  - Broken and external symlinks are skipped with appropriate warnings.

- **Documentation**
  - Documented artifact packaging, extraction safeguards, and symlink handling behavior.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1082,6 +1082,12 @@ graph TD
 7. **Target becomes source**: After receiving weights or installing a cache artifact, publishes own metadata and starts its own heartbeat
 8. **Stale detection**: Server-side reaper marks workers STALE if `updated_at` > 90s old; `ListSources(READY)` also applies this heartbeat freshness check at query time so expired READY records are not returned while waiting for the next reaper pass. GC deletes STALE workers after 1 hour
 
+Tarred cache artifacts carry regular files and directories only. The source
+side enumerates members explicitly and hands tar that list, so nothing else can
+enter the archive regardless of how a cache directory is laid out; the target
+side enforces the same invariant in `_validate_tar_members` before extraction.
+Symlinks are skipped at packaging time (see [`DEPLOYMENT.md`](DEPLOYMENT.md#p2p-metadata-exchange)).
+
 Cache artifact checksums protect transfer integrity but do not authenticate the
 source or attest the contents. TorchInductor, Triton, DeepGEMM, TileLang, CuTe
 DSL, and FlashInfer caches may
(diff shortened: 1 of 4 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/679/commits/dbe8e87fe79fb24a5512da98d2d87f003e6391dd

zhengluo-nv (approved):

PR #681, opened 2026-08-24
feat(metrics): PodMonitors, alert rules and a Grafana dashboard

yixinh-nv:
## Summary

Four PRs have added Prometheus metrics. Nothing consumes them, and on
kube-prometheus-stack nothing even **scrapes** them: the chart's only scrape
mechanism was `prometheus.io/*` pod annotations, which the Operator ignores
entirely. The endpoint is served and never visited — indistinguishable from a
crashed exporter.

This adds the scrape path, alerts and a dashboard. All of it defaults to off.

```mermaid
flowchart TB
  subgraph chart["created by this chart -- all default off"]
    direction LR
    cm["dashboard<br/>ConfigMap"]
    pm["PodMonitor<br/>podMonitor"]
    cpm["PodMonitor<br/>clientPodMonitor"]
    rule["PrometheusRule"]
  end
  sp["ModelExpress server pod<br/>port metrics :9401"]
  cp["inference engine pod -- not ours<br/>port mx-metrics :9402"]
  prom["Prometheus"]
  graf["Grafana"]
  pm -. selects .-> sp
  cpm -. selects .-> cp
  sp -- scrape --> prom
  cp -- scrape --> prom
  rule -- alerts --> prom
  cm -- sidecar --> graf
  prom -- PromQL --> graf
```
**Creating these is not the same as them being used**, and each is gated
differently:

| Object | Picked up by | Gate |
| --- | --- | --- |
| `podMonitor`, `clientPodMonitor` | Prometheus Operator | Labels must match the Prometheus CR's `podMonitorSelector` |
| `PrometheusRule` | Prometheus Operator | Labels must match its `ruleSelector` |
| dashboard `ConfigMap` | Grafana sidecar | Carries the sidecar's label, `grafana_dashboard` by default |

A gate that does not match is not an error: the object is created, `kubectl get`
shows it, nothing logs a complaint, and it is ignored for as long as it exists.

| File | Purpose |
| --- | --- |
| `podmonitor.yaml` | Server scrape target, selected by the named `metrics` port |
| `podmonitor-client.yaml` | Client metrics, which live in the **inference engine's** pod |
| `prometheusrule.yaml` | 13 alerts |
| `dashboards/modelexpress.json` + `dashboard.yaml` | Grafana dashboard via sidecar ConfigMap |

## The one production-code change: counters pre-created at zero

`Family` creates children lazily, so a counter's **first exported sample is 1**.
Prometheus has no earlier point to subtract from, so `increase()` returns 0 and
the first occurrence in each process is invisible:

| Scrape | Lazy counter | `increase()` | Pre-created | `increase()` |
| --- | --- | --- | --- | --- |
| t=0 (pod start) | *no series* | — | `0` | — |
| t=30 (takeover) | `1` | **0** — no prior point | `1` | **1** |
| t=60 | `1` | 0 | `1` | 1 |

`MXDownloadTakeover` therefore missed the **first** takeover after every restart
— and a takeover is a full re-pull of the model. Fixed in
`RegistryMetrics::register`; costs 23 always-present series per pod, scoped to the
three families whose alerts key on rare one-shot events. gRPC and backend would
cost 154 and 93 for ratio alerts that need sustained traffic anyway.

## Three chart decisions worth reviewing

- **Enabling an Operator resource without the CRDs fails the install** rather than
  rendering nothing. A silent skip reproduces the exact failure this PR fixes.
  `skipApiVersionCheck` covers CRDs arriving in the same apply.
- **The client PodMonitor's selector is required, not defaulted.** An empty
  selector matches *every* pod in the namespace, and there is no
  ModelExpress-owned label to fall back on — those pods belong to
  vLLM/TRT-LLM/Dynamo.
- **Alerts only, no recording rules.** Recording rules would make the dashboard
  depend on the rules being installed; enabling one and not the other would give
  empty panels and no error.

## Dashboard

Five rows, outcome-first, modelled on Dynamo's dashboard structure.

| Row | Answers |
| --- | --- |
| **Overview** | Is anything wrong right now? (8 stat tiles) |
| **Downloads** | Is the primary job working, and how fast? |
| **Server internals** | gRPC + storage backend: rate, errors, p99, in flight |
| **P2P clients** | Selection funnel, transfer time, NIXL health |
| **Capacity** | Map growth, evictions, version skew |

Six Overview tiles are coloured by the same condition an alert fires on
(`MXServerMetricsAbsent`, `MXRegistryStatsStale`, `MXDownloadFailureRatio`,
`MXDownloadTakeover`, `MXDownloadLeaseLost`, `MXGrpcErrorRatio`); each tile's
description names its alert. It is **not** a complete alert summary — seven
alerts have no tile — and `METRICS.md` says so, so an all-green Overview is not
read as an all-clear.

## Testing

| Check | Result |
| --- | --- |
| `helm_alert_rules.rs` | Cross-checks every `mx_*` name in rules **and** dashboard against a registry with all 15 server families registered and sampled |
| Python side | Closes the loop for client families, parsed out of the Rust source so the lists cannot drift |
| `promtool test rules` | SUCCESS — first takeover, one lost lease, wedged download, dead store, plus a healthy case that must fire nothing |
| Mutation check | Reverting `MXDownloadLeaseLost` to its pre-fix `rate(...[5m])` form fails; a takeover on a counter *born at 1* correctly does **not** fire |
| `helm lint` | Clean against all five values files |
| Unit tests | 305 server, 4 Helm-source, 1413 Python |

The name cross-check matters because the failure is silent: a query naming a
family that does not exist evaluates to an empty vector, so a rename disarms its
own alert with nothing logged.

**CI does not render this chart** — it never has. The promtool suite is run by
hand today. I have a `helm-verify` job written that lints, renders in both
capability states and runs promtool; left out to keep this PR's scope to the
chart, happy to raise it separately.

## Validated on a real cluster

Deployed the CI-built server image with the Redis metadata backend, a real
`EnsureModelDownloaded`, live gRPC traffic and a client exporter, scraped by
Prometheus with the chart's own alert rules loaded, and rendered the dashboard
through Grafana's image renderer.

| Checked | Result |
| --- | --- |
| Both PodMonitors render and select the right pods | Server by named `metrics` port, client by `mx-metrics` |
| Prometheus scrapes both | 2 targets up; 137 `mx_*` series from the server |
| All 13 alert rules load and evaluate | `health: ok`, no expression errors |
| Real download end to end | `EnsureModelDownloaded` → `DOWNLOADED` in 2m03s |
| Dashboard provisions and renders | All 34 panels, 5 rows, no panel stripped |

Three defects the render caught, each fixed in this PR:

| Symptom on the cluster | Cause | Fix |
| --- | --- | --- |
| "Downloads completed: 0" while the server exported `mx_download_seconds_count{outcome="downloaded"} 1` | Histogram children created lazily — same bug as the counters, one family further on. Also silences `MXDownloadFailureRatio` on a first-download failure (0/0 is no data, not a ratio) | Pre-create the two terminal outcomes |
| "Download failures: No data" on a healthy system, where 0 was meant | `noValue` written into `panel.options`; Grafana reads `fieldConfig.defaults` | Moved on all 10 panels |
| Bottom of the dashboard cut off; a full-width panel showing the number `1` | Layout | Resized, text panel one cell taller |

**And the alerts found a fourth defect, outside this PR.** On the Kubernetes
registry backend the download never completed:
`mx_download_lease_refresh_total{result="lost"}` climbed to 9 on a single-replica
idle server and `MXDownloadLeaseLost` evaluated to **10.26 > 0**. That backend's
lease refresh is rejected by the API server with `Invalid` (422); the server
misreads it as lost ownership and aborts. Tracked separately, untouched here —
but it is the alert documented as "the leading indicator of a duplicate
download", finding one on first contact with a real deployment.

## Not included

- **Per-tier latency attribution.** The dashboard can say a transfer took 90s, not
  where the 90s went. Needs the L0–L4 tiers — the next piece of work.
- **Refit stage timings.** Still no production call sites.
- **Two known-stale `METRICS.md` claims** in sections this PR does not otherwise
  touch: the `_started_total`/`_finished_total` passage, and the client family
  table omitting the optional `source_worker_id` label. Happy to fix in a
  follow-up.


.github/workflows/ci.yml
```diff
@@ -99,6 +99,87 @@ jobs:
     - name: Run doc tests
       run: cargo test --doc
 
+  helm-verify:
+    name: Helm Chart Verify
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
```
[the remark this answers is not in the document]
yixinh-nv: Done — added `permissions: contents: read` and `persist-credentials: false`, matching the artifact-transfer-contract job.

modelexpress_client/python/tests/test_metrics.py
```diff
@@ -969,3 +970,55 @@ def _label_counts(collector, label):
...
+
+    claimed = _client_families_claimed_by_rust()
+    assert claimed, "parsed an empty CLIENT_FAMILIES; the regex no longer matches"
+
+    missing = [family for family in claimed if family not in exposition]
+    exported = sorted(
+        line.split()[2]
+        for line in exposition.splitlines()
+        if line.startswith("# TYPE ")
+    )
```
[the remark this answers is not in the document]
yixinh-nv: Good catch — now compares exact exported series names, and mutation-checked that the prefix `mx_p2p_transfer_seconds` fails where the substring form accepted it.

helm/README.md
```diff
@@ -78,6 +78,34 @@ The following table lists the configurable parameters of the ModelExpress chart
...
+| `metrics.podMonitor.enabled`                 | PodMonitor for the server. Requires the Prometheus Operator CRDs | `false` |
+| `metrics.podMonitor.additionalLabels`        | Labels the Prometheus `podMonitorSelector` matches. Usually `release: <your-prometheus-release>`, without which it is silently ignored | `{}` |
+| `metrics.podMonitor.interval`                | Scrape interval                                | `30s`    |
+| `metrics.podMonitor.scrapeTimeout`           | Scrape timeout                                 | `10s`    |
+| `metrics.podMonitor.relabelings`             | Passed through to the endpoint's `relabelings` | `[]`     |
+| `metrics.podMonitor.metricRelabelings`       | Passed through to the endpoint's `metricRelabelings` | `[]` |
+| `metrics.clientPodMonitor.enabled`           | PodMonitor for client metrics in your inference pods, which this chart does not deploy | `false` |
+| `metrics.clientPodMonitor.selector`          | Label selector for those pods. Required -- an empty selector matches every pod in the namespace | `{}` |
+| `metrics.clientPodMonitor.portName`          | Named port on the engine pod                   | `mx-metrics` |
```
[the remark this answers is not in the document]
yixinh-nv: It is not required — values.yaml supplies `mx-metrics` and rendering with only the selector set works, so I documented the narrower real hazard instead: that default is the shipped example's port name, and a manifest using a different one renders cleanly then matches no port.

.github/workflows/ci.yml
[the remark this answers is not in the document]
yixinh-nv: Moot now — the helm-verify job has been dropped from this PR to keep it scoped to the chart, so ci.yml is untouched; the hardening will come with the job if I raise it separately.

modelexpress_server/src/metrics/registry.rs
```diff
@@ -224,6 +243,36 @@ impl RegistryMetrics {
...
+        // series per pod for ratio alerts that need sustained traffic to fire
+        // anyway.
+        for result in ClaimResult::ALL {
+            let _ = claims.get_or_create(&ClaimLabels { result });
+        }
+        for result in LeaseResult::ALL {
+            let _ = lease_refreshes.get_or_create(&LeaseLabels { result });
+        }
+        for from in StatusLabel::ALL {
+            for to in StatusLabel::ALL {
```
KavinKrishnan: Could we only create the transition pairs the code can actually produce? This makes all 16 combinations, but only six are reachable. A small REACHABLE list, like TERMINAL below, would avoid exporting ten series that can never move from zero.

helm/templates/prometheusrule.yaml
```diff
@@ -0,0 +1,253 @@
...
+{{- end }}
+
+    - name: modelexpress.download
+      rules:
+        # 15m window against a 5m hold: a single lost lease is above zero for one
+        # window-length, so the hold has to sit well inside it. The counter is
+        # pre-created at zero in RegistryMetrics::register, which is what makes
+        # the first loss in a process visible at all.
+        - alert: MXDownloadLeaseLost
+          expr: sum(increase(mx_download_lease_refresh_total{result="lost"}[15m])) > 0
```
KavinKrishnan: Could we keep the pod or instance label here instead of combining every replica into one alert? Right now this says a lease was lost, but not where. The same applies to the takeover alert below.

helm/templates/podmonitor-client.yaml
```diff
@@ -0,0 +1,66 @@
...
+    {{- end }}
+spec:
+  selector:
+    {{- toYaml $sel | nindent 4 }}
+  {{- with $cfg.namespaceSelector }}
+  namespaceSelector:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  podMetricsEndpoints:
+    - port: {{ $cfg.portName }}
```
KavinKrishnan, line 50: Could we add an alert for absent(mx_build_info{component="client"}) when this client PodMonitor is enabled? A wrong port name renders cleanly but scrapes nothing, so today that looks the same as a healthy setup with no client data.

KavinKrishnan (approved):
Approving overall—the chart, tests, and real-cluster validation look solid. I left three small follow-ups that I think we should address before merging. Also, MXP2PListSourcesErrors still shares the server grpcErrorRatio setting; please give it its own threshold so tuning one alert does not quietly change the other.

modelexpress_server/src/metrics/registry.rs
yixinh-nv: Done in 21e1ca7a — down to the six reachable pairs, and the new test drives the real recording methods rather than restating the list, so adding a transition without listing it fails there.

helm/templates/prometheusrule.yaml
yixinh-nv: Done in 1f860728 — both now `sum by (pod)` and name the replica in the annotations, with a promtool case asserting that one replica losing a lease alerts for that pod alone.

helm/templates/podmonitor-client.yaml
yixinh-nv: Added `MXClientMetricsAbsent` in 50346b78, gated on `metrics.clientPodMonitor.enabled` so absence only alerts where something is actually meant to be scraped.

PR #683, opened 2026-08-25
fix(refit): publish the served device ordinal for non-CUDA FSDP shards

yafshar:
## Problem

`build_fsdp_reshard_manifest` derived the published shard ordinal as:

```python
device_id=served.device.index if served.device.type == "cuda" else 0,
```

Any non-CUDA served buffer was advertised as device 0 regardless of where it
actually lives.

The ordinal is not cosmetic. A receiver builds `session_to_device` from these
manifests, and `NixlReshardTransport._ranges_for` puts it straight into the
remote descriptor tuple `(src_addr, dst_addr, nbytes, device_id)` that becomes
`remote_descs` in `prep_xfer_dlist`. A shard served from `xpu:2` therefore
pointed peers at device 0 - a wrong-device read, or a failure at prep.

CUDA was unaffected, which is why this went unnoticed. Single-device masked runs
(`ZE_AFFINITY_MASK=2`) hide it as well, since the visible ordinal is 0 anyway.

## Fix

Use the served buffer's own ordinal, matching the form the Megatron adapter
already uses (`train/engines/megatron/aliases.py`):

```python
device_id=int(served.device.index or 0),
```

## Test

One parametrized case added to the existing `tests/test_refit_fsdp_publisher.py`,
covering `cuda:3`, `xpu:2`, `xpu:0` and `cpu`. It uses a small `_ServedOn`
stand-in exposing only the four members the function touches, because
`torch.device("xpu:2")` constructs fine on a CPU host but a tensor on that device
does not.

Mutation-checked: reverting the one-line change fails exactly the `xpu:2` case.
The other three cases pass under both the old and new code by design - they guard
that CUDA and unindexed-device behavior is unchanged.

An unindexed (CPU) served tensor still maps to 0, unchanged from before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device identification for published model shards across CUDA, XPU, and CPU environments.
  * Device ordinals are now reported consistently, including devices without an explicitly assigned index.
  * This improves compatibility and reliability when publishing or refitting models on non-CUDA hardware.


zhengluo-nv (approved):

zhengluo-nv:
/ok to test [547caa9](https://github.com/ai-dynamo/modelexpress/pull/683/commits/547caa9e19c8cd2ca8ca8c5327c97f1a8f7aa247)

PR #688, opened 2026-08-27
perf(client): default UCX_MEM_EVENTS=n for NIXL agent creation

AndyDai-nv:
## Summary
- Default `UCX_MEM_EVENTS=n` before NIXL agent construction, unless the deployment already sets it. This disables UCX's memory-event tracking (mem hooks / VMM registration-cache invalidation), which is reported to cut NIXL agent creation time by roughly ~50%.
- Safe by construction: ModelExpress registers every transferred tensor explicitly via `MX_POOL_REG` / `MX_VMM_ARENA` / per-tensor registration, so UCX never needs its own mem-event tracking to catch an unregistered `cudaFree`/`cudaMalloc` on our tensors.
- Registers `UCX_MEM_EVENTS` in `envs.py` for reading (write stays inline in `nixl_transfer.py`, matching the existing `UCX_TLS` / `UCX_NET_DEVICES` pattern) and documents the new default in `docs/DEPLOYMENT.md`.

## Test plan
- [x] `modelexpress_client/python` imports cleanly with the new env var
- [x] `pytest tests/test_envs.py tests/test_ucx_utils.py tests/test_nixl_backend.py tests/test_pool_registration.py` passes (66 passed)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring UCX memory-event tracking through the `UCX_MEM_EVENTS` environment variable.
  * Operator-provided values are preserved and applied to subsequent UCX contexts.
  * The setting defaults to disabled when not specified.

* **Documentation**
  * Documented configuration behavior, defaults, overrides, and tensor-registration safety requirements.


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/688/commits/e11622b6f89d15a4eae4b2e6fbab66341f8a4e6d

modelexpress_client/python/modelexpress/nixl_transfer.py
```diff
@@ -230,6 +238,15 @@ def initialize(self) -> None:
         # semantics and the topology probe.
         ucx_utils.apply_nic_pin_for_device(self._device_id)
 
+        # Default UCX_MEM_EVENTS off: cuts NIXL agent creation time roughly
+        # in half by skipping UCX's mem-hook / VMM invalidation setup, which
+        # ModelExpress's explicit tensor registration never relies on. Set
+        # permanently (no restore) so later UCP contexts inherit it too.
+        # Never override an operator-set value.
+        if not envs.is_set("UCX_MEM_EVENTS"):
+            os.environ["UCX_MEM_EVENTS"] = "n"
```
zhengluo-nv, line 258: Do we want to keep this as an explicit deployment opt-in instead of making it a process-wide default? `UCX_MEM_EVENTS` controls global UCM hooks, not just this NIXL agent: UCX registration caches subscribe to `VM_UNMAPPED` / `MEM_TYPE_FREE` and use those events to invalidate cached regions. Because this assignment is permanent, later UCP contexts in the same process inherit disabled invalidation; it also runs when `MX_NIXL_BACKEND=LIBFABRIC`, even though ModelExpress creates no UCX backend. Explicit MX register/deregister only justifies MX-owned buffers when cleanup ordering holds—it does not make unrelated UCX consumers safe. UCX also reads this option in a shared-library constructor, so if UCX was loaded earlier, this assignment can be a no-op. Please keep the optimization opt-in unless process isolation is guaranteed, and add coverage for the LIBFABRIC/non-MX-UCX cases.

AndyDai-nv: Good catch, all fair points. Pushed a fix: the optimization is now opt-in via `MX_UCX_DISABLE_MEM_EVENTS` (default off, so `UCX_MEM_EVENTS` is left untouched unless an operator explicitly opts in), gated to the UCX backend only (no-op under `MX_NIXL_BACKEND=LIBFABRIC`), and still never overrides an operator-set `UCX_MEM_EVENTS`. Docs and the code comment now call out that this is process-wide/permanent and only safe when the process owns UCX exclusively for MX transfers, plus the shared-library-constructor caveat (setting the env var here is a no-op if UCX was already loaded elsewhere in the process). Added `test_nixl_ucx_mem_events.py` covering opt-in/opt-out, the LIBFABRIC no-op, and the operator-override case.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/688/commits/0b2b0ff168768a8c1ae802608cc1ac78d36e35ad

AndyDai-nv:
@zhengluo-nv updated, could you plz take a look?

zhengluo-nv (approved):

PR #689, opened 2026-08-27
fix(ucx): spread RDMA rails across ranks, and report a throughput floor

KavinKrishnan:
## Summary

- Improve GPU-to-NIC selection by preferring distinct RDMA rails while preserving PCIe and NUMA affinity.
- Add shared minimum-throughput reporting for the Megatron and staged PrimeRL refit paths.
- Document configuration and deployment behavior.

The throughput check reports slow transfers without blocking weight installation. Existing operator overrides remain supported.

## Measured performance

Exercised end to end through the PrimeRL FSDP refit path on GB200 (2 x 4 trainer
GPUs, 1 x 4 inference GPUs, four 400 Gb/s RDMA rails per pod), over 44
per-reader transfer records:

| Metric | Value |
| --- | --- |
| Payload per reader | 15.889 GB |
| Mean wire time | 0.3429 s |
| Range | 0.337 s - 0.393 s |
| Per-reader rate | ~371 Gb/s |
| Four-reader aggregate | ~1.48 Tb/s |

With distinct-rail ranking, all readers land in this band. Without it, readers
can share a NIC. In a paired same-node comparison, a forced single-rail arm
measured 42.9x higher wire time than the spread arm. A separate
default-versus-fixed comparison measured 10.35x, but its two arms ran on
different node pairs, so that figure is not confound-free.

## Validation

- [x] All 26 CI checks pass
- [x] Targeted Python suite: 45 passed
- [x] Ranking coverage includes constrained and fully provisioned GPU/NIC topologies
- [x] Throughput reporting coverage includes both receiver paths, disabled settings, healthy transfers, and threshold breaches
- [x] Documentation examples and environment settings validated

## Scope

Investigation-only timing and batch instrumentation is intentionally excluded.


modelexpress_client/python/modelexpress/ucx_utils.py
```diff
@@ -313,34 +369,62 @@ def probe_nic_pin_for_device(
...
     assignments: dict[int, tuple[str, int]] = {}
-    for gi in sorted(gpu_paths.keys()):
+
+    def _best_score(gi: int) -> int:
+        return max(
+            (_pci_common_depth(gpu_paths[gi], nic_path) for *_, nic_path in nics),
+            default=0,
+        )
+
+    visit_order = sorted(gpu_paths.keys(), key=lambda gi: (-_best_score(gi), gi))
```
zhengluo-nv: **[P1] Make the assignment independent of rank-local CUDA visibility**

The assignment counts only spread rails when every worker sees the same multi-GPU `gpu_paths`. With separate one-GPU `CUDA_VISIBLE_DEVICES` masks, each process gets `torch.cuda.device_count() == 1`, rebuilds `assigned_count` from zero, and computes a one-entry map for local device 0. I reproduced that arrangement using this PR's `_MISAFFINE_GPUS` / `_MISAFFINE_NICS` fixture: the four physical GPU views all selected `mlx5_11:1`, which is the exact convergence this change intends to prevent. Please derive the map from host-wide GPU identities or coordinate it across ranks, and add a regression that invokes the probe from four separate rank-local views. Until then the global-assignment claim in `DEPLOYMENT.md` is not valid for this worker topology.

modelexpress_client/python/modelexpress/refit/reshard/throughput.py
```diff
@@ -0,0 +1,103 @@
...
+    Read at call time so a harness can set it per run. Off unless configured
+    because only the operator knows what their fabric should deliver.
+
+    A non-finite floor disables the check rather than being compared against.
+    ``float("nan")`` parses happily and every comparison against NaN is False, so
+    a fat-fingered value would pass the "is it configured" test and then fail the
+    "is the rate acceptable" test, warning on every single refit and turning the
+    guard into noise. Infinity parses too and would condemn every refit outright.
+    """
+    floor = envs.MX_RESHARD_MIN_GBPS
```
zhengluo-nv, line 50: **[P2] Reject a negative throughput floor instead of silently disabling it**

A finite value such as `MX_RESHARD_MIN_GBPS=-50` passes `min_gbps()` and is then treated as disabled by `warn_if_below_floor()`'s `floor <= 0` return. That makes a configuration typo silently disable the CI/performance signal even though the documented disabled value is specifically `0`. The earlier bot thread for this issue is now outdated because the helper moved into this file, but the behavior still reproduces on the current head. Please warn and return `0.0` for negative values, with a focused regression.

zhengluo-nv (commented):
Review against live head `5d76af1`. The advertised 45-test focused suite passes, but two additional regressions fail: rank-local CUDA visibility recreates rail convergence, and a negative throughput floor silently disables the signal.

modelexpress_client/python/modelexpress/ucx_utils.py
KavinKrishnan: good catch — fixed this by building the gpu-to-nic map from the host pci list instead of each worker’s cuda-visible devices. each rank now gets the same full map and looks up its own gpu by pci address. i also added the four separate one-gpu worker test you suggested: the old code sends all four to mlx5_11, while the fix spreads them across four rails. the deployment note now explains this and calls out the device-isolated fallback.

modelexpress_client/python/modelexpress/refit/reshard/throughput.py
KavinKrishnan: fixed — negative values now warn and return 0, so a stray minus sign can’t silently turn the check off. added a focused test for both -50 and -0.5.

modelexpress_client/python/modelexpress/refit/reshard/receiver.py
```diff
@@ -78,6 +78,43 @@ def _max_gbps() -> float:
...
+    """
+    floor = envs.MX_RESHARD_MIN_GBPS
+    if not math.isfinite(floor):
+        logger.warning(
+            "MX_RESHARD_MIN_GBPS=%r is not a finite number; the throughput floor "
+            "is disabled for this run",
+            floor,
+        )
+        return 0.0
+    return floor
```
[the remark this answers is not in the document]
KavinKrishnan: fixed in e4ad8b5 — negative values now warn and return 0, with regression coverage.

zhengluo-nv:
/ok to test [e4ad8b5](https://github.com/ai-dynamo/modelexpress/pull/689/commits/e4ad8b56573eb3bfff6837209f4e5275cacd7b77)

KavinKrishnan:
/ok to test [6cfe7eb](https://github.com/ai-dynamo/modelexpress/pull/689/commits/6cfe7eb70cb0f87cebbe12b63f30ba04ef9ec2d8)

zhengluo-nv (approved):

PR #690, opened 2026-08-27
refactor(refit): compose RL weight update paths

zhengluo-nv:
## Overview

ModelExpress RL currently supports several useful refit paths, but source discovery, transport, payload handling, engine mutation, and lifecycle were coupled inside engine-specific strategies. That made each new combination—trainer or generator source, full tensor or canonical delta, vLLM or SGLang—require another end-to-end path with duplicated control-plane behavior.

This PR refactors **only the ModelExpress RL layer** into composable responsibilities. Existing ModelExpress inference components are reused through narrow wrappers; their implementations are not redesigned. RL frameworks continue to integrate through the same three public clients:

- `ModelExpressControlClient` owns immutable `WeightVersion` lifecycle.
- `ModelExpressTrainerClient` publishes a version through an explicit trainer context and publication method.
- `ModelExpressGeneratorClient` resolves an ordered source, prepares a verified artifact, and installs it through the active inference engine.

The generator path is now:

```text
WeightVersion
    -> ordered SourceResolver          (generator, trainer, object storage)
    -> compatible UpdateMethod         (full-tensor NIXL, canonical XOR delta)
    -> engine-specific EngineInstaller (vLLM, SGLang)
    -> shared WeightUpdateSession      (lease, stage, install, publish, release)
```

The trainer path similarly separates an explicit engine context (`FSDPTrainerContext` or `MegatronTrainerContext`) from the publication method and shared transport/runtime lifecycle. A new engine primarily supplies its target geometry and installer; a new source supplies a resolver; and a new payload/transport path supplies an update or publication method.

### User-visible changes

- `ModelExpressGeneratorConfig.source_order` exposes generator, trainer, and object-storage fallback policy without exposing NIXL or engine internals upstream.
- Full-tensor trainer publication now requires an explicit `FSDPTrainerContext` or `MegatronTrainerContext` instead of selecting an engine through `MX_TRAINER_ENGINE`.
- vLLM and SGLang mutation remains behind engine installers.
- vLLM now surfaces version-correlated `modelexpress_rl` lifecycle logs through its existing EngineCore logging configuration.
- The Dynamo vLLM RL example and architecture documentation use the composed API.

### Scope

This establishes extension points and preserves the current full-tensor and single-delta behavior. It does not add a new RL-framework integration, multi-delta checkpoint reconstruction, or a new ModelExpress inference-core implementation.

## Validation

- review follow-up at `fed1a679`: 65 affected tests passed; Python compilation and `git diff --check` passed
- focused lifecycle/logging tests: 31 passed
- full relevant vLLM logging, transfer-engine, and generator-client files: 138 passed, 8 skipped
- complete Python suite before the observability follow-up: 1,582 passed, 33 skipped, 1 known failure
- Python compilation: passed
- whitespace and merge-conflict hooks: passed
- DCO sign-off and GPG signatures: verified

### Dynamo multi-step reshard E2E

Validated commit `36182a18` with immutable client image digest `sha256:95b4942b5893768dfd832288f54f523e95f14a4f05d26e70531247af6bbbfe7e`:

- Dynamo DGD with vLLM `0.26.1rc1.dev1177+ga9a17e709`
- `Qwen/Qwen3-30B-A3B` in BF16
- two-rank FSDP/DTensor trainer on one B200 node to a TP1 vLLM generator on a different B200 node
- two distinct source slots: `publisher:global-rank:0` and `publisher:global-rank:1`
- three cumulative changes to `model.embed_tokens.weight`, published as versions `f2d822b8`, `4a8c3223`, and `c290fef2`
- exact WeightVersion UID verified after every install; inference returned `ready` after every refit
- each step captured 18,867 reshard copies with zero unsupported or unattributed sources
- stage time was 18.65s for the first update and 4.66s / 4.68s for subsequent updates, reusing the established transfer workspace and agents
- vLLM logs correlate every UID through start, source selection, geometry capture, preparation, installation, apply, release, and finish
- the 30B TP1 run required `--gpu-memory-utilization 0.35`; `0.45` left insufficient space for the rank-local refit workspace

This qualifies repeated changed-weight full-tensor trainer-to-generator reshard lifecycle and post-update inference health. It does not qualify output-semantic correctness, multiple generators, throughput, a complete RL framework loop, or delta/object-storage recovery.

## Known gaps

- `test_reconstructed_checksum_failure_propagates` remains red because the existing canonical-S3 receiver mutates checkpoint bytes before a target-checksum failure is reported. Rollback is outside this composition refactor.
- the repository YAML hook cannot parse the existing `${RL_COORDINATOR}` template in `rl-job.yaml` before the documented `envsubst` rendering step.
- canonical object-storage fallback still handles only an exact local base plus one delta. Reconstructing from the latest full checkpoint and an ordered delta chain remains a TODO.

## Summary by CodeRabbit

* **New Features**
  * Added runtime-based trainer and generator weight update workflows.
  * Added support for full-tensor transfers and canonical-delta object-storage updates.
  * Added configurable source fallbacks across trainers, generator peers, and object storage.
  * Added explicit FSDP and Megatron trainer context selection.
  * Added SGLang and vLLM checkpoint installation support.

* **Improvements**
  * Simplified configuration by replacing environment-based engine selection with explicit contexts.
  * Improved update validation, retries, cleanup, metrics, and resource lifecycle handling.
  * Optional compression support now loads only when needed.

* **Documentation**
  * Expanded architecture and usage documentation for runtimes, sources, installers, and update methods.


modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
```diff
@@ -0,0 +1,309 @@
...
+            ).encode()
+            self._s3.put(uri=staged.object_storage_uri, data=index)
+
+        staged.publish_object_storage_time = self._clock() - started
+        self.snapshot = staged.candidate_snapshot
+        staged.candidate_snapshot = {}
+        self.current_base_version_id = staged.target_version_id
+        self._metric_delta = staged
+        staged.encoded_deltas.clear()
+        staged.checksums.clear()
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in fed1a679. Rank 0 now captures an index-upload exception and all ranks exchange the failure with `all_gather_object` before any snapshot/base-version state transition. Rank 0 re-raises the original exception and peers raise the propagated failure, so the staged handle remains retryable everywhere. Added `test_s3_index_publish_failure_keeps_handle_retryable` and `test_s3_remote_index_failure_prevents_non_root_state_transition`; the focused affected suite passes: 65 tests.

modelexpress_client/python/modelexpress_rl/train/__init__.py
```diff
@@ -10,12 +10,16 @@
...
+    "FSDPTrainerContext",
+    "MegatronTrainerContext",
     "ModelExpressTrainerClient",
     "ModelExpressTrainerConfig",
     "ObjectStorageConfig",
     "StagedWeightVersionShard",
     "TrainerStagingMode",
+    "TrainerEngineContext",
     "WeightPayloadFormat",
 ]
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in fed1a679. `TrainerEngineContext` now precedes `TrainerStagingMode` in `__all__`, satisfying the alphabetical RUF022 ordering while preserving the export set. The focused affected suite passes: 65 tests.

modelexpress_client/python/modelexpress_rl/train/runtime.py
```diff
@@ -0,0 +1,252 @@
...
+                rpc_timeout_seconds=rpc_timeout_seconds,
+                process_group=process_group,
+                read_launch_tensor=read_launch_tensor,
+                s3=S3Client(
+                    endpoint_url=object_storage.endpoint_url,
+                    region_name=object_storage.region_name,
+                ),
+                clock=lambda: perf_counter(),
+            )
+            return cls(method=method, resources=None)
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in fed1a679. `S3Client` is now created first and explicitly closed if `CanonicalDeltaPublicationMethod` construction raises. Added `test_s3_client_closes_when_publication_method_initialization_fails`; the focused affected suite passes: 65 tests.

modelexpress_client/python/modelexpress_rl/inference/session.py
```diff
@@ -0,0 +1,199 @@
...
+    plan: WeightUpdatePlan
+    prepared: PreparedArtifact
+    lease: Any
+    applied: bool = False
+    apply_result: Any = None
+    released: bool = False
+
+
+class WeightUpdateSession:
+    """Own planning, staging, installation, publication, and cleanup."""
```
nv-hwoo: Could you explain a little on what this Session on the inference side is doing? Also how does this work with the object storage delta case?

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/weight_transfer_engine.py
nv-hwoo, line 1: vllm backend seems to cause trouble with the object storage path:

### Failure
```
RuntimeError: NIXL is not available
```

Vime successfully initialized, published the base version as READY, and reached its first update attempt. The vLLM ModelExpress adapter then initialized a NIXL staged-transfer method, although the experiment was configured for S3 object-storage refits.


### Root cause
The composition runtime defaults a vLLM-capable engine to:
```
GENERATOR → OBJECT_STORAGE
```

Selecting GENERATOR requires NIXL. The core configuration supports forcing:
```
source_order=(WeightSource.OBJECT_STORAGE,)
```

but ModelExpressWeightTransferInitInfo does not expose or forward source_order, so Vime cannot request the intended S3-only composition through the public vLLM adapter.

### Recommended fix
Add a typed source_order field to ModelExpressWeightTransferInitInfo and forward it into ModelExpressGeneratorConfig. Then rerun with object storage as the only source.

modelexpress_client/python/modelexpress_rl/inference/runtime.py
```diff
@@ -0,0 +1,251 @@
...
+        engine = _create_engine_runtime(engine_context)
+        enable_object_storage = object_storage is not None and (
+            source_order is None or WeightSource.OBJECT_STORAGE in source_order
+        )
+        if source_order is None:
+            defaults = []
+            if engine.full_tensor is not None:
+                defaults.append(WeightSource.GENERATOR)
+            defaults.append(
+                WeightSource.OBJECT_STORAGE
```
nv-hwoo: For this refactoring PR, we should maintain the original behavior and skip p2p source when it's object storage delta mode. Currently p2p source is not compatible with the canonical delta method because it updates the model on GPU but not the local checkpoint:
```
  Local canonical checkpoint: v0
    └─ v1 loaded from peer
        ├─ GPU becomes v1
        └─ canonical checkpoint remains v0
            └─ v2 peer miss → S3 delta requires base v1
                └─ exact-base validation fails
```
Since the local checkpoint is stale, if the worker ever tries to update the weights via delta bytes from S3, the resulting weight checksum won't match (because we're apply delta to the stale local checkpoints).

nv-hwoo (commented):
Few other behavioral changes compared to the latest main that agent found:

  ## Other behavioral changes

  - The serving-version base check moved from before leasing to after peer discovery and lease acquisition. Wrong-base updates now create/delete a lease and change from RuntimeError to ValueError. 
 
Main:
https://github.com/ai-dynamo/modelexpress/blob/7c3d511ac4828f65209ea025b155f669e1259a75/modelexpress_client/python/modelexpress_rl/inference/client.py#L396-L408
    
PR:
https://github.com/ai-dynamo/modelexpress/blob/97b706f04f9b9110bdf582548563bd09669c7627/modelexpress_client/python/modelexpress_rl/inference/session.py#L49-L84

  - perf/mx_receive_prepare_time disappeared from real canonical updates. The weight-transfer engine still logs staged metrics, but the new canonical method never records this metric. 
 
Main:
https://github.com/ai-dynamo/modelexpress/blob/7c3d511ac4828f65209ea025b155f669e1259a75/modelexpress_client/python/modelexpress_rl/inference/receiver.py#L454-L460

PR:
https://github.com/ai-dynamo/modelexpress/blob/97b706f04f9b9110bdf582548563bd09669c7627/modelexpress_client/python/modelexpress_rl/inference/methods/canonical_delta.py#L64-L100

  - engine_context became the first ModelExpressTrainerConfig field, breaking positional callers.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/weight_transfer_engine.py
zhengluo-nv: Addressed in b296db45. I did not add a vLLM-specific source_order field because mixed GENERATOR -> OBJECT_STORAGE fallback is not safe yet: a peer install advances GPU weights without advancing the canonical checkpoint base. Instead, any generator configured with object_storage now defaults to OBJECT_STORAGE only, so it does not construct MxClient, NIXL staged transfer, or the full-tensor method. ModelExpressGeneratorConfig also rejects an explicit mixed order until checkpoint synchronization is implemented. The focused generator/runtime/vLLM suite passes 71 tests.

modelexpress_client/python/modelexpress_rl/inference/runtime.py
zhengluo-nv: Addressed in b296db45 by restoring the original object-storage isolation. An object-storage generator now resolves only OBJECT_STORAGE and never initializes or consults the generator peer path. Explicit mixed source orders are rejected, which prevents GPU serving state from advancing independently of the canonical checkpoint base. Regression coverage verifies that an available peer is skipped and the canonical object-storage method is used.

modelexpress_client/python/modelexpress_rl/inference/session.py
zhengluo-nv: Addressed in b296db45. WeightUpdateSession now documents its ownership boundary and lifecycle: it owns the version lease and coordinates planner selection, preparation, installation at the engine safe point, optional peer publication, and release; payload and engine behavior remain in the selected method and installer. For object-storage delta, the composition is ObjectStorageSourceResolver -> CanonicalDeltaUpdateMethod -> PreparedCheckpointArtifact -> engine installer. The client also validates the exact serving base before the session acquires a lease.

zhengluo-nv:
Follow-up for the behavioral review in b296db45:

- Restored canonical exact-base validation before lease acquisition. Wrong-base updates again raise RuntimeError and create/delete no lease; the regression test asserts both counts remain zero.
- Restored the original ModelExpressTrainerConfig positional field order by placing the new engine_context field after the pre-existing fields, with positional compatibility coverage.
- perf/mx_receive_prepare_time remains intentionally omitted from CanonicalDeltaUpdateMethod. This refactor does not yet define a common metrics-emission contract across update methods, so I am avoiding a one-off timing path until that design is added.

Focused generator/runtime/trainer/vLLM validation: 71 passed.

nv-hwoo (approved):

PR #696, opened 2026-08-28
docs(helm): document CRD updates for existing clusters

AndyDai-nv:
## Summary

- document that Helm installs missing CRDs but does not update CRDs that already exist
- tell operators to apply the chart's CRDs before upgrades or installs on clusters with older ModelExpress CRDs
- provide commands for source checkouts and packaged/remote charts in the Helm README and release NOTES
- clarify that standalone Kubernetes deployments must reapply CRDs when upgrading

CRDs are cluster-scoped, so installing a new Helm release in another namespace does not replace an older definition. Without the explicit apply step, an application upgrade can succeed while the API server continues validating ModelExpress resources against a stale schema.

## Testing

- `helm lint helm` with Helm v3.16.4
- `cmp examples/crds.yaml helm/crds/modelexpress-crds.yaml`
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Expanded Kubernetes deployment guidance to apply ModelExpress custom resources before initial installations and upgrades.
  * Added Helm instructions for manually applying custom resources when installing or upgrading, including source and packaged chart workflows.
  * Updated Helm installation notes with reminders for deployments upgrading from older versions.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -21,7 +21,33 @@ helm repo add modelexpress https://your-repo-url
 helm repo update
 ```
 
-### 2. Install the chart
+### 2. Install or update CRDs for the Kubernetes backend
+
+The chart ships the `ModelMetadata` and `ModelCacheEntry` CRDs in its `crds/`
+directory. Helm installs CRDs that are missing during `helm install`, but it does
+not update CRDs that already exist during either `helm install` or
+`helm upgrade`. Because CRDs are cluster-scoped, an older definition can remain
+even when installing a new release in a different namespace.
+
+Before upgrading the chart, or installing it on a cluster that already has
+ModelExpress CRDs, apply the definitions from the chart with cluster-admin
+credentials:
+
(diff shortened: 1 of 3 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/696/commits/e9df3cf6cf78896df4902069246cc845dc34976a

zhengluo-nv (approved):

helm/README.md
```diff
@@ -200,6 +226,9 @@ env:
 ## Upgrading
 
 ```bash
+# Helm does not upgrade existing CRDs, so update them first.
+kubectl apply -f helm/crds/modelexpress-crds.yaml
```
zhengluo-nv, line 230: Followup: can we bundle the CRDs into the helm chart?

AndyDai-nv: sure

PR #705, opened 2026-09-01
feat(refit): add support for full checkpoint cadence

nv-hwoo:
### Summary

  Adds FULL_HF_CHECKPOINT as an S3 refit payload, allowing frameworks to establish a new full Hugging
  Face checkpoint base at a configurable cadence while continuing to use XOR deltas between full
  checkpoints.

  v24: XOR_DELTA(base=v23)
  v25: FULL_HF_CHECKPOINT(no base)  ← new canonical base
  v26: XOR_DELTA(base=v25)

  ModelExpress does not run a timer or background checkpoint scheduler. The framework selects the
  payload format synchronously while creating each WeightVersion.

  ### Cadence configuration

  The cadence is opt-in and disabled by default. With a cadence of 25 versions:

  v1–v24   XOR_DELTA
  v25      FULL_HF_CHECKPOINT
  v26–v49  XOR_DELTA
  v50      FULL_HF_CHECKPOINT

  The framework records that choice in the global WeightVersion. Every trainer rank fetches the same
  version and dispatches using its payload_format.

  ### Motivation

  Canonical S3 refit previously supported only sequential XOR deltas from the initial seed
  checkpoint.

  A framework can now establish a new canonical base at selected points during a long-running
  training job. Subsequent deltas reference that full checkpoint instead of the original seed or an
  indefinitely long delta lineage.

  ### Protocol and control plane

  - Adds WEIGHT_PAYLOAD_FORMAT_FULL_HF_CHECKPOINT.
  - Requires full checkpoint versions to:
      - Use S3 object storage.
      - Omit base_version_id.

  - Preserves existing FULL_TENSOR and XOR_DELTA behavior.
  - Keeps the immutable WeightVersion as the source of truth for payload selection and S3 URI.

  ### Trainer publication

  - Reads the target WeightVersion to select delta or full publication.
  - Captures full HF tensors into the retained CPU snapshot.
  - Re-batches snapshot tensors according to MX_REFIT_FULL_CHECKPOINT_BATCH_BYTES, defaulting to 4
    GiB.

  - Uploads full-checkpoint safetensors objects concurrently through the ModelExpress S3 client.
  - Stores per-tensor Adler-32 checksums in safetensors metadata.
  - Publishes a standard HF safetensors index.
  - Converts the published snapshot to byte views and uses it as the next XOR base.
  - Does not create a trainer-side temporary checkpoint.

  ### Generator preparation

  - Dispatches object-storage updates using the WeightVersion payload format.
  - Preserves the existing XOR-delta reconstruction path.
  - Downloads full-checkpoint batches concurrently.
  - Verifies each tensor checksum before copying that tensor.
  - Copies tensors directly into the existing mmap-backed refit checkpoint.
  - Does not materialize a second full local checkpoint.
  - Reuses shared bounded thread-pool utilities for capture, upload, download, and delta application.

  ### Local checkpoint state

  The mutable host-local checkpoint uses an explicit transaction state:

  READY(base) → UPDATING → READY(target)

  - State changes to UPDATING before seed replacement or weight mutation.
  - State returns to READY only after all writes and verification succeed.
  - Preparation and installation reject an interrupted UPDATING checkpoint.
  - reset_initial_checkpoint() owns the reset transaction:
      - Mark UPDATING.
      - Replace the mutable checkpoint from seed_checkpoint_path.
      - Rebuild local tensor mappings.
      - Mark the initial version READY.

  - Co-located TP processes coordinate through the filesystem lock. One process performs the update
    while peers reuse the resulting READY checkpoint.

  - Local safetensors mappings are parsed once during initialization and reused because updates
    preserve the local checkpoint layout.

  ### Compatibility and limitations

  - Existing XOR-only workflows remain unchanged.
  - Full checkpoint publication currently supports S3.
  - Full checkpoints must remain layout-compatible with the seeded local checkpoint because the
    receiver overwrites existing mmap destinations.

  - An interrupted update remains UPDATING until initialization resets the local checkpoint.
  - State writes use atomic replacement but do not yet provide explicit power-loss durability through
    fsync.

  - ModelStreamer integration remains a future optimization.

  ### Validation

  - 164 focused Python refit and engine-integration tests passed.
  - Rust refit service tests passed.
  - Rust formatting and targeted clippy passed.
  - Diff-scoped pre-commit checks passed.
  - E2E validation using Slime and Vime running 100 steps with Qwen3-30B-A3B + AWS S3 Bucket. At every 25th update weight step, it pushed full HF checkpoint to S3 (avg ~720s) and delta bytes for the rest (avg ~42s).

## Summary by CodeRabbit

* **New Features**
  * Added support for publishing and consuming full Hugging Face checkpoints alongside XOR deltas.
  * Added concurrent, batched checkpoint uploads and downloads with configurable batch sizing.
  * Added checkpoint validation, tensor checksums, recovery handling, and support for rebasing from full checkpoints to deltas.
  * Added configuration for full-checkpoint publication cadence and batch size.

* **Bug Fixes**
  * Improved interrupted-update handling and prevents reuse of incomplete checkpoint state.

* **Documentation**
  * Updated architecture, deployment, storage, and client documentation for full checkpoint workflows.


modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
```diff
@@ -213,11 +256,152 @@ def collect_one() -> None:
...
+                weight_map[name] = shard_name
+        index = json.dumps(
+            {
+                "metadata": {"total_size": total_size},
+                "weight_map": weight_map,
+            },
+            sort_keys=True,
+            separators=(",", ":"),
+        ).encode()
+        self._s3.put(uri=staged.object_storage_uri, data=index)
```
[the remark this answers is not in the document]
nv-hwoo: If rank 0 fails, that should be handled by the framework/caller or the RL orchestrator

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -226,31 +280,40 @@ def prepare(self, version: _S3Version) -> PreparedCheckpoint:
...
+            if version.payload_format is WeightPayloadFormat.XOR_DELTA:
+                weight_map, self.decompressor = _parse_delta_manifest(index_data)
+                download_started = time.perf_counter()
+                shards = self._download_deltas(weight_map, version.uri)
+                download_time = time.perf_counter() - download_started
+                apply_started = time.perf_counter()
+                self._apply_shards(shards)
+                apply_time = time.perf_counter() - apply_started
+            else:
+                weight_map = _parse_full_checkpoint_manifest(index_data)
```
anson627, line 300: If weight_map omits a tensor, the old value remains and the hybrid checkpoint is marked READY, we probably want to require the full tensor set to match the local checkpoint before mutation

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -261,31 +324,111 @@ def _download_deltas(
...
                 except Exception as error:
                     raise RuntimeError(
-                        f"canonical delta download failed for {filename!r}"
+                        f"full HF checkpoint download failed for {filename!r}"
                     ) from error
-                shards[filename] = (data, names)
-        return shards
+                download_time = time.perf_counter() - download_started
+
+                header, data_start = read_safetensors_header(data, repr(filename))
```
anson627, line 379: We probably want to check each source tensor’s dtype, shape, and byte size against the local checkpoint, otherwise incompatible bytes can be copied under the old safetensors metadata and silently corrupt the model.

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
```diff
@@ -213,11 +256,152 @@ def collect_one() -> None:
...
+            uploads,
+            upload,
+            max_workers=rl_envs.MX_S3_UPLOAD_WORKERS,
+            thread_name_prefix="modelexpress-full-hf-upload",
+        ):
+            local_map.update(batch_map)
+            staged.wire_bytes += wire_bytes
+
+        contributions = [None] * self._world_size if self._rank == 0 else None
+        dist.gather_object(
```
anson627, line 356: If one rank fails before gather_object, other ranks can block indefinitely, do we want to exchange upload results across all ranks before entering the manifest gather?

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -160,22 +170,68 @@ def initialize(self) -> None:
...
-                self.locations, _ = index_checkpoint_tensors(self.local_checkpoint)
-                reusable = state.get("files") == _checkpoint_files_state(self.locations)
-            if not reusable:
-                _seed_checkpoint(self.seed_checkpoint_path, self.local_checkpoint)
-                self.locations, _ = index_checkpoint_tensors(self.local_checkpoint)
-                _write_state(
-                    self.state_path,
-                    {
-                        "version": self.initial_version,
-                        "files": _checkpoint_files_state(self.locations),
```
anson627, line 176: This removes the existing file fingerprint checks, allowing modified checkpoint files to be reused or installed. should we retain and update the fingerprint alongside the new READY/UPDATING state?

anson627 (approved):

modelexpress_client/python/modelexpress_rl/inference/receiver.py
nv-hwoo: READY/UPDATING handles all ModelExpress-controlled mutation failures already. Previously it was looking at all the file bytes and modified times of every files to check whether the checkpoint was modified, which I think was a bit unnecessary and could be achieved same with a simple checkpoint states. Do you envision any case where we might want to check these fingerprints as well?

modelexpress_client/python/modelexpress_rl/inference/receiver.py
nv-hwoo: Sure, we can add these as well. I was only checking the weight checksums at the moment

modelexpress_client/python/modelexpress_rl/inference/receiver.py
nv-hwoo: Good point, I will add a check

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
nv-hwoo: Good point. I will address this as a separate, follow-up PR since 
1. The trainer side as a whole needs proper failure handling around it's collectives so it's not just this one
2. This is on the trainer side, so not a blocker in our case

zhengluo-nv (approved):

modelexpress_client/python/modelexpress_rl/inference/receiver.py
anson627: READY/UPDATING covers ModelExpress-controlled mutations, but not changes outside that transaction path, for example, a file being replaced/truncated by another process, a reused volume containing stale files, or modification between preparation and installation, we can cover this in future iterations

PR #709, opened 2026-09-01
feat(refit): harden immutable checkpoint lifecycle

zhengluo-nv:
## Summary

- synchronize distributed full-checkpoint publication failures before manifest collection and retain file fingerprints in READY state
- store canonical full checkpoints, deltas, and resolved chains as immutable host-local artifacts
- materialize delta targets into version-scoped install directories and advance `active.json` only after successful engine installation
- extract filesystem layout, locking, persistence, fingerprints, activation, and quota enforcement into `LocalCheckpointStore`
- add an optional per-model payload-byte quota that evicts stale materializations before stale canonical artifacts while protecting active/prepared lineages
- document the cache lifecycle, quota, and rollback-safe capacity requirements

## Validation

- 1,666 passed, 33 skipped across the full Python suite
- 81 passed across the focused checkpoint-store, S3 receiver, and vLLM weight-transfer suites
- Python bytecode compilation for the changed persistence and receiver modules
- repository pre-commit hooks for all changed files
- `git diff --check`

## Qualification limits

- AWS GPU end-to-end validation is still pending
- the quota is optional for backward compatibility; unset means unlimited
- rollback-safe preparation must fit the active lineage plus the incoming artifact/materialization, even though stale entries are evicted
- the quota bounds ModelExpress payload files, not unrelated consumers of the underlying filesystem
- `active.json` is host-local; global serving-version commit remains generator orchestration ownership

## Planned E2E

Use the AWS cluster to validate full checkpoint -> delta -> delta installation, parameter/generation parity, restart reuse, quota rejection/eviction, and failure rollback before marking this ready for review.

## Summary by CodeRabbit

* **New Features**
  * Added host-local caching for full checkpoints, deltas, and reconstructed versions.
  * Added configurable checkpoint cache size limits with automatic eviction.
  * Added checkpoint integrity, source, and checksum validation.
  * Added support for safely reusing cached checkpoints across restarts.

* **Bug Fixes**
  * Failed downloads, reconstruction, or installation now preserve the previously active checkpoint.
  * Distributed checkpoint publication errors now propagate consistently without advancing version state.

* **Documentation**
  * Updated architecture and setup guidance for checkpoint caching, lineage, activation, and storage limits.


modelexpress_client/python/modelexpress_rl/inference/checkpoint_store.py
```diff
@@ -0,0 +1,333 @@
...
+                protected.update(
+                    self.delta_path(delta) for delta in deltas if isinstance(delta, str)
+                )
+        return protected
+
+    def _eviction_candidates(self, protected: set[Path]) -> list[Path]:
+        groups = (
+            self.materialized_cache,
+            self.chain_cache,
+            self.delta_cache,
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in 4c6233f7. Chain manifests are no longer independent quota-eviction candidates. A manifest is removed only after that version has no full, delta, or materialized artifact. The eviction loop now subtracts payload bytes instead of rescanning the cache, and the regression test verifies both chain survival and final cleanup.

modelexpress_client/python/tests/test_vllm_weight_transfer_engine.py
```diff
@@ -119,6 +121,7 @@ def test_weight_transfer_engine_parses_vime_object_storage_init_info(monkeypatch
         {"initial_base_version_id": "base-a"},
         {"seed_checkpoint_path": "/models/launch"},
         {"refit_checkpoint_dir": "/cache/modelexpress"},
+        {"refit_checkpoint_max_size_bytes": 1_000_000},
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in 4c6233f7. The optional refit_checkpoint_max_size_bytes entry was removed from the incomplete-configuration parameterization. Nonpositive quota validation remains covered separately.

modelexpress_client/python/tests/test_refit_s3_receiver.py
```diff
@@ -826,14 +1140,14 @@ def test_canonical_s3_full_download_failure_leaves_checkpoint_updating(
...
-    state = json.loads(adapter._checkpoint.state_path.read_text())
+    state = json.loads(adapter._checkpoint.store.state_path.read_text())
     assert state["status"] == "UPDATING"
     assert state["version"] == "base-a"
     assert torch.equal(
         load_file(adapter._checkpoint.local_checkpoint / "model.safetensors")["weight"],
         torch.tensor([1.0, 2.0]),
     )
-    assert not (adapter._checkpoint.cache / "checkpoint.tmp").exists()
+    assert not (adapter._checkpoint.store.cache / "checkpoint.tmp").exists()
```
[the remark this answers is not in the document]
zhengluo-nv: Fixed in 4c6233f7. The failure test now checks the actual full/full-a.tmp path created by replace_directory.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -318,6 +358,182 @@ def prepare(self, version: _S3Version) -> PreparedCheckpoint:
...
+                    manifests[0].read_bytes()
+                )
+                shards = {
+                    filename: (
+                        (delta_path / filename).read_bytes(),
+                        names,
+                    )
+                    for filename, names in _group_tensors_by_shard(delta_map).items()
+                }
+                self._apply_shards(shards)
```
nv-hwoo: We should use the latest materialized checkpoint to apply a single delta instead of always applying delta chains from the latest full checkpoint. This would mean that with full checkpoint interval N=25, we're applying total 300 deltas:
```
   Target         Deltas reapplied
  ━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━
   Step 1                        1
  ─────────  ──────────────────────
   Step 2                        2
  ─────────  ──────────────────────
   Step 3                        3
  ─────────  ──────────────────────
   …                             …
  ─────────  ──────────────────────
   Step 24                      24
  ─────────  ──────────────────────
   Total      1 + 2 + … + 24 = 300

```

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -318,6 +358,182 @@ def prepare(self, version: _S3Version) -> PreparedCheckpoint:
...
+            self.store.path_size_bytes(full_path),
+            protected_versions={
+                self.store.active_version(),
+                version.version_id,
+            },
+        )
+        with self.store.replace_directory(
+            target,
+            copy_from=full_path,
+        ) as temporary:
```
nv-hwoo, line 514: We're also creating a full materialized checkpoint per delta versions:
```
 full/
    v0/                    ~61 GB

  deltas/
    v1/
    v2/
    v3/

  materialized/
    v1/                    ~61 GB
    v2/                    ~61 GB
    v3/                    ~61 GB
```
We should instead just keep one single materialized checkpoint for reloading purposes.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -547,18 +758,35 @@ def apply_one(item) -> None:
...
+                or state.get("files")
+                != checkpoint_files_state(self.checkpoint_paths)
+            ):
+                raise ReceiverInstallError(
+                    "prepared checkpoint changed during installation"
+                )
+            self.store.activate(prepared.target_version)
+            self.store.enforce_capacity(
+                protected_versions={prepared.target_version},
+            )
```
nv-hwoo: Looks like there is some race condition concern in this two phased installation lock:
```
  Process A: [shared lock ─── install v2 ───]     [exclusive ─ activate v2]
                                               ^
                                               gap

  Process B:                                  [exclusive ─ prepare v3 ─────]

```
Should we add something like a INSTALLING fence to prevent other prepare to interrupt the installation?

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
```diff
@@ -343,14 +343,42 @@ def upload(
...
+                max_workers=rl_envs.MX_S3_UPLOAD_WORKERS,
+                thread_name_prefix="modelexpress-full-hf-upload",
+            ):
+                local_map.update(batch_map)
+                staged.wire_bytes += wire_bytes
+        except Exception as error:  # noqa: BLE001 - synchronize rank failures.
+            upload_error = error
+            upload_error_message = f"{type(error).__name__}: {error}"
+
+        upload_errors = [None] * self._world_size
```
nv-hwoo: Should we handle this in a separate PR? I think the trainer side overall lacks these distributed failure handling

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: Fixed in e99a3747. The previous materialization is retained only through the rollback-safe staging/install window. After activation, all older materializations are pruned, leaving one steady-state materialized checkpoint.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: Fixed in e99a3747 with a dedicated cross-process installation fence. Co-located installers hold .install.lock shared through reload and activation; prepare/initialize require it exclusively, so prepare cannot enter the old gap. A flock was used instead of persisted INSTALLING state so process exit releases the fence automatically.

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
zhengluo-nv: Agreed. e99a3747 removes the trainer-side distributed publication failure handling and its two tests from this PR; those files now match the PR base. The broader trainer failure contract should be handled holistically in a separate PR.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: Fixed in e99a3747. Delta preparation now copies the exact active full/materialized checkpoint and applies only the incoming delta. The resolved canonical chain is still retained and its prior artifacts are integrity-checked, but they are not replayed.

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
zhengluo-nv: Updated

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: Updated

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: I think since we already have the eviction, it will delete the materialized checkpoint first when the disk usage is above the max size. Keeping more than one materialized checkpoint may still be helpful in some optimization use case, e.g. when a new replica cold starts, it can p2p transfer a materialized checkpoint from a peer

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: Fixed

nv-hwoo (approved):

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -318,6 +359,182 @@ def prepare(self, version: _S3Version) -> PreparedCheckpoint:
...
+            self.store.verify_artifact_source(
+                artifact,
+                _source_identity(version),
+            )
+        else:
+            protected_versions = {
+                self.store.active_version(),
+                version.version_id,
+            }
+            shards = self._download_deltas(weight_map, version.uri)
```
anson627, line 478: Could we validate delta shard filenames before using them as local paths, similar to `_parse_full_checkpoint_manifest`? Currently an absolute path or .. segment from weight_map can escape the artifact directory when `(temporary / filename).write_bytes(data)` runs. Restricting filenames to safe basenames would keep all writes inside the checkpoint cache

modelexpress_client/python/modelexpress_rl/inference/checkpoint_store.py
```diff
@@ -0,0 +1,390 @@
...
+                shutil.copytree(copy_from, temporary)
+            yield temporary
+            shutil.rmtree(target, ignore_errors=True)
+            temporary.replace(target)
+        except Exception:
+            shutil.rmtree(temporary, ignore_errors=True)
+            raise
+
+    def _version_path(self, root: Path, version: str) -> Path:
+        return root / quote(version, safe="")
```
anson627, line 136: Could we ensure encoded version paths remain direct children of their cache directories? `quote()` leaves . and .. unchanged, while version IDs currently allow any non-empty string. For example, `full_path("..")` resolves to the model cache root, which can cause directory replacement to remove cache metadata and artifacts. Rejecting or explicitly encoding dot-segment IDs would prevent this

anson627 (approved):

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: Will address in the next PR. Merge now to unblock development

modelexpress_client/python/modelexpress_rl/inference/checkpoint_store.py
zhengluo-nv: Will address in the next PR. Merge now to unblock development

PR #710, opened 2026-09-01
feat(refit): add target-version delta replay

### LINKED ISSUE #703: Add target-version replay support to the refit API
## Problem statement

The refit API currently applies a delta only to its immediate base. New, restarted, or lagging workers may be several revisions behind and must replay every missing delta in order:

```text
full weights v0 → delta v1 → delta v2 → delta v3
```

Here, `v0` is a full checkpoint, while `v1`, `v2`, and `v3` are deltas relative to their preceding versions. Applying only the latest delta would produce invalid weights. The API should safely bring a worker from a verified local revision to an explicitly requested target.

## Contract and behavior

The caller may specify any target version, including one that is not adjacent to the current local version. MX must resolve the complete chain from the MX server cache before modifying the checkpoint. The caller is responsible for ensuring that the full chain has already been registered with MX and is available in the cache. If the target or any intermediate revision is missing, MX must fail fast with an error that clearly identifies the missing revision.

Replay must validate the resolved chain, including model and format identity, checksum continuity, cycles, and a maximum chain length. It must then apply and verify each delta in order while exposing no intermediate state. Only the final target is installed into the inference engine, and an already-installed target is a successful no-op.

Failures must never expose or report an invalid checkpoint as ready. The serving model should remain unchanged if replay fails before installation; if installation may have partially changed weights, the worker must be marked unsafe. Errors should identify the target and failing revision.

The caller remains responsible for selecting the target revision.

## Test plan

- Cover direct replay, multi-hop replay from a full checkpoint, and already-at-target requests.
- Fail fast when the target or an intermediate revision is missing from the MX server cache.
- Reject invalid revisions, broken chains, cycles, format mismatches, corruption, and excessive chain length.
- Verify safe recovery after download, validation, installation, or interruption failures.
- Confirm byte equality and inference parity with a full target load, with exactly one final engine installation.
- Verify concurrent ranks do not duplicate replay or observe intermediate revisions.

anson627:
## Summary

Adds target-version replay for canonical refit checkpoints by resolving and validating the complete delta chain before mutation, then installing only the final target. Replay is bounded, lease-protected, safe across shared ranks, and fences workers after potentially partial installation failures.

Closes #703.

## Verification

Validated end-to-end on B200 GPUs with images built:

- Packaged versions: ModelExpress 0.5.1, Dynamo 1.3.0, and vLLM 0.27.1
- Required vLLM weight-transfer imports passed before launching the engine.
- On one B200, a non-adjacent `v0 -> v2` request replayed the `v1` and `v2` XOR deltas and performed exactly one final v2 installation.
- The replayed v2 checkpoint matched the independently prepared full v2 checkpoint: 311 tensors, 1,503,264,768 bytes, aggregate SHA-256 `d85667c10f8122a1d62f6a2a40cba2e9bf94e9574ad3fa5eb363d079e70ae95a`.
- Deterministic inference after replay matched a clean engine loaded directly from full v2, including content, finish reason, and token usage.
- A missing target failed with a clear diagnostic without corrupting the serving engine; inference remained unchanged and the next valid update succeeded.
- Re-requesting the current v2 target completed as a no-op without another installation.
- TP=2 passed on two B200 GPUs after mounting a 1 GiB memory-backed `/dev/shm`. Both ranks installed v2; only rank 0 downloaded and replayed the shared chain, while rank 1 reported zero download/apply time. The TP=2 materialized checkpoint matched the same full-v2 hash and served inference successfully.

## Summary by CodeRabbit

* **New Features**
  * Added support for replaying multiple weight revisions in sequence.
  * Added configurable limits for replay-chain length, defaulting to 64 revisions.
  * Reapplying an already-serving version now completes as a no-op without unnecessary transfer.
  * Added comprehensive validation for revision chains and delta metadata.

* **Bug Fixes**
  * Failed installations now mark checkpoints as unsafe and block further updates until recovery.
  * Updates apply the complete chain atomically and install only the requested final version.

### DIFF
--- a/docs/S3_DELTA_WEIGHT_REFIT.md
+++ b/docs/S3_DELTA_WEIGHT_REFIT.md
@@ -181,6 +181,7 @@ response = requests.post(
             "refit_checkpoint_max_size_gb": 500,
             "object_storage_region_name": "us-west-2",
             "max_transfer_attempts": 3,
+            "max_replay_chain_length": 64,
             "rpc_timeout_seconds": 30,
         }
     },
@@ -259,6 +260,17 @@ activation.
       config.json
       ...
 ```
+
+The generator may request a target several revisions ahead of its active
+version. ModelExpress first resolves the complete READY chain, rejecting cycles,
+missing or incompatible revisions, and chains longer than
+`max_replay_chain_length` (64 by default). It then prepares the ordered chain as
+one immutable target checkpoint and installs only that final target. If engine
+installation starts and fails, the checkpoint remains `READY`, `active.json`
+continues to identify the last successfully installed version, and the local
+engine is marked uncertain. The next request may reinstall that active version
+or install any target reconstructed from it; either successful installation
+clears the uncertain state.
 
 All ranks sharing one host filesystem can share the same cache. Each host without
 a shared filesystem needs its own cache.
--- a/modelexpress_client/python/modelexpress_rl/inference/client.py
+++ b/modelexpress_client/python/modelexpress_rl/inference/client.py
@@ -9,6 +9,7 @@ import logging
 import threading
 import uuid
 from dataclasses import dataclass
+from enum import Enum
 from typing import Any
 
 import grpc
@@ -29,6 +30,11 @@ from .runtime import GeneratorRuntime
 from .session import SessionUpdate
 
 logger = logging.getLogger("modelexpress_rl.inference.client")
+
+
+class _EngineState(str, Enum):
+    READY = "READY"
+    UNCERTAIN = "UNCERTAIN"
 
 
 def _required(value: str, name: str) -> str:
@@ -55,6 +61,8 @@ class ModelExpressGeneratorConfig:
     lease_ttl_seconds: int | None = None
     # Maximum source-discovery and transfer attempts for one staged update.
     max_transfer_attempts: int = 3
+    # Maximum number of payload revisions applied by one target replay.
+    max_replay_chain_length: int = 64
     # Deadline applied independently to each control-plane or manifest RPC.
     rpc_timeout_seconds: float = 30.0
     # Canonical object-storage checkpoint settings.
@@ -74,6 +82,9 @@ class ModelExpressGeneratorConfig:
             rl_envs.require_positive_int(self.lease_ttl_seconds, "lease_ttl_seconds")
         rl_envs.require_positive_int(
             self.max_transfer_attempts, "max_transfer_attempts"
+        )
+        rl_envs.require_positive_int(
+            self.max_replay_chain_length, "max_replay_chain_length"
         )
         rl_envs.require_positive_float(self.rpc_timeout_seconds, "rpc_timeout_seconds")
         if self.source_order is not None:
@@ -116,11 +127,12 @@ class StagedWeightHandle:
         *,
         client: ModelExpressGeneratorClient,
         version_id: str,
-        update: SessionUpdate,
+        update: SessionUpdate | None,
     ) -> None:
         self._client = client
         self.version_id = version_id
         self._update = update
+        self._no_op_released = False
 
     def release(self) -> None:
         """Release local staging buffers; repeated calls are idempotent."""
@@ -129,11 +141,15 @@ class StagedWeightHandle:
     @property
     def metrics(self) -> dict[str, float]:
         """Return preparation metrics exposed by the selected adapter."""
+        if self._update is None:
+            return {}
         return self._update.prepared.metrics
 
     @property
     def applied(self) -> bool:
         """Return whether engine installation completed."""
+        if self._update is None:
+            return True
         return self._update.applied
 
 
@@ -179,6 +195,7 @@ class ModelExpressGeneratorClient:
         self._operation_lock = threading.RLock()
         self._active_handle: StagedWeightHandle | None = None
         self._serving_version_id: str | None = None
+        self._engine_state = _EngineState.READY
         self._runtime: GeneratorRuntime | None = None
         self._closed = False
 
@@ -219,6 +236,7 @@ class ModelExpressGeneratorClient:
         client._registration_ttl_seconds = registration_ttl_seconds
         client._lease_ttl_seconds = lease_ttl_seconds
         client._rpc_timeout_seconds = config.rpc_timeout_seconds
+        client._max_replay_chain_length = config.max_replay_chain_length
         try:
             runtime = GeneratorRuntime.initialize(
                 engine_context=config.engine_context,
@@ -260,9 +278,32 @@ class ModelExpressGeneratorClient:
                 if self._active_handle.version_id == version.version_id:
                     return self._active_handle
                 raise RuntimeError("another generator update is still active")
-            ready = self._get_ready_version(version.version_id)
             assert self._runtime is not None
-            update = self._runtime.session.stage(ready)
+            if (
+                self._runtime.initial_version_id is not None
+                and version.version_id == self._serving_version_id
+                and self._engine_state is _EngineState.READY
+            ):
+                self._fetch_ready_version(
+                    version.version_id,
+                    target_version_id=version.version_id,
+                )
+                self._active_handle = StagedWeightHandle(
+                    client=self,
+                    version_id=version.version_id,
+                    update=None,
+                )
+                return self._active_handle
+            if self._runtime.initial_version_id is not None:
+                chain = self._resolve_replay_chain(version.version_id)
+                update = (
+                    self._runtime.session.stage(chain[0])
+                    if len(chain) == 1
+                    else self._runtime.session.stage_chain(chain)
+                )
+            else:
+                ready = self._get_ready_version(version.version_id)
+                update = self._runtime.session.stage(ready)
             self._active_handle = StagedWeightHandle(
                 client=self,
                 version_id=version.version_id,
@@ -275,11 +316,21 @@ class ModelExpressGeneratorClient:
         if not isinstance(staged, StagedWeightHandle) or staged._client is not self:
             raise ValueError("staged handle does not belong to this client")
         with self._operation_lock:
+            if staged._update is None:
+                if staged._no_op_released:
+                    raise RuntimeError("staged weight has already been released")
+                return None
             if staged._update.released:
                 raise RuntimeError("staged weight has already been released")
             assert self._runtime is not None
-            result = self._runtime.session.apply(staged._update)
+            try:
+                result = self._runtime.session.apply(staged._update)
+            except BaseException:
+                if staged._update.installation_started and not staged._update.applied:
+                    self._engine_state = _EngineState.UNCERTAIN
+                raise
             self._serving_version_id = staged.version_id
+            self._engine_state = _EngineState.READY
             return result
 
     def close(self) -> None:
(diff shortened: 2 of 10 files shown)


anson627:
vLLM FlashInfer mismatch issue needed workaround to finish the verification:
https://github.com/vllm-project/vllm/issues/49804

docs/S3_DELTA_WEIGHT_REFIT.md
```diff
@@ -180,6 +180,7 @@ response = requests.post(
             "refit_checkpoint_dir": "/var/cache/modelexpress",
             "object_storage_region_name": "us-west-2",
             "max_transfer_attempts": 3,
+            "max_replay_chain_length": 64,
```
c2w-sea: About having this limit, is the concern that one reply may take too long? 

I don't know if we should enforce a default limit. The actual issue depends on use cases and model size. So I'd suggest either default to have no limit by default (null) and make this configurable by MX users

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -548,7 +624,7 @@ def apply_one(item) -> None:
     @contextmanager
     def installation_context(self, prepared: PreparedCheckpoint):
         with self.lock_path.open("a+") as handle:
-            fcntl.flock(handle, fcntl.LOCK_SH)
+            fcntl.flock(handle, fcntl.LOCK_EX)
```
c2w-sea: what does it mean to swap the lock here? This seems to have wider implication

modelexpress_client/python/modelexpress_rl/inference/session.py
```diff
@@ -163,6 +227,14 @@ def apply(self, update: SessionUpdate) -> Any:
             return update.apply_result
         except BaseException as error:
             primary_error = error
+            if install_started:
+                try:
+                    update.plan.method.installation_failed(update.prepared)
```
c2w-sea, line 232: agent reported no test coverage for this. 

"installation_failed and the install_started branch have no direct test. Add a session-level test where the installer raises, and one where installation_context.__enter__ raises and no fence happens."

modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -260,9 +271,31 @@ def stage_weight(self, *, version: WeightVersionRef) -> StagedWeightHandle:
...
+                self._fetch_ready_version(
+                    version.version_id,
+                    target_version_id=version.version_id,
+                )
+                self._active_handle = StagedWeightHandle(
+                    client=self,
+                    version_id=version.version_id,
+                    update=None,
+                )
+                return self._active_handle
```
c2w-sea, line 296: for this no-op path, there may have a re-entry problem here

Assume original version is v1, a new update to v2
1. if v2 install failed and serving instance is not functional, the `_serving_version_id` can still remains v1 but the live serving weight is actually broken
2. A re-attempt on v1 will be no-op and pass through here, but that's wrong because we want to reset the serving weight to v1

c2w-sea (commented):
Initial comments, not yet finished review

modelexpress_client/python/modelexpress_rl/inference/receiver.py
anson627: Good point. The exclusive lock was intended to make the READY → UNSAFE transition atomic with installation failure, but it also serializes installation across co-located ranks. With #709’s dedicated installation fence, we can keep concurrent installers on a shared lock, block preparation exclusively, and update state separately under the cache lock. I’ll reconcile this during the rebase rather than retain the broader lock change

modelexpress_client/python/modelexpress_rl/inference/client.py
anson627: You’re right. `_serving_version_id records` the last successful version, but it does not prove the engine is still healthy after a partially failed installation. The no-op path must be disabled once installation has started and failed. The immediate safe behavior is to fence subsequent requests instead of incorrectly reporting success, recovery can then explicitly reinstall the last known-good version from the immutable checkpoint maintained by #709

modelexpress_client/python/modelexpress_rl/inference/session.py
anson627: 2 test cases added

modelexpress_client/python/modelexpress_rl/inference/checkpoint_store.py
```diff
@@ -20,12 +20,21 @@ class CheckpointState(str, Enum):
 
     READY = "READY"
     UPDATING = "UPDATING"
+    UNSAFE = "UNSAFE"
```
nv-hwoo: I think the UNSAFE checkpoint state is a bit misleading because this actually happens when the engine installation fails, not when the checkpoint update fails. So the checkpoint itself is actually ready even though the installation might have failed. We probably need to define a separate engine state.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -267,10 +296,44 @@ def reset_initial_checkpoint(self) -> None:
...
+        """Prepare an ordered chain into one immutable target checkpoint."""
+        if not versions:
+            raise ValueError("canonical replay chain is empty")
         with self.store.installation_locked(), self.store.locked():
             state = self.store.state()
             if state is None:
                 raise RuntimeError("local checkpoint state is missing")
+            active_version = self.store.active_version()
+            target = versions[-1]
+            if state.get("status") == CheckpointState.UNSAFE:
```
nv-hwoo: Related to my comment above, in case of retry, it seems like the code will reject the latest checkpoint since it's set to UNSAFE, when in reality the checkpoint is actually ready but engine installation failed. 

For example, in the following case:
```
  materialized/v3 checkpoint: valid and READY
  active.json:                v2
  live engine:                possibly mixed v2/v3  # failed v3 install
```
The code will treat the v3 checkpoint state as UNSAFE

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -281,41 +344,60 @@ def prepare(self, version: _S3Version) -> PreparedCheckpoint:
...
+                        f"{version.version_id!r}"
+                    )
+                expected_base = version.version_id
+                started = time.perf_counter()
+                try:
+                    index_data = self.s3.get(version.uri)
+                    if version.payload_format is WeightPayloadFormat.XOR_DELTA:
+                        _parse_delta_manifest(index_data, version)
+                    else:
+                        _parse_full_checkpoint_manifest(index_data)
```
nv-hwoo: Should we remove these extra calls?
```suggestion
```

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -109,19 +109,48 @@ def _zstd_stream_reader(data: memoryview) -> Any:
 
 def _parse_delta_manifest(
     data: bytes,
+    version: _S3Version,
```
nv-hwoo, line 122: Do we need to pass version here? I don't think it's used for validation.

modelexpress_client/python/modelexpress_rl/inference/checkpoint_store.py
anson627: Agreed. UNSAFE conflates checkpoint integrity with engine health. The immutable checkpoint can remain READY, installation failure should instead mark the local engine state as uncertain while leaving active.json at the last successfully installed version. I’ll separate those states.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
anson627: Agreed. In this example, materialized/v3 remains valid and cacheable, while active.json stays at v2 and only the live engine is uncertain. Retrying v2 should bypass the no-op path and reinstall the cached v2 checkpoint, a successful reinstall clears the engine failure state

modelexpress_client/python/modelexpress_rl/inference/receiver.py
anson627: version is used to validate the manifest’s version and base_version metadata against the control-plane revision. That prevents a valid but incorrectly associated delta manifest from entering the chain, so I think we should keep it

modelexpress_client/python/modelexpress_rl/inference/receiver.py
anson627: The upfront calls ensure every manifest in the chain is validated before we write UPDATING or materialize the first revision. The later parsing is redundant though. I’ll retain upfront validation and pass the parsed results into preparation so each manifest is parsed once

modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -260,9 +278,40 @@ def stage_weight(self, *, version: WeightVersionRef) -> StagedWeightHandle:
                 if self._active_handle.version_id == version.version_id:
                     return self._active_handle
                 raise RuntimeError("another generator update is still active")
-            ready = self._get_ready_version(version.version_id)
             assert self._runtime is not None
-            update = self._runtime.session.stage(ready)
+            if (
+                self._engine_state is _EngineState.UNCERTAIN
+                and version.version_id != self._serving_version_id
+            ):
```
c2w-sea: So when engineState is uncertain, the caller must only request a new target version same as serving version?

Why do we want to have this restriction? It seems ok to me to have a unseen new target version to reset the state cleanly and recover from the failure

anson627: Good point. Canonical replay materializes a complete checkpoint from the last successfully active version, so recovery does not depend on the live engine contents. I removed the fence. The uncertain state now only bypasses the same-version no-op. I also added coverage for both recovery paths

c2w-sea (approved):

modelexpress_client/python/modelexpress_rl/inference/client.py
c2w-sea: I asked my agent to perform more testing, there may still be issues in the recovery scenarios.

chain 1: v1 (full)-> v2 -> v3
chain 2: v4 (full)-> v5

Assume the current state is that (1) v1 is serving (2) v2 apply failed. The failure is  transient (next v2 apply would pass)

Can the code ensure that
1. Caller can apply v3 and succeed (have v1+v2+v3 applied)?
2. Caller can apply v4 or v5 and get the state to v4 or v5 successfully

anson627: yes, I added a test case [here](https://github.com/ai-dynamo/modelexpress/pull/710/changes#diff-aee0f5248763fff1f06e20a928018a16b4eb8312d13e12c82792496caa4902b9R854) to cover the exact recovery states:
  - v3: rebuilds from active v1, applies cached v2 and then v3, and installs the final materialized v3 checkpoint once.
  - v4: installs the independent full checkpoint directly.
  - v5: materializes v4 full + v5 delta, then installs v5 once.

  In each case, active.json remains v1 until installation succeeds, then advances to the requested target

nv-hwoo (approved):

nv-hwoo:
/ok to test [3b4af98](https://github.com/ai-dynamo/modelexpress/pull/710/commits/3b4af98c9f96a15c43d0b0d49f5680b7f9417727)

PR #711, opened 2026-09-02
fix(vllm): refresh host attention scales after RDMA

AndyDai-nv:
## Summary

- refresh vLLM q/k/v host scale mirrors after RDMA overwrites the registered accelerator tensors
- synchronize the CPU mirrors used by host-scalar attention backends
- fail closed unless the complete q/k/v host-scale contract is valid and FlashInfer caches are still cold
- avoid re-running all of `process_weights_after_loading()`, because the received tensors already use the source's processed runtime representation

Tracking bug: [NVBug 6702877](https://nvbugspro.nvidia.com/bug/6702877)

## Root cause

vLLM stores each FP8 attention scale in more than one representation:

- registered accelerator buffers: `_q_scale`, `_k_scale`, `_v_scale`
- Python host scalars: `_q_scale_float`, `_k_scale_float`, `_v_scale_float`
- CPU mirrors for applicable backends: `_k_scale_cpu`, `_v_scale_cpu`

The source runs post-load processing after loading the real checkpoint, so these values agree. An RDMA target first runs post-load processing on dummy weights to expose the final runtime tensor layout. RDMA then overwrites the registered accelerator buffers, but Python floats and CPU tensors are not part of the transferred tensor manifest and retain dummy-derived values (usually `1.0`).

On Blackwell, vLLM's FlashInfer TRTLLM-Gen path passes the Python values as host scalars. The target therefore executes with real device scales but stale launch scalars, producing deterministic degenerate output despite a complete byte transfer.

## Relevant vLLM 0.26.0 source

These links are pinned to the vLLM `v0.26.0` commit (`568afb3a13806beb53bb2e6bd518269357b237c0`):

- vLLM creates both accelerator buffers and host mirrors: [`attention.py#L124-L145`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/layers/attention/attention.py#L124-L145)
- post-load processing copies checkpoint scales into the accelerator buffers, Python floats, and CPU mirrors: [`kv_cache.py#L128-L149`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/layers/quantization/kv_cache.py#L128-L149) and [`kv_cache.py#L182-L186`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/layers/quantization/kv_cache.py#L182-L186)
- architecture split: SM90/H200 supports XQA decode only, while SM100/SM103 enables TRTLLM attention for prefill and decode: [`utils/flashinfer.py#L374-L394`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/utils/flashinfer.py#L374-L394)
- SM100 specifically enables pre-quantized FP8 query input, while the comment documents the different SM90 behavior: [`flashinfer.py#L1587-L1603`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L1587-L1603)
- FlashInfer derives cached BMM launch scales from the Python host mirrors: [`flashinfer.py#L1708-L1716`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L1708-L1716)
- the TRTLLM prefill API receives q/k/v Python host scalars directly: [`flashinfer.py#L1925-L1933`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L1925-L1933)
- decode passes the derived host scalars to the TRTLLM kernel: [`flashinfer.py#L2193-L2224`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L2193-L2224)
- Blackwell also selects a FlashInfer FP8 linear kernel that explicitly requires compute capability 100+: [`scaled_mm/flashinfer.py#L39-L53`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/kernels/linear/scaled_mm/flashinfer.py#L39-L53)

The architecture-specific code does not corrupt RDMA bytes. It exposes the stale host state because the Blackwell execution path consumes those non-transferred Python values differently from the H200 path observed in the same test.

### Unit tests

```text
python -m pytest modelexpress_client/python/tests/test_vllm_adapter.py -q
36 passed in 0.09s
```

All staged pre-commit hooks passed.

### B200 end-to-end

Configuration:

- `nvidia/Llama-3.1-405B-Instruct-FP8`
- two B200 nodes, TP=4 on each node
- vLLM 0.26.0, FP8 KV cache, `arch=sm100`, `decode_backend=trtllm-gen`
- source loads locally; target receives over RDMA

All four target ranks transferred 2272 tensors / 102.55 GB. Each rank then logged:

```text
Refreshed 378 vLLM host attention scales across 126 modules after RDMA receive (245 stale dummy values replaced)
```

Before the fix, the target produced repeated/garbled text from the first token. With this change it produces coherent English and matches the source for the first 27 generated tokens in the captured 32-token request.

There is still a normal-English continuation difference after token 27, so the existing strict 32-token byte-equality assertion is not satisfied by this run. This PR fixes the catastrophic stale-scale corruption; it does not claim cross-node Blackwell inference is bit-exact.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vLLM attention quantization scale handling after RDMA loading.
  - Refreshed host-side scale values from accelerator-backed data for more consistent inference behavior.
  - Preserved probability scale values and registered buffer storage during post-load processing.
  - Cleared temporary implementation scale fields after processing to prevent stale state.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -41,6 +41,12 @@ _VLLM_POST_RDMA_FINALIZER_NAMES = (
     "finalize_mhc_broadcast_weights",
 )
 
+# vLLM keeps accelerator-backed attention scales together with host mirrors
+# consumed by backends such as FlashInfer and AITER. Only the accelerator
+# tensors are part of the ModelExpress manifest, so an RDMA target must refresh
+# the mirrors after the received values overwrite its dummy-loaded tensors.
+_VLLM_ATTENTION_SCALE_NAMES: tuple[str, ...] = ("q", "k", "v")
+
 # MTP draft weights live under an "mtp." prefix in the shared checkpoint. The
 # draft's embedding and lm_head come from the target, so these are all it needs.
 _DRAFT_WEIGHT_PREFIXES: tuple[str, ...] = ("mtp.",)
@@ -231,9 +237,10 @@ class VllmAdapter(EngineAdapter):
 
     def after_rdma_receive(self, result: LoadResult) -> LoadResult:
         """Build target-local tensors derived from the received weights."""
-        return self._finalize_model_specific_weights(
+        result = self._finalize_model_specific_weights(
             result, _VLLM_POST_RDMA_FINALIZER_NAMES
         )
+        return self._refresh_host_quantization_state(result)
 
     def apply_weight_iter(
         self,
@@ -387,6 +394,84 @@ class VllmAdapter(EngineAdapter):
                 result.model,
                 self.model_config,
                 self.target_device,
+            )
+        return result
+
+    @torch.no_grad()
+    def _refresh_host_quantization_state(self, result: LoadResult) -> LoadResult:
+        """Refresh vLLM host mirrors after RDMA overwrites device scales.
+
+        vLLM's attention PWAL copies ``_q/_k/_v_scale`` into Python floats and
+        CPU tensors for backends whose launch APIs require host scalars. RDMA
+        receives only the registered accelerator tensors, leaving those mirrors
+        at values derived from the target's dummy load. Re-running PWAL is not a
+        safe repair because the incoming model is already in its processed
+        runtime representation; update only the non-transferable mirrors.
+
+        Compressed-tensors may use per-head scales. Match vLLM's own conversion
+        by using their maximum for the scalar host mirror.
+        """
+        if result.model is None:
+            raise RuntimeError("vLLM RDMA post-load processing requires result.model")
+
+        refreshed_values = 0
+        stale_values = 0
+        refreshed_modules = 0
+
+        for module_name, module in result.model.named_modules():
+            module_refreshed = False
+            for scale_name in _VLLM_ATTENTION_SCALE_NAMES:
+                tensor_name = f"_{scale_name}_scale"
+                float_name = f"{tensor_name}_float"
+                scale = getattr(module, tensor_name, None)
+                if not isinstance(scale, torch.Tensor) or not hasattr(
+                    module, float_name
+                ):
+                    continue
+                if not self.accelerator_backend.is_accel_tensor(scale):
+                    continue
+                if scale.numel() == 0:
+                    continue
+
+                value = float(scale.detach().float().max().item())
+                previous = getattr(module, float_name)
+                if previous != value:
+                    stale_values += 1
+                    logger.debug(
+                        "Refreshing vLLM host scale %s.%s: %r -> %r",
+                        module_name or type(module).__name__,
+                        float_name,
+                        previous,
+                        value,
+                    )
+                setattr(module, float_name, value)
+
+                cpu_mirror = getattr(module, f"_{scale_name}_scale_cpu", None)
+                if isinstance(cpu_mirror, torch.Tensor):
+                    cpu_mirror.fill_(value)
+
+                refreshed_values += 1
+                module_refreshed = True
+
+            if not module_refreshed:
+                continue
+
+            refreshed_modules += 1
+            # FlashInfer lazily derives these from the host scale mirrors. They
+            # should still be None during cold loading, but invalidate them to
+            # keep this hook correct if initialization order changes upstream.
+            impl = getattr(module, "impl", None)
+            for cache_name in ("bmm1_scale", "bmm2_scale"):
+                if impl is not None and hasattr(impl, cache_name):
+                    setattr(impl, cache_name, None)
+
+        if refreshed_values:
+            logger.info(
+                "Refreshed %d vLLM host attention scales across %d modules "
+                "after RDMA receive (%d stale dummy values replaced)",
+                refreshed_values,
+                refreshed_modules,
+                stale_values,
             )
         return result
 
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -401,12 +401,13 @@ class VllmAdapter(EngineAdapter):
     def _refresh_host_quantization_state(self, result: LoadResult) -> LoadResult:
         """Refresh vLLM host mirrors after RDMA overwrites device scales.
 
-        vLLM's attention PWAL copies ``_q/_k/_v_scale`` into Python floats and
-        CPU tensors for backends whose launch APIs require host scalars. RDMA
-        receives only the registered accelerator tensors, leaving those mirrors
-        at values derived from the target's dummy load. Re-running PWAL is not a
-        safe repair because the incoming model is already in its processed
-        runtime representation; update only the non-transferable mirrors.
+        vLLM's attention PWAL copies ``_q/_k/_v_scale`` into Python floats and,
+        for some backends, CPU tensors whose launch APIs require host scalars.
+        RDMA receives only the registered accelerator tensors, leaving those
+        mirrors at values derived from the target's dummy load. Re-running PWAL
+        is not a safe repair because the incoming model is already in its
+        processed runtime representation; update only the non-transferable
+        mirrors that the selected backend exposes.
 
         Compressed-tensors may use per-head scales. Match vLLM's own conversion
         by using their maximum for the scalar host mirror.
@@ -422,7 +423,7 @@ class VllmAdapter(EngineAdapter):
         device_names = tuple(
             f"_{scale_name}_scale" for scale_name in _VLLM_ATTENTION_SCALE_NAMES
         )
-        required_names = device_names + float_names + cpu_names + ("_o_scale_float",)
+        required_names = device_names + float_names + ("_o_scale_float",)
         flashinfer_cache_names = ("bmm1_scale", "bmm2_scale", "o_sf_scale")
 
         cache_config = getattr(self.vllm_config, "cache_config", None)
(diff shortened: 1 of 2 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/711/commits/7b1b8cc279da019865c79fea127f4083eb0138f2

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
```diff
@@ -390,6 +397,84 @@ def _process_weights_after_loading(
...
+        for module_name, module in result.model.named_modules():
+            module_refreshed = False
+            for scale_name in _VLLM_ATTENTION_SCALE_NAMES:
+                tensor_name = f"_{scale_name}_scale"
+                float_name = f"{tensor_name}_float"
+                scale = getattr(module, tensor_name, None)
+                if not isinstance(scale, torch.Tensor) or not hasattr(
+                    module, float_name
+                ):
+                    continue
```
zhengluo-nv: Could we fail closed instead of silently skipping modules that expose part of this private scale contract? This hook exists because stale host mirrors produce valid-looking but incorrect output. If vLLM renames or removes an attribute, every scale can be skipped and the RDMA load still reports success. Once any q/k/v host mirror is present, please require the complete q/k/v device/host state and valid, nonempty, finite, positive accelerator tensors. Also raise when FP8 KV cache is expected but no module is refreshed. The resulting exception already becomes a mutated strategy failure, so reinitialization or fallback is safer than serving with stale metadata.

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
```diff
@@ -390,6 +397,84 @@ def _process_weights_after_loading(
...
+                continue
+
+            refreshed_modules += 1
+            # FlashInfer lazily derives these from the host scale mirrors. They
+            # should still be None during cold loading, but invalidate them to
+            # keep this hook correct if initialization order changes upstream.
+            impl = getattr(module, "impl", None)
+            for cache_name in ("bmm1_scale", "bmm2_scale"):
+                if impl is not None and hasattr(impl, cache_name):
+                    setattr(impl, cache_name, None)
```
zhengluo-nv: This invalidation is incomplete for a pre-warmed FlashInfer layer: output-scale state also lives in module._o_scale_float and impl.o_sf_scale. Resetting only bmm1/bmm2 can cause later initialization to skip the output-scale adjustment. For this short-term shim, please enforce the cold-load invariant instead: verify the full cache family is uninitialized and raise if it is not. That keeps the supported lifecycle explicit rather than partially supporting re-entry or an upstream ordering change.

modelexpress_client/python/tests/test_vllm_adapter.py
```diff
@@ -197,6 +197,39 @@ def test_after_rdma_receive_runs_derived_weight_finalizers(monkeypatch):
...
+        torch_device_type="cpu"
+    )
+    model = nn.Module()
+    model.attn = _AttentionWithStaleHostScales()
+    pointers = {
+        name: tensor.data_ptr()
+        for name, tensor in model.attn.named_buffers(recurse=False)
+    }
+
+    result = adapter.after_rdma_receive(LoadResult(value=model, model=model))
```
zhengluo-nv, line 215: Could this test make ordering and failure behavior explicit? Have finalize_mhc_broadcast_weights change one received q/k/v scale, then assert the host mirror sees the finalized value. Please also add focused cases for (1) FP8 expected but no recognizable scale module and (2) a preinitialized FlashInfer cache. The current assertions cover only the happy-path assignments, so the two lifecycle regressions above could pass unnoticed.

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
AndyDai-nv: Addressed in `135ac4cc`. The hook now validates every candidate module before mutating any host state. Once any q/k/v host mirror is present, it requires the complete q/k/v accelerator/Python/host-tensor contract plus `_o_scale_float`, and accelerator scales must be nonempty floating tensors containing only finite positive values. FP8 expectation is detected from the cache config and per-module `kv_cache_dtype`; an FP8 load with zero recognizable modules now raises so the RDMA strategy cannot report success.

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
AndyDai-nv: Addressed in `135ac4cc`. I removed the partial `bmm1_scale`/`bmm2_scale` invalidation. The shim now enforces the cold-load invariant before updating mirrors: `_o_scale_float`, `impl.bmm1_scale`, `impl.bmm2_scale`, and `impl.o_sf_scale` must all be uninitialized. A partial FlashInfer cache family or any initialized member raises and lets the strategy reinitialize or fall back.

modelexpress_client/python/tests/test_vllm_adapter.py
AndyDai-nv: Addressed in `135ac4cc`. Tests now verify that `finalize_mhc_broadcast_weights` changes `_q_scale` before host-mirror refresh, FP8 with no recognizable scale module fails, an incomplete module contract fails before any earlier valid module is mutated, invalid empty/NaN/non-positive accelerator scales fail, and each preinitialized cache member (`_o_scale_float`, `bmm1_scale`, `bmm2_scale`, `o_sf_scale`) fails. The focused adapter suite is now `36 passed`.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/711/commits/135ac4cc3cac4e304f224b755f1fc3ac624925b9

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
```diff
@@ -390,6 +397,185 @@ def _process_weights_after_loading(
...
+                        "Invalid vLLM attention CPU scale on "
+                        f"{module_label}: {cpu_name} must be a floating-point "
+                        "singleton tensor"
+                    )
+
+            impl = getattr(module, "impl", None)
+            cache_presence = [
+                hasattr(impl, cache_name) for cache_name in flashinfer_cache_names
+            ]
+            if any(cache_presence) and not all(cache_presence):
```
zhengluo-nv: [P1] This all-or-none cache contract rejects a valid vLLM 0.26 backend. `FlashInferMLAImpl` intentionally defines `bmm1_scale` and `bmm2_scale` but no `o_sf_scale` ([upstream source](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/mla/flashinfer_mla.py#L114-L166)). Its attention layer still has the shared q/k/v host mirrors, so an MLA RDMA target reaches this branch and raises `missing o_sf_scale`. Please remove the cache-family completeness requirement and only reject each known cache attribute when it exists and is non-`None`. This also gives a smaller implementation: validate and update in one pass; the outer RDMA strategy already marks any exception here as mutated and reinitializes before fallback.

zhengluo-nv (commented):
One compatibility issue remains in the defensive cache validation.

modelexpress_client/python/modelexpress/engines/vllm/adapter.py
AndyDai-nv: Addressed in `bf543a44`. You are right that `FlashInferMLAImpl` has `bmm1_scale` and `bmm2_scale` but intentionally lacks `o_sf_scale`. I removed the all-or-none cache-family requirement. Each known cache attribute is now checked independently: if it exists it must still be `None`; absence means that backend does not use that cache. I also changed the refresh to one pass, relying on the RDMA wrapper to convert any exception to `StrategyFailed(mutated=True)` and the strategy chain to reinitialize before fallback. The former incomplete-cache failure test is now an MLA-compatible success test. Focused suite: `36 passed`; hooks pass, and the same change is cherry-picked into #716.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/711/commits/bf543a443badfa1b7ce52917ffb7d379cabb767a

zhengluo-nv (approved):

AndyDai-nv:
CI follow-up: both failures came from the same compatibility regression in the hardening added for the latest review. The CI vLLM v0.17.1 attention implementation does not expose `_k_scale_cpu` / `_v_scale_cpu`, but commit `bf543a44` treated those backend-specific CPU mirrors as mandatory. This made RDMA fail closed and fall back to a local load. In rolling-update all five replacement pods logged the missing CPU fields; in the MLA test the target subsequently registered/published 433 locally loaded tensors, explaining the extra 29,648 MiB and the peak/final VRAM assertions. Commit `056498ff` keeps accelerator q/k/v scales and Python q/k/v host scalars mandatory, while validating and refreshing each CPU mirror only when that backend exposes it. Added a regression test for a backend with no CPU mirrors. Local verification: 37 targeted tests passed; full Python suite 1702 passed, 4 skipped; hooks on changed files passed. New CI is running.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/711/commits/056498ffb5264978d783a68cb117e1cba287aa53

PR #716, opened 2026-09-02
fix(vllm): refresh host attention scales after RDMA (cherry-pick #711)

AndyDai-nv:
## Backport

Cherry-pick of #711 onto `release/0.6.0`.

- source commit on `main` PR: `7b1b8cc279da019865c79fea127f4083eb0138f2`
- backport commit: `0b0d2a8881ca903ee9e485d4957142c61cee4dc8`

## Summary

- refresh vLLM q/k/v host scale mirrors after RDMA overwrites the registered accelerator tensors
- synchronize the CPU mirrors used by host-scalar attention backends
- fail closed unless the complete q/k/v host-scale contract is valid and FlashInfer caches are still cold
- avoid re-running all of `process_weights_after_loading()`, because the received tensors already use the source's processed runtime representation

Tracking bug: [NVBug 6702877](https://nvbugspro.nvidia.com/bug/6702877)

## Root cause

vLLM stores each FP8 attention scale in more than one representation:

- registered accelerator buffers: `_q_scale`, `_k_scale`, `_v_scale`
- Python host scalars: `_q_scale_float`, `_k_scale_float`, `_v_scale_float`
- CPU mirrors for applicable backends: `_k_scale_cpu`, `_v_scale_cpu`

The source runs post-load processing after loading the real checkpoint, so these values agree. An RDMA target first runs post-load processing on dummy weights to expose the final runtime tensor layout. RDMA then overwrites the registered accelerator buffers, but Python floats and CPU tensors are not part of the transferred tensor manifest and retain dummy-derived values (usually `1.0`).

On Blackwell, vLLM's FlashInfer TRTLLM-Gen path passes the Python values as host scalars. The target therefore executes with real device scales but stale launch scalars, producing deterministic degenerate output despite a complete byte transfer.

## Relevant vLLM 0.26.0 source

These links are pinned to the vLLM `v0.26.0` commit (`568afb3a13806beb53bb2e6bd518269357b237c0`):

- vLLM creates both accelerator buffers and host mirrors: [`attention.py#L124-L145`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/layers/attention/attention.py#L124-L145)
- post-load processing copies checkpoint scales into the accelerator buffers, Python floats, and CPU mirrors: [`kv_cache.py#L128-L149`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/layers/quantization/kv_cache.py#L128-L149) and [`kv_cache.py#L182-L186`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/layers/quantization/kv_cache.py#L182-L186)
- architecture split: SM90/H200 supports XQA decode only, while SM100/SM103 enables TRTLLM attention for prefill and decode: [`utils/flashinfer.py#L374-L394`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/utils/flashinfer.py#L374-L394)
- SM100 specifically enables pre-quantized FP8 query input, while the comment documents the different SM90 behavior: [`flashinfer.py#L1587-L1603`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L1587-L1603)
- FlashInfer derives cached BMM launch scales from the Python host mirrors: [`flashinfer.py#L1708-L1716`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L1708-L1716)
- the TRTLLM prefill API receives q/k/v Python host scalars directly: [`flashinfer.py#L1925-L1933`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L1925-L1933)
- decode passes the derived host scalars to the TRTLLM kernel: [`flashinfer.py#L2193-L2224`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/attention/backends/flashinfer.py#L2193-L2224)
- Blackwell also selects a FlashInfer FP8 linear kernel that explicitly requires compute capability 100+: [`scaled_mm/flashinfer.py#L39-L53`](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/model_executor/kernels/linear/scaled_mm/flashinfer.py#L39-L53)

The architecture-specific code does not corrupt RDMA bytes. It exposes the stale host state because the Blackwell execution path consumes those non-transferred Python values differently from the H200 path observed in the same test.

### Unit tests

```text
python -m pytest modelexpress_client/python/tests/test_vllm_adapter.py -q
36 passed in 0.09s
```

All pre-commit hooks passed on the release branch.

### B200 end-to-end

Configuration:

- `nvidia/Llama-3.1-405B-Instruct-FP8`
- two B200 nodes, TP=4 on each node
- vLLM 0.26.0, FP8 KV cache, `arch=sm100`, `decode_backend=trtllm-gen`
- source loads locally; target receives over RDMA

All four target ranks transferred 2272 tensors / 102.55 GB. Each rank then logged:

```text
Refreshed 378 vLLM host attention scales across 126 modules after RDMA receive (245 stale dummy values replaced)
```

Before the fix, the target produced repeated/garbled text from the first token. With this change it produces coherent English and matches the source for the first 27 generated tokens in the captured 32-token request.

There is still a normal-English continuation difference after token 27, so the existing strict 32-token byte-equality assertion is not satisfied by this run. This backport fixes the catastrophic stale-scale corruption; it does not claim cross-node Blackwell inference is bit-exact.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -41,6 +41,12 @@ _VLLM_POST_RDMA_FINALIZER_NAMES = (
     "finalize_mhc_broadcast_weights",
 )
 
+# vLLM keeps accelerator-backed attention scales together with host mirrors
+# consumed by backends such as FlashInfer and AITER. Only the accelerator
+# tensors are part of the ModelExpress manifest, so an RDMA target must refresh
+# the mirrors after the received values overwrite its dummy-loaded tensors.
+_VLLM_ATTENTION_SCALE_NAMES: tuple[str, ...] = ("q", "k", "v")
+
 # MTP draft weights live under an "mtp." prefix in the shared checkpoint. The
 # draft's embedding and lm_head come from the target, so these are all it needs.
 _DRAFT_WEIGHT_PREFIXES: tuple[str, ...] = ("mtp.",)
@@ -231,9 +237,10 @@ class VllmAdapter(EngineAdapter):
 
     def after_rdma_receive(self, result: LoadResult) -> LoadResult:
         """Build target-local tensors derived from the received weights."""
-        return self._finalize_model_specific_weights(
+        result = self._finalize_model_specific_weights(
             result, _VLLM_POST_RDMA_FINALIZER_NAMES
         )
+        return self._refresh_host_quantization_state(result)
 
     def apply_weight_iter(
         self,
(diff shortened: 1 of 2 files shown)


AndyDai-nv:
Synced the PR 711 CI compatibility follow-up via `cherry-pick -x` in `c81058f1`. vLLM backends that do not expose `_k_scale_cpu` / `_v_scale_cpu` now remain on the RDMA path; CPU mirrors are still type-checked and refreshed when present. Targeted tests: 37 passed; changed-file hooks passed.

Issue #714, opened 2026-09-02
RDMA weight loading is silently skipped when `model_loader_extra_config` is set (vLLM DummyModelLoader rejects it)


undergg:
## Summary

When `--load-format modelexpress` is combined with a non-empty
`--model-loader-extra-config`, the RDMA (P2P) strategy is silently skipped and MX
falls through to the S3 streamer — P2P transfer never happens even with a warm
peer available. It does **not** crash the engine; the pod comes up healthy, just
without P2P.

Cause: `prepare_rdma_target()` builds the empty-tensor target by copying the
active `LoadConfig` and setting `load_format = "dummy"`, but leaves
`model_loader_extra_config` in place. vLLM's `DummyModelLoader.__init__` raises on
any non-empty extra config, so the RDMA strategy fails and the chain moves on.

## Code pointers

`modelexpress_client/python/modelexpress/engines/vllm/adapter.py`,
`prepare_rdma_target()`:

```python
dummy_config = copy.copy(self.load_config)   # still carries model_loader_extra_config
dummy_config.load_format = "dummy"
DummyModelLoader(dummy_config).load_weights(result.model, self.model_config)  # ← raises in __init__
```

Where it raises — vLLM
[`dummy_loader.py#L27-L31`](https://github.com/vllm-project/vllm/blob/0ad5652a523d7bfe191a08b97ca45ead585b1897/vllm/model_executor/model_loader/dummy_loader.py#L27-L31):

```python
if load_config.model_loader_extra_config:
    raise ValueError(
        f"Model loader extra config is not supported for "
        f"load format {load_config.load_format}"
    )
```

The `ValueError` is caught in `rdma_strategy.py` (`_load_as_target`) and
converted to `StrategyFailed`, which the chain (`load_strategy/__init__.py`)
swallows and continues past — hence the silent skip rather than a crash.

## Reproduce

Serve any model with `--load-format modelexpress` + e.g.
`--model-loader-extra-config '{"distributed": true, "memory_limit": 10000000000}'`
(valid `runai_streamer` keys, per
[vLLM's docs](https://docs.vllm.ai/en/stable/models/extensions/runai_model_streamer/#sharded-model-loading)),
run ≥2 replicas so a peer source exists, and observe replica 2 log
`Strategy rdma failed, trying next: Model loader extra config is not supported for load format dummy`
then load from S3.

## Fix

Strip the extra config for the dummy target only — rebind rather than mutate,
since the streamer fallback still needs the original:

[...]

PR #718, opened 2026-09-02
fix: RDMA P2P silently disabled when model_loader_extra_config is set

undergg:
`prepare_rdma_target()` copies the source LoadConfig and switches it to dummy format, but leaves model_loader_extra_config
  attached.
  
  vLLM's DummyModelLoader **rejects** any extra config (Model loader extra config is not supported for load format dummy),
  so whenever the deployment passes streamer knobs like distributed/memory_limit, the RDMA strategy always fails and falls back
  to loading from origin — P2P never engages, silently.

  Fix: rebind the dummy config's extra config to {} (source config left intact, so the streamer fallback keeps its knobs).

  Tests: add coverage that exercises the real prepare_rdma_target — the existing suite only mocked it, which is why this slipped
  through.
  
  Closes issue #714 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RDMA target preparation to avoid carrying over incompatible model-loading options.
  * Preserved the original model configuration while applying RDMA-specific settings.

* **Tests**
  * Added coverage for RDMA target preparation, including configuration isolation and empty-option handling.

### DIFF
--- a/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
+++ b/modelexpress_client/python/modelexpress/engines/vllm/adapter.py
@@ -216,6 +216,12 @@ class VllmAdapter(EngineAdapter):
             dummy_config.load_format = "dummy"
         except AttributeError:
             object.__setattr__(dummy_config, "load_format", "dummy")
+        # DummyModelLoader rejects any model_loader_extra_config. The RDMA
+        # target only allocates empty tensors, so strip the streamer-oriented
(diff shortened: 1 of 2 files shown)


AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/718/commits/bcf8142dc050d0b58fe858704a79e3f48e78fce8

AndyDai-nv (approved):
lgtm

PR #719, opened 2026-09-02
fix(vllm): select MTP draft shards for the extra-layer convention (GLM-family)

### LINKED ISSUE #715: Draft/MTP shard selection misses GLM-family models — streams the whole base checkpoint
When loading an MTP draft with `load_format=modelexpress`, the pre-download shard selector identifies MTP tensors by a `mtp.` name prefix only. Models that keep MTP weights in an **extra decoder layer** (GLM-4.5/5.2/5.3) have no `mtp.` tensors, so the selector matches nothing and streams **all** shards of the base checkpoint — re-downloading the full model to extract one extra layer. DeepSeek uses the `mtp.` prefix, so it's unaffected.

Two valid MTP naming conventions exist, both supported by vLLM:

| Convention | Tensor names | Models |
|---|---|---|
| Prefix | `mtp.*` | DeepSeek |
| Extra layer | `model.layers.{N}`, `N >= num_hidden_layers` | GLM-4.5 / 5.2 / 5.3 |

**Root cause:** the selector (`modelexpress/engines/vllm/adapter.py`, `_select_draft_weight_files`) uses `model.safetensors.index.json` + a name prefix. But the index only maps tensor→shard; *which* layers are MTP comes from **`config.json`** (`num_hidden_layers`, `num_nextn_predict_layers`) — which is exactly what vLLM uses (`vllm/model_executor/models/utils.py`, `get_spec_layer_idx_from_weight_name`):

```python
def get_spec_layer_idx_from_weight_name(config, weight_name):
    if not (n := getattr(config, "num_nextn_predict_layers", 0)):
        return None
    base = config.num_hidden_layers
    for i in range(n):
        if weight_name.startswith((f"model.layers.{base + i}.", f"layers.{base + i}.")):
            return base + i
    return None
```

This function matches **only** the extra-layer form. GLM ships in that form and calls it directly (`glm4_moe_mtp.py`). DeepSeek's `mtp.*` weights are **renamed** to `model.layers.{base + i}.*` by model-specific code before this function sees them (e.g. `deepseek_v4/.../mtp.py`).

**Fix:** make the selector config-driven — read `config.json`, derive the MTP layer names `model.layers.{base}` … `{base + n - 1}`, and match tensors by **either** convention (`mtp.*` OR those indices) before streaming. Both are required: the `mtp.` prefix isn't derivable from `config.json`, and `get_spec_layer_idx_from_weight_name` doesn't cover it — so the prefix stays alongside the config-derived layer names.

undergg:
## Problem

  With `load_format=modelexpress` + speculative decoding, the draft (MTP) load streams the **entire base checkpoint** just to
  extract one draft layer. `_select_draft_weight_files` matches draft tensors by an `mtp.` name prefix only, so checkpoints that
  store the MTP head as an **extra decoder layer** match nothing → `NO_DRAFT_WEIGHTS` → fall back to streaming every shard.

  There is no single naming standard for MTP weights. Two conventions exist:

  | Convention | Tensor names | Models |
  |---|---|---|
  | Prefix | `mtp.*` | DeepSeek |
  | Extra layer | `model.layers.{N}`, `N >= num_hidden_layers` | GLM-4.5 / 5.x |

  The selector knew only the first, so GLM-family checkpoints re-stream the whole model.

  ## Fix

  Union the existing `mtp.` prefix with the extra-layer names derived from `config.json` (`num_hidden_layers` +
  `num_nextn_predict_layers`), mirroring vLLM's `get_spec_layer_idx_from_weight_name` (`model.layers.{base+i}.` /
  `layers.{base+i}.` / `model.language_model.layers.{base+i}.`). A tensor is a draft weight if it matches **either** convention;
  the matches are then resolved to shards via the existing index lookup.

  DeepSeek's `mtp.` path is unchanged — this only adds the extra-layer convention.

  ## Verification

  Checked against the public safetensors indexes:

  | Model | before | after |
  |---|---|---|
  | DeepSeek-V4-Pro (`mtp.0.*`) | 1 / 64 shards | 1 / 64 (unchanged) |
  | GLM-5.3 (`model.layers.78.*`) | 141 / 141 (all) | 3 / 141 |

  Unit tests cover both conventions plus fallback/edge cases.

  Closes #715

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model loading to recognize draft weights using both MTP-prefixed and configuration-derived layer names.
  * Correctly handles models with multiple draft layers, missing configuration details, unsupported shard indexes, and no matching draft weights.
  * Improved fallback diagnostics when draft weights cannot be found across supported storage paths.

* **Tests**
  * Expanded coverage for GLM-style and DeepSeek MTP models, including local and object-storage loading scenarios.


AndyDai-nv (commented):
Although it should work fine for some GLM models, this might not be safe to merge.

  _mtp_layer_prefixes() derives checkpoint tensor names from:

  hf_config = model_config.hf_config
  base = hf_config.num_hidden_layers
  n = hf_config.num_nextn_predict_layers

  However, for a speculative draft, this is the post-override draft config, not necessarily the original checkpoint config. In several vLLM MTP implementations, `num_hidden_layers` is deliberately rewritten to zero so that the draft module is constructed with local layer indices starting at zero. That value cannot be used as the physical layer index in the original checkpoint.

  There is no validation that the selected files contain a complete MTP module. Therefore, an accidental match on the ordinary model.layers.0.* tensors prevents the existing safe “stream all shards” fallback.

  ### Concrete counterexample: XiaomiMiMo/MiMo-7B-RL

  The official checkpoint config has:

  "num_hidden_layers": 36,
  "num_nextn_predict_layers": 1

  But current vLLM rewrites the MiMo draft config as follows:

  if hf_config.architectures[0] == "MiMoForCausalLM":
      hf_config.update(
          {
              "num_hidden_layers": 0,
              "n_predict": n_predict,
              "architectures": ["MiMoMTPModel"],
          }
      )

  Consequently, this PR generates:

  mtp.
  model.layers.0.
  layers.0.
  model.language_model.layers.0.

  The real HF index shows:

  model.layers.0.*      -> model-00001-of-00004.safetensors
  model.mtp_layers.0.*  -> model-00004-of-00004.safetensors

  Index:

  https://huggingface.co/XiaomiMiMo/MiMo-7B-RL/blob/6299b5a2c45daf0c429285c92b8e61a5bd011c0d/model.safetensors.index.json

  Therefore this PR selects shard 1, while all actual MTP tensors are in shard 4.

  This is likely a silent partial load rather than a clean failure. MiMo filters the incoming iterator to only MTP tensors, embeddings, and the LM head:

  if "mtp_layers" in name or "embed_tokens" in name or "lm_head" in name:
      yield name, weight

  Source:

  https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/model_executor/models/mimo_mtp.py#L213-L222

  As a result, ModelStreamer can report success with only the embedding initialized and the actual MTP block left unloaded.

  ### Concrete counterexample: XiaomiMiMo/MiMo-V2-Flash

  The same problem exists for MiMo-V2-Flash. vLLM explicitly sets:

  {
      "num_hidden_layers": 0,
      "num_nextn_predict_layers": 1,
      "architectures": ["MiMoV2MTPModel"],
  }

  Source:

  https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/config/speculative.py#L743-L758

  The official checkpoint index shows:

  model.layers.0.*     -> model_0.safetensors
  model.mtp.layers.0.* -> model_mtp.safetensors

  Index:

  https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/blob/1afd314a2406c282e0956375c34a676501c78649/model.safetensors.index.json

  This PR selects model_0.safetensors; the correct file is model_mtp.safetensors.

  The MiMo-V2 loader explicitly ignores everything except model.mtp, the shared embedding, and the LM head:

  if (
      "model.mtp" not in name
      and "model.embed_tokens" not in name
      and not name.startswith("lm_head")
  ):
      continue

  Source:

  https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/model_executor/models/mimo_v2_mtp.py#L270-L280

  model_0.safetensors contains none of those required MTP weights. The loader can return an empty loaded_params set without raising, and Modelexpress currently ignores that result. This
  leaves the local MTP draft uninitialized while treating the ModelStreamer strategy as successful.

  ### Similar collision: GLM-4.7-Flash

  The official GLM-4.7-Flash config has:

  "num_hidden_layers": 47,
  "num_nextn_predict_layers": 1

  but vLLM rewrites the draft num_hidden_layers to zero:

  https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/config/speculative.py#L770-L779

  The real checkpoint stores the MTP module under model.layers.47.*, while this PR selects model.layers.0.*.

  The index maps:

  model.layers.0.*  -> model-00001-of-00048.safetensors
  model.layers.47.* -> model-00001, model-00047, model-00048

  Therefore the selector returns only shard 1 and omits most of the actual MTP weights. Since the GLM MTP loader also uses the rewritten base value when classifying layer names, ordinary
  layer-0 attention/MLP weights may be loaded into the draft MTP block, while MTP-specific projection, normalization, shared-head, and expert weights remain missing.

  ### Downstream effect

  The draft load is intentionally excluded from P2P:

  ctx.p2p_enabled = not _is_speculative_draft(vllm_config, model_config)

  so this does not publish corrupted draft weights to other peers. However, every worker locally executes the same incorrect shard selection. The main model may load correctly over RDMA
  while the local MTP proposer is silently incomplete or loaded with incorrect base-layer weights.

  Expected symptoms include:

  - extremely low speculative acceptance rate;
  - decode throughput worse than running without MTP;
  - NaN/Inf logits when quantized weights or scales remain unloaded;
  - request-time failures even though model loading reported success.

undergg:
> Although it should work fine for some GLM models, this might not be safe to merge.
> 
> _mtp_layer_prefixes() derives checkpoint tensor names from:
> 
> hf_config = model_config.hf_config base = hf_config.num_hidden_layers n = hf_config.num_nextn_predict_layers
> 
> However, for a speculative draft, this is the post-override draft config, not necessarily the original checkpoint config. In several vLLM MTP implementations, `num_hidden_layers` is deliberately rewritten to zero so that the draft module is constructed with local layer indices starting at zero. That value cannot be used as the physical layer index in the original checkpoint.
> 
> There is no validation that the selected files contain a complete MTP module. Therefore, an accidental match on the ordinary model.layers.0.* tensors prevents the existing safe “stream all shards” fallback.
> 
> ### Concrete counterexample: XiaomiMiMo/MiMo-7B-RL
> The official checkpoint config has:
> 
> "num_hidden_layers": 36, "num_nextn_predict_layers": 1
> 
> But current vLLM rewrites the MiMo draft config as follows:
> 
> if hf_config.architectures[0] == "MiMoForCausalLM": hf_config.update( { "num_hidden_layers": 0, "n_predict": n_predict, "architectures": ["MiMoMTPModel"], } )
> 
> Consequently, this PR generates:
> 
> mtp. model.layers.0. layers.0. model.language_model.layers.0.
> 
> The real HF index shows:
> 
> model.layers.0.* -> model-00001-of-00004.safetensors model.mtp_layers.0.* -> model-00004-of-00004.safetensors
> 
> Index:
> 
> https://huggingface.co/XiaomiMiMo/MiMo-7B-RL/blob/6299b5a2c45daf0c429285c92b8e61a5bd011c0d/model.safetensors.index.json
> 
> Therefore this PR selects shard 1, while all actual MTP tensors are in shard 4.
> 
> This is likely a silent partial load rather than a clean failure. MiMo filters the incoming iterator to only MTP tensors, embeddings, and the LM head:
> 
> if "mtp_layers" in name or "embed_tokens" in name or "lm_head" in name: yield name, weight
> 
> Source:
> 
> https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/model_executor/models/mimo_mtp.py#L213-L222
> 
> As a result, ModelStreamer can report success with only the embedding initialized and the actual MTP block left unloaded.
> 
> ### Concrete counterexample: XiaomiMiMo/MiMo-V2-Flash
> The same problem exists for MiMo-V2-Flash. vLLM explicitly sets:
> 
> { "num_hidden_layers": 0, "num_nextn_predict_layers": 1, "architectures": ["MiMoV2MTPModel"], }
> 
> Source:
> 
> https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/config/speculative.py#L743-L758
> 
> The official checkpoint index shows:
> 
> model.layers.0.* -> model_0.safetensors model.mtp.layers.0.* -> model_mtp.safetensors
> 
> Index:
> 
> https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/blob/1afd314a2406c282e0956375c34a676501c78649/model.safetensors.index.json
> 
> This PR selects model_0.safetensors; the correct file is model_mtp.safetensors.
> 
> The MiMo-V2 loader explicitly ignores everything except model.mtp, the shared embedding, and the LM head:
> 
> if ( "model.mtp" not in name and "model.embed_tokens" not in name and not name.startswith("lm_head") ): continue
> 
> Source:
> 
> https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/model_executor/models/mimo_v2_mtp.py#L270-L280
> 
> model_0.safetensors contains none of those required MTP weights. The loader can return an empty loaded_params set without raising, and Modelexpress currently ignores that result. This leaves the local MTP draft uninitialized while treating the ModelStreamer strategy as successful.
> 
> ### Similar collision: GLM-4.7-Flash
> The official GLM-4.7-Flash config has:
> 
> "num_hidden_layers": 47, "num_nextn_predict_layers": 1
> 
> but vLLM rewrites the draft num_hidden_layers to zero:
> 
> https://github.com/vllm-project/vllm/blob/d410fc12f30b4224fca71a2faca75dcd4a5c3557/vllm/config/speculative.py#L770-L779
> 
> The real checkpoint stores the MTP module under model.layers.47._, while this PR selects model.layers.0._.
> 
> The index maps:
> 
> model.layers.0.* -> model-00001-of-00048.safetensors model.layers.47.* -> model-00001, model-00047, model-00048
> 
> Therefore the selector returns only shard 1 and omits most of the actual MTP weights. Since the GLM MTP loader also uses the rewritten base value when classifying layer names, ordinary layer-0 attention/MLP weights may be loaded into the draft MTP block, while MTP-specific projection, normalization, shared-head, and expert weights remain missing.
> 
> ### Downstream effect
> The draft load is intentionally excluded from P2P:
> 
> ctx.p2p_enabled = not _is_speculative_draft(vllm_config, model_config)
> 
> so this does not publish corrupted draft weights to other peers. However, every worker locally executes the same incorrect shard selection. The main model may load correctly over RDMA while the local MTP proposer is silently incomplete or loaded with incorrect base-layer weights.
> 
> Expected symptoms include:
> 
> * extremely low speculative acceptance rate;
> * decode throughput worse than running without MTP;
> * NaN/Inf logits when quantized weights or scales remain unloaded;
> * request-time failures even though model loading reported success.

@AndyDai-nv

You are correct. Fixed by deriving the extra-layer prefixes from the checkpoint's on-disk
config.json (pre-override), and only for the top-level GLM shape; any other shape returns no prefixes and falls back to streaming all shards.

I validated against the real public config.json + model.safetensors.index.json for 34 models, comparing the selector's output to the shards that actually hold MTP tensors. Result: 0 wrong/partial selections. Specifically for your counterexamples:
- MiMo-7B-RL and MiMo-V2-Flash: now NO_DRAFT_WEIGHTS → stream all shards (correct load, unoptimized), instead of mis-selecting shard 1.
- GLM-4.7-Flash (Glm4MoeLiteForCausalLM): reads base=47 from the on-disk config and selects model.layers.47.* correctly (3 shards), so it's actually optimized.
- GLM-4.5/4.6, DeepSeek-V3/R1/V3.1/V3.2, and Qwen3-Next also resolve correctly; ERNIE-4.5 and LongCat fall back safely.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/719/commits/15d791f51f906f6eb63834bdd91e21c2e6d533f7

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/719/commits/6f6faa3ce4939f2a571c50437553dc6db4d7ee3a

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/719/commits/617757763c26e0198b7b646906451fdf5a0470f7

nicolasnoble (approved):
lgtm. reading the on-disk config.json before the override is I think the right fix - it's what makes MiMo-7B-RL fall back to streaming all shards instead of picking shard 1, and falling back seems like the safe default for any shape the selector doesn't know.

AndyDai-nv (approved):

Issue #713, opened 2026-09-02
[Bug][refit] Full checkpoint validation rejects standard HF shards and leaves receiver unrecoverable


c2w-sea:
## Summary

`FULL_HF_CHECKPOINT` support added by #705 has two coupled failure modes:

1. The receiver requires every full-checkpoint safetensors shard to contain an MX-specific per-tensor Adler-32 map in `__metadata__`. Standard Hugging Face safetensors files may have no metadata or unrelated metadata such as `format`, so otherwise valid full checkpoints are rejected with `has invalid checksums`.
2. A preparation failure leaves the local checkpoint state as `UPDATING`, even when validation fails before any tensor bytes are modified. Every subsequent update then fails immediately with `local checkpoint update is incomplete` until the receiver is reinitialized.

This makes standard HF full checkpoints incompatible with `FULL_HF_CHECKPOINT` and allows a recoverable input-validation error to disable further updates on that worker.

## Affected behavior

In `modelexpress_rl/inference/receiver.py`:

- `prepare()` writes `UPDATING` before calling `_download_full_checkpoint()`.
- `_download_full_checkpoint()` rejects a shard when `__metadata__` is absent or its keys do not exactly equal the shard tensor names.
- `READY(target)` is written only after the update succeeds.
- There is no failure path that restores `READY(base)` when validation fails before mutation.
- A later `prepare()` rejects any state other than `READY`.

## Reproduction

1. Create a normal safetensors checkpoint with `safetensors.torch.save_file()` and a standard HF `model.safetensors.index.json`, without MX-specific per-tensor checksum metadata.
2. Register it as an S3-backed `FULL_HF_CHECKPOINT`.
3. Stage the version on a generator.
4. Observe:

   ```text
   full HF checkpoint shard '...' has invalid checksums
   ```

5. Register and stage a second valid version, including one with the expected checksum metadata.
6. Observe that it fails immediately:

   ```text
   local checkpoint update is incomplete
   ```

7. Reinitialize the receiver and retry the second version. It can then proceed.

## Expected behavior

- Per-tensor checksum metadata remains required for `XOR_DELTA`, where it validates the reconstructed target tensors.
- `FULL_HF_CHECKPOINT` accepts standard HF safetensors shards without MX-specific checksum metadata.
- If supported checksum metadata is present on a full shard, the receiver validates it.
- A download or validation failure before the first checkpoint mutation preserves or restores `READY(base)`.
[...]

PR #720, opened 2026-09-02
fix(refit): support optional full checkpoint checksums

nv-hwoo:
## Summary

1. Full checkpoint weight update during delta refit no longer requires checksum inside `__metadata__`. It is optional and only gets checked when it exists.
2. Added a specifications for our delta index json file formats stored in S3.
3. Small refactor around the delta/full checkpoint index json file parsing.

The PR addresses https://github.com/ai-dynamo/modelexpress/issues/713 part 1. The part 2 is currently addressed by #709.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-tensor Adler-32 checksum support for full checkpoints.
  * Added validation for checkpoint metadata, tensor presence, and corrupted tensor data.

* **Bug Fixes**
  * Failed checkpoint preparation, downloads, validation, and installation now preserve the active checkpoint and support retry.
  * Temporary checkpoints are discarded when preparation fails, preventing incomplete updates from becoming active.

* **Documentation**
  * Updated deployment and refit guides with checkpoint lifecycle, validation, recovery, and artifact metadata details.

### DIFF
--- a/docs/S3_DELTA_WEIGHT_REFIT.md
+++ b/docs/S3_DELTA_WEIGHT_REFIT.md
@@ -44,6 +44,7 @@ Environment variables used by the clients and vLLM engine:
 | `AWS_DEFAULT_REGION` | unset | S3 region when `region_name` is not supplied in client configuration. |
 | `MX_REFIT_DELTA_BUCKET_BYTES` | `536870912` (512 MiB) | Optional tensor-bucket size override for framework integrations. |
 | `MX_REFIT_DELTA_WORKERS` | `min(32, CPU count)` | CPU workers used to compute and apply XOR deltas. |
+| `MX_REFIT_CHECKSUM_FORMAT` | `adler32` | Checksum algorithm written by canonical S3 trainers. |
 | `MX_REFIT_FULL_CHECKPOINT_BATCH_BYTES` | `4294967296` (4 GiB) | Maximum tensor bytes grouped into one full-checkpoint safetensors object. |
 | `MX_S3_UPLOAD_WORKERS` | `8` | Maximum concurrent multipart uploads per trainer rank. |
 | `MX_S3_DOWNLOAD_WORKERS` | `16` | Generator download concurrency. |
@@ -396,7 +397,8 @@ Full checkpoints use a standard Hugging Face safetensors index:
 ```json
 {
   "metadata": {
-    "total_size": 8
+    "total_size": 8,
+    "checksum_format": "adler32"
   },
   "weight_map": {
     "model.layers.0.example.weight": "model-00001-of-00004.safetensors"
@@ -405,17 +407,19 @@ Full checkpoints use a standard Hugging Face safetensors index:
 ```
 
 The generator requires a non-empty `weight_map` covering exactly the local
-checkpoint tensors. The index `metadata` field is optional and ignored.
+checkpoint tensors. The index `metadata` field is optional. When it contains
+`checksum_format`, the only supported value is `adler32`.
 
 Each referenced shard contains native HF tensors. Safetensors `__metadata__`
-is optional and may contain arbitrary string-to-string entries. A namespaced
-entry adds optional integrity checking for one tensor:
+may contain arbitrary string-to-string entries. When the index declares
+`checksum_format="adler32"`, it must also contain a checksum keyed by tensor
+name for every referenced tensor in the shard:
 
 ```json
 {
   "__metadata__": {
     "format": "pt",
-    "mx.adler32:model.layers.0.example.weight": "12ab34cd"
+    "model.layers.0.example.weight": "12ab34cd"
   },
   "model.layers.0.example.weight": {
     "dtype": "F32",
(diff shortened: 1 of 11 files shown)


docs/S3_DELTA_WEIGHT_REFIT.md
```diff
@@ -329,6 +330,91 @@ seed checkpoint, write the cache, or validate the base version.
...
+Each referenced shard contains native HF tensors. Safetensors `__metadata__`
+is optional and may contain arbitrary string-to-string entries. A namespaced
+entry adds optional integrity checking for one tensor:
+
+```json
+{
+  "__metadata__": {
+    "format": "pt",
+    "mx.adler32:model.layers.0.example.weight": "12ab34cd"
+  },
```
c2w-sea, line 423: this __metadata__ has a different form from the __metadata__ in the XOR_DELTA above, shouldn't they be consistent?

zhengluo-nv (approved):

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -107,36 +107,82 @@ def _zstd_stream_reader(data: memoryview) -> Any:
...
+    ``mx.adler32:<tensor-name>`` supplies an optional per-tensor checksum.
+    """
+    _, weight_map = _parse_index(data, "full HF checkpoint")
+    if not weight_map:
         raise ValueError("full HF checkpoint index has no tensors")
-    for name, filename in weight_map.items():
-        if not isinstance(name, str) or not isinstance(filename, str):
-            raise ValueError("full HF checkpoint weight_map must contain strings")
-        if not filename or Path(filename).name != filename:
-            raise ValueError(f"invalid full HF checkpoint shard filename {filename!r}")
```
c2w-sea, line 178: these checks were dropped in the refactoring of using the common helper _parse_index, are they needed?

modelexpress_client/python/modelexpress_rl/utils.py
```diff
@@ -17,6 +17,8 @@
 import numpy as np
 
 
+ADLER32_METADATA_PREFIX = "mx.adler32:"
```
c2w-sea: is this full-weight-only special prefix intentional?

docs/S3_DELTA_WEIGHT_REFIT.md
nv-hwoo: These are optional metadata as described above:
> Safetensors `__metadata__` is optional and may contain arbitrary string-to-string entries. A namespaced entry adds optional integrity checking for one tensor:

So these can be omitted in FULL_HF_CHECKPOINT. For XOR_DELTA, it uses it for checksums, so it's required in that case.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
nv-hwoo: Yes, I dropped them since they seemed a bit unnecessary checks (str type checks, etc). I changed to more targeted checks for our required formats.

modelexpress_client/python/modelexpress_rl/utils.py
nv-hwoo: For now yes. This is to distinguish from other metadata fields.

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
```diff
@@ -332,7 +338,7 @@ def upload(
         ) -> tuple[dict[str, str], int]:
             filename, tensors = item
             checksums = {
-                name: adler32_checksum(
+                f"{ADLER32_METADATA_PREFIX}{name}": adler32_checksum(
```
zhengluo-nv: I think we should not assume ADLER32 is always applied

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
zhengluo-nv, line 340: It's better to model `tuple[str, dict[str, torch.Tensor]]` into a python class

docs/S3_DELTA_WEIGHT_REFIT.md
c2w-sea: I guess my question is why you need this prefix for full weight not for delta weight?

for delta, you require the `model.safetensors.index.json` metadata which gives you the checksum_format:
```
{
  "metadata": {
    "version": "v1",
    "base_version": "v0",
    "delta_encoding": "xor",
    "compression_format": "zstd",
    "checksum_format": "adler32"
  },
```

for full weight, could we do the same and drop the special prefix? it would be nice to keep full weight and delta contract consistent

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
nv-hwoo: Unrelated to this PR, let's address it in a separate PR

docs/S3_DELTA_WEIGHT_REFIT.md
nv-hwoo: Makes sense, I have dropped the prefix and matched the metadata format for both

modelexpress_client/python/modelexpress_rl/train/methods/canonical_delta.py
nv-hwoo: Added checksum factory to support other checksum algos in the future

modelexpress_client/python/modelexpress_rl/utils.py
nv-hwoo: Dropped the prefix

PR #721, opened 2026-09-03
fix(client): install weights into the snapshot root the engine resolved

scydas:
## Summary

Follows #592 and #656 — fixes the server-cache path installing weights under a cache root the engine does not read from.

Without shared storage a multi-node engine resolves the model on one node and loads the weights in an `EngineCore` process on another. That process receives `ModelConfig.model` as a snapshot path the frontend resolved against *its* cache root, and it loads from that path regardless of what this worker would default to. Two ways that went wrong:

- **The path does not exist on this node.** `_snapshot_path()` fell through to `ensure_metadata(repo_id, ModelConfig.revision)`, which installed under this worker's own root. The install succeeded and the native loader then opened the original path, which was still empty.
- **The path exists but the roots differ.** `install_weight_files()` built its `ModelSnapshotCache` from `resolve_cache_root()`, so `SnapshotPatch` rejected the snapshot on its containment check: `Snapshot resolves outside the cache root`.

`ServerCacheStrategy` now recovers the cache root and the commit from the standard `<root>/models--<org>--<name>/snapshots/<commit>` layout and passes both down, so metadata and weights land where the engine reads.

### Dispatch

| `ModelConfig.model` | Present locally | Action | Cache root passed down |
|---|---|---|---|
| standard snapshot layout | yes | use it | recovered from the path |
| standard snapshot layout | no | install the commit the directory names | recovered from the path |
| anything else | yes | use it (unchanged) | none |
| anything else | no | install `ModelConfig.revision` (unchanged) | none |

The commit comes from the directory name rather than `ModelConfig.revision`. `resolve_revision()` returns the revision unchanged for a local path, so by the time `EngineCore` sees it the field is usually unset; asking the server for its default revision would install a snapshot under a different directory than the one the engine is going to open.

`snapshot_location()` matches that layout structurally — lowercase 40-hex directory name, `snapshots/` parent, repo directory matching the resolved repo id — and touches no filesystem. Anything else returns None and keeps today's behaviour, including the repo root itself and `refs/` paths.

### Behaviour change

For a snapshot path missing on this node, an install that previously landed on the server's default revision now fails when the server cannot confirm the commit the directory names. A server predating #656 reports no revision at all and so fails this way.

Failing is the intent. Under no shared storage, nodes silently holding different commits of one model is harder to diagnose than a clean miss, and #656 already established "no confirmation, no install" as the client contract.

### Scope

`ensure_metadata()` takes the root as a keyword and keys its per-process memo on the resolved root, since one revision installed under two roots is two snapshots and only one of them is the one the caller reads from. Leaving the keyword unset preserves `resolve_cache_root()`'s existing order — explicit, then `MODEL_EXPRESS_CACHE_DIRECTORY`, then `HF_HUB_CACHE` — so deployments with a single cache root are unaffected, and the documented requirement that the two variables name the same path is unchanged.

The `snapshot_download` patch now forwards the caller's `cache_dir`. vLLM passes `load_config.download_dir` there (`weight_utils.py`), and installing somewhere else leaves the original call resolving against a root the snapshot is not in.

Not in scope: paths that do not follow the cache layout still fall back rather than being rejected outright. `SnapshotPatch` accepts a repo root as a snapshot directory, so `--model` or `VLLM_MODEL_REDIRECT_PATH` pointing at `models--org--repo/` writes weights where no loader reads them. That is a separate defect, unrelated to cache roots, and is left for its own change.

## Validation

- `uv run --no-sync pytest tests/test_hf_snapshot_prefetch_patch.py tests/test_model_prefetch.py tests/test_model_client.py tests/test_model_snapshot.py tests/test_server_cache_strategy.py`
  - `227 passed`
- Added coverage for: a snapshot path pinning the client to its own root over `MODEL_EXPRESS_CACHE_DIRECTORY`; a structurally valid path missing on this node installing under the path's root with the directory's commit; a server that will not confirm that commit producing `StrategyFailed(mutated=False)` with no client constructed; non-layout paths and bare repo ids keeping their existing dispatch and unset root; one revision under two roots counting as two installs while one root written two ways counts as one; the patch forwarding `cache_dir`; and `snapshot_location()` rejecting repo roots, `refs/` paths, uppercase and truncated SHAs, and a repo directory belonging to another model.

## Summary by CodeRabbit

- **Improvements**
  - Model metadata and weights are now installed and loaded from the same resolved Hugging Face snapshot and cache location.
  - Server-backed caching now supports deployments where model resolution and loading occur in different processes or nodes.
  - Prefetching correctly respects explicitly selected cache directories and avoids mixing snapshots across cache locations.
  - Snapshot paths are validated more reliably, with existing fallback behavior preserved for nonstandard paths.

- **Documentation**
  - Updated architecture and deployment guidance to describe consistent snapshot and cache-root handling.


AndyDai-nv (approved):

AndyDai-nv (commented):
Overall LGTM

One non-blocking test nit: in `test_prefetch_failure_keeps_original_semantics`, the patched production call now always supplies `cache_directory`, but the test double is still declared as `failing_ensure(repo_id, revision=None)`. As a result, the test passes by swallowing an unexpected-keyword `TypeError`; the intended `RuntimeError("server unreachable")` is never raised. Please change it to accept `*, cache_directory=None` (and ideally assert the warning contains `server unreachable`) so the test exercises the failure it claims to cover.

AndyDai-nv:
/ok to test https://github.com/ai-dynamo/modelexpress/pull/721/commits/d89d33de188da51db1eed4846730b8e47decbfa0

scydas:
> Overall LGTM
> 
> One non-blocking test nit: in `test_prefetch_failure_keeps_original_semantics`, the patched production call now always supplies `cache_directory`, but the test double is still declared as `failing_ensure(repo_id, revision=None)`. As a result, the test passes by swallowing an unexpected-keyword `TypeError`; the intended `RuntimeError("server unreachable")` is never raised. Please change it to accept `*, cache_directory=None` (and ideally assert the warning contains `server unreachable`) so the test exercises the failure it claims to cover.

Confirmed and fixed in ee5f086.

Reproduced it first — the warning read:

ModelExpress metadata prefetch failed for org/model:
failing_ensure() got an unexpected keyword argument 'cache_directory'

so the RuntimeError never fired, and since the same handler swallows both, the assertions could not tell a TypeError from the failure the test names.

The double now takes *, cache_directory=None and the test asserts the warning carries server unreachable. I checked the assertion actually bites: reverting the double to its old signature now fails the test instead of passing silently.

zhengluo-nv:
/ok to test [ee5f086](https://github.com/ai-dynamo/modelexpress/pull/721/commits/ee5f086b1080e0b1ee7656bd8a90c9afd3f317ac)

PR #722, opened 2026-09-03
feat(crd): generate manifests from Rust types

kevin85421:
## Summary

- generate the ModelMetadata and ModelCacheEntry CRDs from their Rust `CustomResource` definitions
- keep example, Helm, and CI CRD manifests generated and synchronized
- add CI checks that reject stale generated manifests

## Motivation

Almost all Kubernetes-related projects generate CRD OpenAPI schemas from source definitions rather than hand-maintaining the complete schema. The usual workflow treats typed source definitions and their schema annotations as the source of truth, generates the CRD manifests from them, and commits the generated YAML for installation and review. This avoids drift between the types used by the controller and the API schema enforced by Kubernetes.

[KubeRay follows this pattern](https://github.com/ray-project/kuberay/blob/master/ray-operator/DEVELOPMENT.md#consistency-check): its Go `*_types.go` API definitions are synchronized to CRD YAML under `ray-operator/config/crd/bases/` with [`make manifests`](https://github.com/ray-project/kuberay/blob/master/ray-operator/Makefile#L64-L65), and the generated CRDs are then synchronized to the Helm chart. This PR applies the same pattern using `kube`, `schemars`, and a small `crdgen` binary.

## Generated CRD differences

- replace the hand-written header, remove CRD metadata labels, and reorder YAML fields without changing the core API identity
- add generated schema titles and descriptions, `nullable` markers for optional fields, and defaults derived from the Rust/Serde definitions
- require the top-level `spec` and the non-optional fields under `artifactSource`
- add `int32`/`int64` formats for integer fields; unsigned Rust integers use non-negative `int64` schemas because OpenAPI has no `uint32` format
- preserve the group, version, kind, scope, short names, printer columns, enum values, and status subresources

## TODO

- Consider replacing the schema-only enum adapters (`BackendTypeSchema`, `WorkerLifecycleSchema`, `ConditionStatusSchema`, and `ConditionTypeSchema`) and their corresponding `String` fields with typed enums. Before making these changes, evaluate whether stricter deserialization and future enum evolution would constitute a breaking change to the CRD or its wire compatibility, including compatibility with existing objects and rolling upgrades.
- Review the OpenAPI representation of unsigned Rust integer fields such as `nodeRank`, `fileCount`, and `chunkCount`. Kubernetes supports `int32` and `int64` formats but not `uint32`, so these fields are currently represented as non-negative `int64` values. Consider adding the `u32::MAX` upper bound to keep CRD validation aligned with Rust deserialization, and evaluate whether tightening that validation would be a breaking CRD change.

## Testing

- `cargo test -p modelexpress-server k8s_types`
- `cargo clippy -q -p modelexpress-server --bin crdgen -- -D warnings`
- deterministic generator and manifest synchronization checks

## Summary by CodeRabbit

* **New Features**
  * Kubernetes custom resources now provide clearer status information, printer columns, defaults, and validation rules.
  * Resource schemas include improved descriptions, nullability, required fields, supported value constraints, and date-time validation.
  * Worker, artifact, provider, and lifecycle fields now expose more consistent defaults and status semantics.

* **Bug Fixes**
  * Added automated checks to detect differences between generated resource definitions and published manifests, helping prevent schema drift.


kevin85421:
Hi ModelExpress team,

This PR is not yet ready for review, but I’d like to discuss whether this approach makes sense before polishing it further. Manually maintaining CRD YAML files is uncommon. All Kubernetes-related projects I’m familiar with generate their CRD OpenAPI schemas automatically from code (for example, KubeRay’s [raycluster_types.go](https://github.com/ray-project/kuberay/blob/master/ray-operator/apis/ray/v1/raycluster_types.go)). This is a significant change, but it makes ModelExpress much more cloud-native-friendly.

If you are OK with the changes, I will polish the PR and make it ready for review. Thanks!

zhengluo-nv:
Sounds good. Please check the generated CRD is identical or backward compatible to the old CRD. Thanks

modelexpress_server/src/registry/k8s_types.rs
```diff
@@ -31,6 +38,7 @@ pub struct ModelCacheEntrySpec {
     pub model_name: String,
 
     /// Provider string — `"HuggingFace"`, `"Ngc"`, `"Gcs"`, or `"S3"`.
+    #[schemars(with = "ProviderSchema")]
     pub provider: String,
```
kevin85421, line 42: Follow-up: Use `ProviderSchema` instead of String as the type of provider. This may be a breaking change to the CRD, so it should be addressed in a follow-up PR.

modelexpress_server/src/p2p/k8s_types.rs
```diff
@@ -146,18 +155,22 @@ pub struct ArtifactSourceStatus {
...
     pub file_count: u32,
 
     /// Number of transfer chunks in the sealed artifact manifest.
     #[serde(rename = "chunkCount")]
+    #[schemars(with = "i64", range(min = 0))]
     pub chunk_count: u32,
 
     /// Distributed node rank that owns this node-scoped artifact.
     #[serde(rename = "nodeRank", default)]
+    #[schemars(with = "i64", range(min = 0))]
```
kevin85421, line 184: Follow-up: OpenAPI does not support u32 as an integer format, so use int64 for now. A follow up is to change Rust type or setting OpenAPI integer range.

kevin85421:
Hi @yixinh-nv @zhengluo-nv,

Would you mind enabling CI for this PR so that I can verify whether the OpenAPI schema changes introduce any breaking changes? Thanks.

kevin85421:
All CI checks for https://github.com/ai-dynamo/modelexpress/commit/de1b6d4629f3ec4e721430ed3d5bb2fdd24f1278 passed. I pushed another commit to resolve the merge conflicts. This PR should be ready for review. Thanks!

yixinh-nv:
can you resolve the coderabbit comment?

examples/crds.yaml
```diff
@@ -1,285 +1,320 @@
...
+                        description: Total artifact bytes across all manifest files. Kubernetes OpenAPI exposes this as int64; the backend converts to/from the proto uint64.
+                        format: int64
+                        minimum: 0.0
+                        type: integer
+                    required:
+                    - artifactId
+                    - chunkCount
+                    - fileCount
+                    - totalSize
+                    type: object
```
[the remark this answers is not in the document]
yixinh-nv: This is a false positive. `ArtifactSourceStatus` on `main` already has no serde defaults on `artifactId`, `totalSize`, `fileCount`, and `chunkCount`, so a partial `artifactSource` object already failed to deserialize in `api.get` before this PR. The new `required` block only adds write-time validation matching what the Rust type has always required, and the only producer (`KubernetesBackend`) always writes all four fields. Existing objects are not re-validated when the CRD is updated, so no migration is needed. Resolving.

yixinh-nv (approved):

kevin85421:
Thank @yixinh-nv for the review!

zdtsw:
I guess after this change, it’s worth calling out (maybe in next release notes) that anyone who already has `ModelMetadata` CRs with a partial `artifactSource` (missing any of artifactId / totalSize / fileCount / chunkCount) needs to update or remove those objects before upgrading to the new CRD as we keep the same api version the schema check could break until they are fixed.

kevin85421:
Hi @zdtsw,

Thank you for the comment. I was wondering whether you foresee any cases where a partial `artifactSource` could occur. In Rust, none of the fields under `artifactSource` are optional, so if any field is missing, deserialization will fail, and the MX server will report an error such as:

Screenshot 2026-09-09 at 12 09 39 AM

In the screenshot above, I manually created a CR with a partial `artifactSource` for testing purposes only. Users should not write the CR status manually. The controller (the MX server) should be the only writer of the CR status, and I would be very surprised if the MX server could produce a partial `artifactSource`.

PR #727, opened 2026-09-04
feat(refit): add RL cold-start load strategy chain

zhengluo-nv:
## Summary

- add an engine-agnostic `MX_LOAD_STRATEGY_CHAIN` selector and an RL-specific cold-start policy
- when a desired version is configured, load it through exact-version P2P or S3 full-checkpoint plus delta-chain replay and fail startup if neither succeeds
- when no desired version is configured, fall back through `MX_MODEL_URI` and then the engine default loader
- reuse the active-refit checkpoint receiver and a shared immutable-version chain resolver for S3 replay
- update the Dynamo vLLM refit example to read the desired version from a ConfigMap

## Motivation

An RL generator created during scale-up or replacement must not silently restart from an older base model. When `MX_REFIT_DESIRED_VERSION_UID` is set, this change first tries to load that exact immutable version from a live peer and then reconstructs it from object storage. If neither exact-version path succeeds, startup fails. The existing model URI and engine loader remain fallbacks only when no desired version is configured.

## Load order

| Priority | Strategy | Availability |
| --- | --- | --- |
| 1 | desired-version P2P | desired version is set and RDMA loading is available |
| 2 | desired-version S3 replay | desired version and checkpoint directory are set, and the engine supports an explicit checkpoint URI |
| 3 | ModelStreamer | desired version is unset and `MX_MODEL_URI` loading is available |
| 4 | engine default | desired version is unset |

A configured desired UID is a correctness constraint: ModelStreamer and the engine default loader are not attempted if both exact-version strategies fail. The S3 strategy resolves a READY `FULL_HF_CHECKPOINT -> XOR_DELTA*` chain, materializes it through the existing canonical delta receiver, and passes the resulting local checkpoint to the engine loader. Chain validation is shared with active refit; active refit continues to reuse its existing control stub, while cold start uses a short-lived control client.

## Validation

- `UV_CACHE_DIR=/tmp/mx-uv-cache uv run pytest -q tests/test_envs.py tests/test_refit_envs.py tests/test_refit_load_strategy.py tests/test_refit_s3_receiver.py tests/test_sglang_loader.py tests/test_trtllm_loader.py tests/test_vllm_loader.py`
  - 256 passed, 8 skipped
- `pre-commit run --from-ref origin/main --to-ref HEAD`
- `git diff --check origin/main...HEAD`

## E2E status

Tested on nscale at `8492c91738fa4bb61fda866c3e7e899bcfed1bf3` with Qwen2.5-0.5B, vLLM `0.26.1rc1.dev1177+ga9a17e709`, ModelExpress `0.5.1`, TP1, and two B200 nodes.

| Path | Result | Evidence |
| --- | --- | --- |
| desired-version P2P miss -> S3 full checkpoint plus delta replay | load and inference passed | Published a full HF root plus a non-empty XOR delta target. The worker logged the P2P miss, selected `desired_version_s3`, materialized the desired version, reached API readiness, and served inference. |
| desired-version P2P | load and inference passed | The target had no disk weights and no S3/checkpoint configuration. It discovered the source on another node and transferred 267 tensors / 0.99 GB in 0.048 s at approximately 164.3 Gbps, reached API readiness, and served the same deterministic completion. |

Known gaps found by E2E and review:

- Both cold-start paths loaded the desired version, but vLLM `/weight_info` still returned `default`. The loader updates `LoadContext.identity.revision`, while vLLM initializes a separate `EngineCore._weight_version`. Calling `/update_weight_version` with the desired UID corrected the report. Cold-start completion must synchronize this state before routing can rely on it.
- The subsequent active-refit client also initializes its internal serving version from the configured initial base rather than the desired version installed during cold start. A cold-start-then-active-refit E2E is still required after that state is synchronized.
- This run used direct vLLM TP1. Dynamo routing/admission and a multinode vLLM worker were not exercised.
- The test image required installing the declared `zstandard` dependency at startup because the thin development image installs ModelExpress with `--no-deps`.

All disposable nscale resources and test artifacts were removed after the run.

## Summary by CodeRabbit

- **New Features**
  - Added configurable inference and reinforcement-learning model loading strategies.
  - Added support for loading a desired refit version through peer-to-peer transfer or object storage.
  - Added checkpoint caching and reuse for faster RL cold starts.
  - Added configurable desired-version and checkpoint-directory settings.
  - Added validation for replay chains, checkpoints, versions, and storage metadata.

- **Bug Fixes**
  - Improved loading fallbacks, retry handling, and recovery when a loading strategy fails.
  - vLLM now avoids unnecessary pre-downloads during RL initialization.


modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @staticmethod
+    def run(model: nn.Module, ctx: LoadContext) -> nn.Module:
+        return execute_load_strategies(
+            model,
+            ctx,
+            [
+                DesiredVersionP2PStrategy(),
+                DesiredVersionS3Strategy(),
+                ModelStreamerStrategy(),
```
anson627, line 235: The current chain continues to ModelStreamerStrategy and DefaultStrategy after both exact-version strategies fail, those loaders do not guarantee `MX_REFIT_DESIRED_VERSION_UID`, their success marks the worker ready and publishes it, so a replacement can therefore silently serve the base model. Should exhaustion of P2P/S3 must be terminal when a desired UID is set?

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,202 @@
...
+                        payload_format=root.payload_format,
+                        uri=root.object_storage.uri,
+                    ),
+                    refit_checkpoint_dir=checkpoint_dir,
+                    s3=s3,
+                )
+            finally:
+                s3.close()
+
+            method = CanonicalDeltaUpdateMethod(
```
anson627, line 144: Should installed desired UID and prepared cache state initialize the refit client’s serving state? The S3 strategy records the target only in the temporary boot-time ctx.identity, then closes its checkpoint method. Later, ModelExpressGeneratorClient still initializes _serving_version_id from runtime.initial_version_id, not the desired revision actually loaded into the engine. The engine and reported active version can therefore disagree immediately after cold start

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Make sense. In that case the worker will fail, is that ok?

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -179,6 +179,188 @@ def _group_tensors_by_shard(
...
+    target = store.full_path(version.version_id)
+    source = _source_identity(version)
+    with store.installation_locked(), store.locked():
+        reusable = target.exists()
+        if reusable:
+            try:
+                store.verify_artifact_source(target, source)
+                checkpoint_paths, _, _ = index_checkpoint_tensors(target)
+            except (OSError, RuntimeError, ValueError):
+                reusable = False
```
nv-hwoo: Quick question: What happens if we want v23 when the current checkpoint has full v20 and deltas v21, v22, and v23? Will this try to fetch the full v20 only?

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+    bootstrap_s3_checkpoint,
+)
+from .version_chain import resolve_replay_chain
+
+_MAX_REPLAY_CHAIN_LENGTH = 64
+logger = logging.getLogger(__name__)
+
+
+class DesiredVersionP2PStrategy(RdmaStrategy):
+    """Load the desired immutable version from an existing generator."""
```
nv-hwoo, line 65: Since RDMA candidates are rank-specific, can there be some risk of one rank loading desired N, while another falls through to base B?
```
  rank 0 -> desired-N
  rank 1 -> base-B
  both -> success
```
Maybe we need some kind of consensus on the target revision before returning.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+                        exc_info=True,
+                    )
+
+
+def _resolve_s3_replay_chain(
+    ctx: LoadContext,
+    target_version_uid: str,
+) -> tuple[WeightVersion, ...]:
+    """Resolve and validate a target back to its full HF root."""
+    with ModelExpressControlClient.connect(server_url=ctx.mx_server_url) as client:
```
ajcasagrande, line 200: [HIGH] **CRITICAL — desired-version S3 path never verifies the returned UID matches the requested one.**

`client.get_weight_version(version_uid)` is fetched here and validated for state, model and storage, but never for `version.version_id == version_uid`. `version_chain.py:43-68` then accepts the returned object and can terminate on its payload, and `load_strategy.py:123-126` loads that checkpoint and unconditionally sets `ctx.identity.revision = desired_version_uid`.

The established live-refit path does enforce exactly this check — `modelexpress_rl/inference/client.py:410-415`. Without it, a wrong, stale, or compromised control-plane response for desired version A that carries version B results in **B being served and published under A's UID**, which defeats the fail-closed guarantee this chain exists to provide.

Suggested fix: mirror the `_fetch_ready_version` assertion and raise before the checkpoint is materialized.

*Corroborated 7/8, zero refutations. A second member filed the same missing check independently.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class DesiredVersionP2PStrategy(RdmaStrategy):
+    """Load the desired immutable version from an existing generator."""
+
+    name = "desired_version_p2p"
+
+    def is_available(self, ctx: LoadContext) -> bool:
+        return envs.MX_REFIT_DESIRED_VERSION_UID is not None and super().is_available(ctx)
+
```
ajcasagrande, line 71: [HIGH] **CRITICAL — desired-version P2P loading bypasses the weight-version control plane entirely.**

This path only overwrites `ctx.identity.revision` and delegates straight to `RdmaStrategy.load`; it never calls `GetWeightVersion`, and never checks `WeightVersionState` or model ownership. `modelexpress/load_strategy/rdma_strategy.py:273-289` discovers only generic P2P source instances, and server-side `modelexpress_server/src/p2p/service.rs:200-225,239-263` filters on generic source identity/status/heartbeat with no coupling to the refit WeightVersion lifecycle.

Because the sibling S3 path *does* perform READY/model checks (`load_strategy.py:158-170`), the two branches enforce different policies: a fresh P2P metadata record can satisfy a desired UID even when that version is absent, non-READY, or in an invalid lifecycle state. That lets stale or adversarially published weights in through the P2P door while the S3 door is locked.

*Corroborated 7/8, zero refutations.*

examples/rl/dynamo_vllm_refit/dgd.yaml
```diff
@@ -36,6 +45,8 @@ spec:
           emptyDir:
             medium: Memory
             sizeLimit: 16Gi
+        - name: mx-refit-checkpoints
```
ajcasagrande, line 48: [MEDIUM] `mx-refit-checkpoints` is an `emptyDir` with no `sizeLimit`, while the checkpoint store defaults to a 500 GB quota.

The sibling `dshm` volume at `dgd.yaml:44-47` sets `sizeLimit: 16Gi`; this one sets nothing. It is mounted at `/var/cache/modelexpress/refit` (`dgd.yaml:101-102`) and wired to `MX_REFIT_CHECKPOINT_DIR` (`dgd.yaml:81-82`), and `receiver.py:39` defines `DEFAULT_REFIT_CHECKPOINT_MAX_SIZE_GB = 500`, used as the default in `bootstrap_s3_checkpoint` (`receiver.py:293`) and `ObjectStorageGeneratorConfig` (`receiver.py:50`). The manifest sets no override and the resources block (`dgd.yaml:90-97`) declares no `ephemeral-storage` limit — so **the kubelet will evict the pod for ephemeral storage long before `LocalCheckpointStore` ever trims the cache**.

Separately worth considering: `emptyDir` is wiped on pod restart, so this cache cannot survive the restarts it exists to accelerate. A PVC may be the better fit.

*Corroborated 7/8, zero refutations.*

modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -131,7 +131,7 @@ def load_model(
                             )
 
                     with metrics.time_load_phase("vllm", model_id, "chain"):
```
ajcasagrande, line 139: [MEDIUM] The process-global RL desired-version chain is also applied to speculative draft models.

`loader.py:96-113` identifies speculative drafts and only disables P2P for them, but this line still dispatches **every** model through `run_load_strategy_chain`. `modelexpress_rl/inference/load_strategy.py:187-191` then selects the desired P2P/S3 strategies purely from the process-global desired UID. With P2P disabled for the draft, S3 resolves that *main-model* UID and rejects its model name against the draft at `load_strategy.py:158-165` — and there is no native fallback after that.

Net effect: RL deployments using speculative decoding cannot cold-start once a desired UID is configured. The draft model likely needs to opt out of the desired-version chain the same way it opts out of P2P.

*Corroborated 7/8, zero refutations.*

modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -157,6 +157,13 @@ def load_model(
```
ajcasagrande, line 163: [HIGH] `download_model()` returns early for `MX_LOAD_STRATEGY_CHAIN=RL`, but the RL chain's unset-UID fallback may still expect pre-downloaded weights.

When `MX_REFIT_DESIRED_VERSION_UID` is unset, `RLLoadStrategyChain` falls back to `ModelStreamerStrategy` + `DefaultStrategy` (`load_strategy.py:188-197`) — both version-agnostic. This early return also skips the `model_prefetch.is_enabled()` server-cache path at `loader.py:167-176`.

Note the dissent, which the council found credible: `DefaultStrategy` is defined as engine-native disk *or hub* loading (`default_strategy.py:3,16-30`) and `adapter.py:344-357` invokes vLLM's `DefaultModelLoader(...).load_weights(...)`, so this is not a guaranteed missing-weight failure — the hub path remains. The residual concern is the lost prefetch/cache benefit and an unclear failure mode if hub access is unavailable.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+            max_chain_length=_MAX_REPLAY_CHAIN_LENGTH,
+        )
+
+
+class RLLoadStrategyChain:
+    """Load using the RL-specific cold-start fallback order."""
+
+    @staticmethod
+    def run(model: nn.Module, ctx: LoadContext) -> nn.Module:
+        # A desired UID is a correctness constraint: version-agnostic fallbacks
```
ajcasagrande, line 221: [HIGH] The unset-UID branch silently loads version-agnostic weights on exactly the boot path the fail-closed guarantee is meant to cover.

`if envs.MX_REFIT_DESIRED_VERSION_UID is not None: [DesiredVersionP2P, DesiredVersionS3] else: [ModelStreamer, Default]`. The example ConfigMap ships `desired_version_uid: ""` (`dgd.yaml:9-12`) and `envs.py:60-62` maps empty string to `None` — so a worker restarted **before the orchestrator has written the desired version** lands on the else branch and serves whatever ModelStreamer/Default can load, with no error.

`modelexpress_client/python/README.md:258` documents that startup fails rather than using a version-agnostic fallback when a desired version is configured; the unset case is equally reachable at boot and does the opposite.

The dissent argues this is intended: the guarantee is explicitly conditional on a configured UID, and empty means "no desired version yet" (`dgd.yaml:3-4`). That reading is defensible — but the race between pod start and orchestrator write is real, and at minimum deserves an explicit log line so a stale-weight boot is not silent.

*Corroborated 6/8, 1 refutation. A related LOW finding on the same branch drew 5/2.*

ajcasagrande (commented):
## 🏛️ Review Council — corroboration mode

Eight reviewers (Coda, Jixa, Garo, Pyra, Zeta, Nyra, Loxo, Navo) reviewed this PR independently, then each corroborated or refuted every other member's findings. **25 findings** were submitted; those below reached net-positive support. Every addressable finding is posted as an inline comment on its exact line.

The core design here — a dedicated RL chain that fails closed on a pinned version UID — is a genuinely good idea, and the test additions around it are welcome. The findings concentrate on two seams: **the two desired-version strategies enforce different policies**, and **the unset-UID boot window behaves opposite to the documented guarantee**.

### Blocking / CI-failing

Two HIGH findings reached unanimous corroboration (7/7, zero refutations). Both are correctness regressions in the new fail-closed path and are worth resolving before merge:

1. **`modelexpress_rl/inference/load_strategy.py:158` — the S3 desired-version path never verifies the returned UID matches the requested one.** It can load version B's bytes and publish them as desired version A. The live-refit path already enforces this check at `client.py:410-415`; the new path does not.
2. **`modelexpress_rl/inference/load_strategy.py:43` — desired-version P2P loading bypasses the weight-version control plane entirely.** No `GetWeightVersion`, no `WeightVersionState`, no model-ownership check — so a fresh P2P metadata record can satisfy a desired UID for a version that is absent, non-READY, or in an invalid lifecycle state. The S3 sibling *does* perform these checks, so the two branches disagree on policy.

Taken together these are the same underlying gap: **the desired-version contract is enforced in one branch and not the other.** Fixing them symmetrically is probably one change, not two.

### Suggested fix order

1. Add the `version.version_id == version_uid` assertion to the S3 path (small, high value).
2. Gate the P2P path through the same control-plane validation so both branches share one policy.
3. Decide and document the unset-UID boot behaviour (`load_strategy.py:186`) — either fail closed or log loudly; today it is silent.
4. Exempt speculative draft models from the process-global desired-version chain (`vllm/loader.py:133`).
5. Set a `sizeLimit` on the `mx-refit-checkpoints` `emptyDir`, or move it to a PVC (`dgd.yaml:48`).

### Documentation (no addressable diff line)

`AGENTS.md:200-210` requires documentation updates for architecture, configuration and CLI changes, and states that *"a feature is incomplete until documentation is updated."* The diff touches only `modelexpress_client/python/README.md` (plus `helm/README.md:4`):

- **[MEDIUM]** No updates to `docs/ARCHITECTURE.md`, `docs/DEPLOYMENT.md`, `docs/CLI.md` or `CONTRIBUTING.md` for the new RL cold-start loading, the desired-version env vars, or the ConfigMap pattern. *(7/7, zero refutations)*
- **[LOW]** The three new env vars — `MX_LOAD_STRATEGY_CHAIN`, `MX_REFIT_DESIRED_VERSION_UID`, `MX_REFIT_CHECKPOINT_DIR` (`README.md:257-259`) — are absent from `docs/DEPLOYMENT.md`, their mandated home, which already carries env tables at lines 612/642/1040 with 95 `MX_` references. Two existing statements are now stale: `docs/ARCHITECTURE.md:1094` still says new strategies are added by registering in `LoadStrategyChain.run()`, which no longer covers the RL chain; and `docs/ARCHITECTURE.md:904-914` lists the `modelexpress_rl/inference` modules but omits the new `load_strategy.py` and `version_chain.py`. *(7/7, zero refutations)*

<details>
<summary><b>Contested findings</b> (mixed or negative verdict — recorded, not actioned)</summary>

**[HIGH] `receiver.py:785-788` — `_prepare_delta` renames the base checkpoint before delta validation.** `base_checkpoint.replace(target)` runs prior to `_apply_delta_artifact`; if that raises, the base directory is gone and `target` is left partially modified, with no rollback in `prepare_chain`. *Corroborated 3 / Refuted 3.* **Not a regression in this PR** — the diff contains no `_prepare_delta` or `base_checkpoint.replace(target)` change, and blame attributes the logic to 7c3d511a/01aad490. Suggest tracking as a separate issue; the cache-corruption risk looks real on its own merits.

**[HIGH] `docker/Dockerfile` — `USER 1000:1000` removed while `runAsNonRoot: true` remains.** *Corroborated 3 / Refuted 4.* **Refuted: this PR does not touch `docker/Dockerfile` or `helm/values.yaml`** — `git diff --name-only origin/main...HEAD` lists only the RL example and client Python files.

**[LOW] `helm/values-local-storage.yaml:29-32` — `runAsNonRoot: false` for local testing.** *Corroborated 3 / Refuted 4.* **Refuted: file unchanged by this PR.**

**[LOW] `version_chain.py:46-52` — chain-mismatch wrapper masks the underlying fetch exception.** *Corroborated 1 / Refuted 5.* **Refuted:** the wrapper embeds the original text as `; {error}` and raises `from error`, preserving `__cause__` and the traceback.

</details>

---
*Corroboration-mode council · 8 members · 25 findings submitted · convened by Deliberator. Severity tags reflect member consensus; corroborate/refute counts are noted on each comment so you can weigh contested items yourself.*

🤖 *Every checkpoint asked for ID before boarding — your friendly neighbourhood travel agent.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+        # could load different weights and let the worker serve the wrong version.
+        if envs.MX_REFIT_DESIRED_VERSION_UID is not None:
+            strategies = [
+                DesiredVersionP2PStrategy(),
+                DesiredVersionS3Strategy(),
+            ]
+        else:
```
ajcasagrande, line 228: [MEDIUM] The unset-UID fallback contains no RDMA/P2P strategy at all, so a restarted or scaled replica has no peer-bootstrap path.

The else branch is `[ModelStreamerStrategy(), DefaultStrategy()]` — `RdmaStrategy` is absent, unlike `modelexpress/load_strategy/__init__.py:70-80` (`LoadStrategyChain.all_strategies`) which puts `RdmaStrategy` first. Combined with `vllm/loader.py:160-165` making `download_model` a no-op for the RL chain (which also skips the prefetch path at `loader.py:167-176`), the only remaining acquisition route is `DefaultStrategy`'s implicit hub download.

The same pod still requests `rdma/ib` and sets `MX_RDMA_NIC_PIN` (`dgd.yaml:83-84,97`), which is then unusable on initial load — requested hardware sitting idle on the exact path it would help most.

Counterpoint from the dissent: with an empty ConfigMap there is no immutable version to key P2P on, so version-pinned peer bootstrap is undefined here, making this an optimization gap rather than a correctness failure.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -179,6 +179,188 @@ def _group_tensors_by_shard(
...
+            raise ValueError(
+                f"full HF checkpoint shard {filename!r} tensor set differs "
+                "from its index"
+            )
+
+        view = memoryview(data)
+        for name in tensor_names:
+            tensor_header = header[name]
+            begin, end = tensor_header["data_offsets"]
+            if tensor_metadata is not None:
```
ajcasagrande, line 245: [MEDIUM] Cold-start full-checkpoint download can perform **no** content validation when the producer's index omits `checksum_format`.

Both validation blocks are conditional: the dtype/shape/byte_size comparison at 243-250 is guarded by `if tensor_metadata is not None:` (line 241), and the checksum comparison at 257 is guarded by `if checksum_format is not None:` (line 194). `bootstrap_s3_checkpoint` (`receiver.py:333-341`) calls `_download_full_checkpoint` **without** `tensor_metadata`, whereas the steady-state path always supplies it (`receiver.py:864`). Since commit 01aad49 made `checksum_format` genuinely optional, both guards can be false simultaneously — an unvalidated cold-start load.

`tests/test_refit_s3_receiver.py:509-541` covers only the checksummed happy path; no test exercises a checksum-less cold-start root.

Fairly noted by the dissent: `_download_full_checkpoint` with a `None` default predates this PR and the artifact is still structurally verified against its own index (`receiver.py:231-235`). This is a new *reachable* path rather than a new bug — but the new path is the one that ships unvalidated.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/modelexpress_rl/inference/receiver.py
```diff
@@ -179,6 +179,188 @@ def _group_tensors_by_shard(
...
+                reusable = False
+        if not reusable:
+            index_data = s3.get(version.uri)
+            metadata, weight_map = _parse_index_manifest(
+                index_data,
+                is_delta=False,
+            )
+            store.ensure_capacity(
+                len(index_data),
+                protected_versions={version.version_id},
```
ajcasagrande: [MEDIUM] `bootstrap_s3_checkpoint` omits the non-weight files that `_prepare_full` copies for the same version.

This writes only `(temporary / Path(version.uri).name)` plus the `_download_full_checkpoint` output. Compare `receiver.py:692-695`, where `_prepare_full` calls `self._copy_non_weight_files(self.seed_checkpoint_path, temporary)` before writing the index.

That divergence has a concrete consequence: `modelexpress_rl/utils.py:191` does `tied = _tied_names(root if root.is_dir() else root.parent)` and `utils.py:83-91` reads `config.json` from the artifact directory to decide whether to drop `lm_head.weight` (`utils.py:153-154`). **With `config.json` absent the weight tie is never honored.** The reuse branch at `receiver.py:662-682` then accepts the pre-existing directory via `verify_artifact_source` for the same `version_id`, so the config-less artifact is reused by the live refit loop.

The dissent notes the engine model is already initialized from `model_config` before the chain and the directory is passed only to a weight iterator — so the tokenizer/config requirement is not fully established. The tied-weights path is the part worth checking directly.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+        try:
+            chain = _resolve_s3_replay_chain(ctx, desired_version_uid)
+            root = chain[0]
+            assert root.object_storage is not None
+            s3 = S3Client()
+            try:
+                seed_path = bootstrap_s3_checkpoint(
+                    model_name=ctx.identity.model_name,
+                    version=_S3Version(
+                        version_id=root.version_id,
```
ajcasagrande, line 133: [LOW] Possible double materialization of the cold-start root.

This calls `bootstrap_s3_checkpoint(...)` for a seed path, then constructs `CanonicalDeltaUpdateMethod._LocalCheckpoint` with `initial_base_version_id=root.version_id`. `initialize()` finds READY `full/<root>`; if `_prepare_delta` (`receiver.py:767-801`) is subsequently reached, `base_checkpoint.parent` (full cache) != `materialized_cache`, so `ensure_capacity` + `replace_directory(copy_from=base_checkpoint)` duplicates the entire root into `materialized/` (`receiver.py:790-797`).

The dissent makes a good case that this does not fire for a single-full-checkpoint chain: after bootstrap writes READY state for the root (`receiver.py:344-358`), `prepare_chain` sees `state['version'] == target.version_id` and returns at `receiver.py:506-523`; and for root+delta the copy is *required* to avoid mutating the immutable root.

Remaining actionable point regardless of which reading holds: `test_refit_s3_receiver.py:509-543` exercises `bootstrap_s3_checkpoint` in isolation and never the full strategy chain, so neither behaviour is covered.

*Corroborated 5/8, 2 refutations.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+                config=ObjectStorageGeneratorConfig(
+                    storage_type=ObjectStorageType.S3,
+                    initial_base_version_id=root.version_id,
+                    seed_checkpoint_path=seed_path,
+                    refit_checkpoint_dir=checkpoint_dir,
+                ),
+            )
+            prepared = method.prepare_chain(
+                tuple(
+                    (
```
ajcasagrande, line 155: [MEDIUM] `DesiredVersionP2PStrategy.load` leaves `ctx.identity.revision` mutated on success — only the exception path restores it.

`original_revision = ctx.identity.revision; ctx.identity.revision = desired_version_uid; try: ... except BaseException: ctx.identity.revision = original_revision; raise`. The revision participates in the source identity hash (`metadata/source_id.py:70`, `"revision": identity.revision.lower()`).

The concern raised: a worker whose boot context revision was overwritten could later advertise the pinned UID even when serving a different version, breaking the P2P handshake on subsequent refits.

The dissent is substantive and may well be dispositive — the active vLLM refit runtime defines `build_identity(version_id)` at `modelexpress_rl/inference/engines/vllm/__init__.py:43-46`, builds a *fresh* engine identity and explicitly assigns each active version ID, so the boot context is not reused for later publications. Leaving the boot revision at the loaded UID is then correct for boot-source publication.

Worth a confirming look at whether any code path publishes from the boot `ctx.identity` after a later refit; if not, this can be closed.

*Corroborated 4/8, 2 refutations.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+            raise
+        except Exception as error:
+            ctx.identity.revision = original_revision
+            raise StrategyFailed(str(error), mutated=engine_loaded) from error
+        finally:
+            if method is not None:
+                try:
+                    if prepared is not None:
+                        method.release(prepared)
+                except Exception:
```
ajcasagrande, line 181: [MEDIUM] Server-URL fallback — please verify, the council split on this one.

The finding: `ModelExpressControlClient.connect(server_url=ctx.mx_server_url)` uses only `ctx.mx_server_url`, with no fallback to `envs.MODEL_EXPRESS_URL` / `envs.MX_SERVER_ADDRESS`, unlike the established pattern at `rdma_strategy.py:125-132` (`ctx.mx_server_url or envs.MODEL_EXPRESS_URL or envs.MX_SERVER_ADDRESS`). On decentralized k8s-service backends where `ctx.mx_server_url` may be unset, `get_weight_version` would raise `grpc.RpcError`, failing the boot on the path that was supposed to be the fallback.

The strongest refutation appears correct on the code as written: `connect` delegates to `_get_server_url` (`control.py:127`), and `modelexpress/client.py:93-111` already falls back from a falsey explicit URL to `MODEL_EXPRESS_URL`, then `MX_SERVER_ADDRESS`, then localhost. If that chain holds, no change is needed here.

Flagging only so the delegation is confirmed rather than assumed — the consequence if it does not hold is a hard boot failure.

*Corroborated 4/8, 3 refutations — net positive but genuinely contested.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def load(self, result: LoadResult, ctx: LoadContext) -> LoadResult:
+        desired_version_uid = envs.MX_REFIT_DESIRED_VERSION_UID
+        if desired_version_uid is None:
+            raise StrategyFailed("desired version is not configured", mutated=False)
+        original_revision = ctx.identity.revision
+        ctx.identity.revision = desired_version_uid
+        try:
+            return super().load(result, ctx)
+        except BaseException:
```
ajcasagrande, line 93: [MEDIUM] Asymmetric interrupt handling between the two desired-version strategies.

`DesiredVersionP2PStrategy.load()` catches `BaseException` to restore the revision (lines 52-54); `DesiredVersionS3Strategy.load()` catches only `Exception` (lines 128-133).

The dissent explains the asymmetry convincingly: S3 does not mutate `ctx.identity.revision` before the interruptible work — it captures the original at line 77 and assigns the desired UID only *after* `load_uri` succeeds (lines 122-125), so an interrupt leaves the original identity already intact. P2P mutates before calling RDMA (lines 47-50) and therefore genuinely needs `BaseException`.

So the difference is defensible rather than a bug. A one-line comment noting *why* the two differ would stop the next reader (human or bot) from "fixing" it into symmetry.

*Corroborated 5/8, 2 refutations.*

modelexpress_client/python/modelexpress_rl/envs.py
```diff
@@ -55,6 +57,12 @@
     "MX_REFIT_CHECKSUM_FORMAT": lambda: (
         os.environ.get("MX_REFIT_CHECKSUM_FORMAT", "adler32").strip().lower()
     ),
+    "MX_REFIT_DESIRED_VERSION_UID": lambda: (
```
ajcasagrande, line 60: [LOW] Blank/whitespace values collapse to `None` via `or None`, which is what makes the empty-ConfigMap boot race above reachable.

The DGD example ships `desired_version_uid: ""` (`dgd.yaml:11`), so this maps to `None` and the desired-version strategies are disabled until the orchestrator populates the value.

The dissent is right that this is explicit rather than accidental — `dgd.yaml:3-4` documents that empty means "no desired version yet" and `envs.py:59-64` implements exactly that representation. Recorded mainly as the mechanism behind the higher-severity finding on `load_strategy.py:186`, not as an independent defect.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/modelexpress/load_strategy/__init__.py
```diff
@@ -142,3 +99,68 @@ def _reinit_for_retry(
...
+
+
+def run_load_strategy_chain(model: nn.Module, ctx: LoadContext) -> nn.Module:
+    """Dispatch to the configured engine-neutral loading policy."""
+    from .. import envs
+
+    chain = envs.MX_LOAD_STRATEGY_CHAIN
+    if chain == "INFERENCE":
+        return LoadStrategyChain.run(model, ctx)
+    if chain == "RL":
```
ajcasagrande, line 159: [LOW] Layering inversion: the base `modelexpress` package now imports `modelexpress_rl` at dispatch time.

`if chain == "RL": from modelexpress_rl.inference.load_strategy import RLLoadStrategyChain`. That module imports at top level from `modelexpress.load_strategy`, `..s3`, `.methods` and `.receiver` (`load_strategy.py:11-30`), reversing the existing direction where `modelexpress_rl` depends on `modelexpress` and not the other way round. A missing or broken RL dependency surfaces as a bare `ImportError` from inside `load_model` rather than the validated `ValueError` — the validation at `__init__.py:163-166` fires only for unknown chain names.

Test gap: `tests/test_vllm_loader.py:156-172` patches `RLLoadStrategyChain.run` directly, so the import is never exercised as a failure mode; `test_vllm_loader.py:174-181` covers only the unknown-value `ValueError`.

Mitigating, per the dissent: the import is lazy and gated on an explicit `RL` selection, both packages ship in the same distribution, and `boto3` is not pulled in until `S3Client.__init__`. Low priority — a wrapped import error with an actionable message would close it.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/modelexpress/engines/vllm/loader.py
```diff
@@ -157,6 +157,13 @@ def load_model(
 
     def download_model(self, model_config: ModelConfig) -> None:
         """Download the model so it can be loaded immediately."""
+        if envs.MX_LOAD_STRATEGY_CHAIN == "RL":
```
ajcasagrande, line 166: [LOW] The RL early-return produces a confusing runtime error rather than a clear startup error when weight sources are unavailable.

`if envs.MX_LOAD_STRATEGY_CHAIN == "RL": logger.info(...); return` reads `envs.MX_LOAD_STRATEGY_CHAIN` directly here, while strategy dispatch goes through `run_load_strategy_chain` — two independent reads of the same switch that must agree.

The dissent correctly notes `DefaultStrategy` is disk-*or-hub* (`default_strategy.py:3,16-30`) so this does not guarantee failure. The surviving point is diagnostic: when the S3/P2P paths are unavailable, nothing logs that the native pre-pull was deliberately skipped, so the eventual failure appears at weight-load time with no breadcrumb back to this early return.

*Corroborated 6/8, 1 refutation.*

modelexpress_client/python/tests/test_refit_load_strategy.py
```diff
@@ -0,0 +1,206 @@
...
+            uri=f"s3://weights/{uid}/model.safetensors.index.json",
+        ),
+        expected_source_slots=(),
+        layout_signature="layout",
+        state=WeightVersionState.READY,
+        created_at_unix_ms=1,
+    )
+
+
+def test_desired_version_does_not_use_version_agnostic_fallbacks(monkeypatch):
```
ajcasagrande, line 64: [LOW] Coverage question on the unset-desired-version branch — likely already satisfied.

The finding argued that `test_desired_version_does_not_use_version_agnostic_fallbacks` only covers the UID-set case, leaving the fallback path untested.

The refutation looks correct: `tests/test_refit_load_strategy.py:121-149` unsets `MX_REFIT_DESIRED_VERSION_UID`, `MX_REFIT_CHECKPOINT_DIR` and `MX_MODEL_URI`, runs `RLLoadStrategyChain`, and asserts the default fallback is called once.

Reported at net-positive support for completeness; no action expected beyond confirming that test does assert the *absence* of the desired-version strategies, not merely the presence of the fallback.

*Corroborated 4/8, 3 refutations.*

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
anson627: Yes, failing startup is the correct behavior when a desired delta version is explicitly configured and neither P2P nor S3 can load it

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
nv-hwoo, line 1: Looks like there might be some identity conflict between the source published by our full tensor nixl p2p method (which publishes pre-PWAL tensors) and the this post-PWAL p2p. The DesiredVersionP2PStrategy only changes `ctx.identity.revision` and reuses the generic `RdmaStrategy`, and there's no way to distinguish these two source by whether they are pre-PWAL or post-PWAL.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: For desired v23, MX resolves the complete chain backward to the full root, then replays it in order: v20 (FULL) → v21 → v22 → v23. It downloads or reuses the full v20 checkpoint, then downloads and applies every required delta. If artifacts are already cached locally, they are validated and reused rather than downloaded again.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: All ranks read the desired version from MX_REFIT_DESIRED_VERSION_UID, which is mounted by a config map. The only edge case is changing the shared ConfigMap while a multinode worker is starting could produce rank 0 targeting N and rank 1 targeting N+1.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Yeah the current FullTensorNixlUpdateMethod is added for the reshard refit path, not intended to work with the delta weight + p2p refit path. This PR is only for inference-loader can work with RL use cases, so it should be addressed in the next PRs.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Good call, we should connect the MX loaded weight version with vLLM persisted weight version, so once a replica is initially loaded via MX, it can correctly report version via vLLM's control API

zhengluo-nv: A proper implementation needs ~1k loc. Will do in the next PR

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Fixed in `d1dfa53b`. Both desired-version paths now use the same control-plane validator, which rejects a response unless `version.version_id` exactly matches the requested UID before any checkpoint is materialized. Added a regression test for a mismatched returned UID.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Fixed in `d1dfa53b`. The P2P strategy now fetches the desired `WeightVersion` first and requires exact UID, `READY` state, and matching model ownership before changing the load identity or attempting RDMA. A validation miss remains non-mutating and falls through to desired-version S3.

examples/rl/dynamo_vllm_refit/dgd.yaml
zhengluo-nv: Addressed in `d1dfa53b`. The example keeps a node-local `emptyDir` by design, now bounds it at `500G` to match the checkpoint-store default, and documents that Pod replacement removes the cache and may require a full redownload and delta replay.

modelexpress_client/python/modelexpress/engines/vllm/loader.py
zhengluo-nv: Updated in `55e40512` after the follow-up design decision. vLLM now rejects `draft + RL` before building the load context or initializing the model, because the desired UID identifies only the main model and there is no version contract for draft weights yet. The regression test asserts both initialization steps are skipped.

modelexpress_client/python/modelexpress/engines/vllm/loader.py
zhengluo-nv: No code change here. The early return is intentional for the RL policy: weight acquisition must remain under the strategy chain so a configured desired UID cannot be preloaded through a version-agnostic path. With no desired UID, `DefaultStrategy` invokes the engine-native loader directly and can still use disk or Hub. The early return already emits an INFO breadcrumb.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: The unset-UID behavior is intentional and now explicit in `d1dfa53b`: it emits a WARNING before using ModelStreamer and engine-default fallbacks. The deployment docs also state this split. Fail-closed applies once an immutable desired UID is configured; an empty ConfigMap means the orchestrator has not established that constraint yet.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: No change. Without a desired immutable UID there is no safe revision key for peer discovery, so adding generic RDMA here could select arbitrary weights. The unset branch intentionally uses `MX_MODEL_URI` and then the engine-native loader; desired-version P2P is enabled only after the orchestrator supplies a UID.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: No change in this PR. Standard HF indexes may omit ModelExpress checksum metadata, so bootstrap cannot compare content to an authoritative digest that does not exist. It still validates the safetensors structure, index-to-shard tensor membership, offsets, and shard completeness. Computing a new local checksum would not authenticate the source; stronger content validation needs a producer/control-plane digest contract.

modelexpress_client/python/modelexpress_rl/inference/receiver.py
zhengluo-nv: No change. This cold-start artifact is intentionally weights-only and is consumed after the engine has already initialized from its own model configuration. Copying non-weight files from the launch model could attach stale configuration to a newer immutable weight version. The checkpoint index remains authoritative for its tensor set, and live loading uses the already-constructed engine model.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Confirmed no erroneous double materialization. A full-only target reuses the immutable full cache directly. A root-plus-delta chain creates one derived materialization because mutating the immutable root would destroy recovery lineage. `test_cold_start_ranks_do_not_rewind_prepared_target` now exercises root-plus-delta preparation and follower reuse without a second S3 reconstruction.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Confirmed this is intentional. On success the boot context must retain the exact revision actually installed so initial source publication identifies the loaded desired version. On failure it is restored. Later active refits build a fresh per-version engine identity rather than reusing this boot context, so they do not publish a stale boot revision.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Confirmed no change is needed. `ModelExpressControlClient.connect(server_url=ctx.mx_server_url)` delegates to `_get_server_url`; a falsey explicit value falls back through `MODEL_EXPRESS_URL`, `MX_SERVER_ADDRESS`, and then localhost. This call follows the established endpoint resolution path.

modelexpress_client/python/modelexpress_rl/inference/load_strategy.py
zhengluo-nv: Documented the intentional asymmetry in `d1dfa53b`. P2P changes `ctx.identity.revision` before interruptible RDMA work, so it catches `BaseException` to restore it. S3 changes the identity only after loading succeeds, so an interrupt before that point leaves the original identity untouched.

modelexpress_client/python/modelexpress_rl/envs.py
zhengluo-nv: Confirmed intentional. Empty or whitespace means the orchestrator has not supplied a desired immutable version, so it maps to `None`. `d1dfa53b` adds a WARNING at that decision point, while a non-empty UID continues to fail closed if desired-version P2P and S3 are exhausted.

modelexpress_client/python/modelexpress/load_strategy/__init__.py
zhengluo-nv: No change. The import is deliberately lazy, runs only for an explicit `RL` selection, and `modelexpress` plus `modelexpress_rl` ship in the same Python distribution. A missing RL package would indicate a broken installation rather than an optional capability miss, so preserving the original import failure is preferable to wrapping it with a second error type.

modelexpress_client/python/modelexpress/engines/vllm/loader.py
zhengluo-nv: Confirmed the current code already logs this decision before returning: `RL initial load is selected; leaving weight acquisition to the RL strategy chain`. No additional branch or duplicated switch is needed; the strategy executor then logs each eligible and attempted loader.

modelexpress_client/python/tests/test_refit_load_strategy.py
zhengluo-nv: Confirmed. `test_missing_optional_sources_reaches_engine_default` clears the desired UID and optional sources, verifies the engine-default strategy is called, and now also asserts the explicit unset-UID WARNING. The desired-version strategies are constructed only in the opposite branch, so they cannot run in this case.

zhengluo-nv:
Addressed the Review Council feedback in `d1dfa53b`:

- unified exact UID, READY-state, and model-ownership validation across desired-version P2P and S3;
- excluded speculative drafts from the main model desired-version policy;
- added an explicit warning for the intentional unset-UID fallback;
- bounded and documented the node-local checkpoint cache;
- updated `docs/ARCHITECTURE.md` and `docs/DEPLOYMENT.md`;
- added focused regression coverage.

Local validation: `1814 passed, 33 skipped`; all selected pre-commit hooks passed. I replied to each inline finding with either the fix or the verified reason no code change is appropriate.

nv-hwoo (approved):

anson627 (approved):

PR #728, opened 2026-09-05
fix(crypto): keep ring out of OpenSSL builds

zdtsw:
## Description
OpenSSL feature builds enabled `ring` through the workspace `object_store` dependency, even when the native-tls backend was selected. This caused the OpenSSL build to include the Rustls crypto backend.

## Changes
- Removed the global `ring` feature from the `object_store` dependency.
- Enabled `object_store/ring` only through the `tls-rustls` feature.
- Kept the OpenSSL path on the native-tls backend.
- Preserved the Rustls backend behavior.

## Testing
- `cargo check --workspace --all-targets --no-default-features --features openssl`
- `cargo check --workspace --all-targets --no-default-features --features tls-rustls`
- Confirmed `ring` is absent from the OpenSSL dependency tree.
- `pre-commit run`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Rustls-based TLS configuration so required cryptography support is enabled consistently when that TLS option is selected.
  - Preserved existing AWS and HTTP client integrations while avoiding unnecessary cryptography features in other configurations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ modelexpress-common = { path = "modelexpress_common", version = "0.5.1", default
 modelexpress-client = { path = "modelexpress_client", version = "0.5.1", default-features = false }
 modelexpress-server = { path = "modelexpress_server", version = "0.5.1", default-features = false }
 once_cell = "1.21.3"
-object_store = { version = "0.14.1", default-features = false, features = ["aws-base", "reqwest", "ring"] }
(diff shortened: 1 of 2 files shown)


zhengluo-nv:
/ok to test [76c01a7](https://github.com/ai-dynamo/modelexpress/pull/728/commits/76c01a765ef1e708f06e3a778dcd41ee7a353ece)

zhengluo-nv (approved):

PR #733, opened 2026-09-08
fix(client): prefer ModelStreamer for explicit model URIs

AndyDai-nv:
## Summary

- skip InstantTensor whenever `MX_MODEL_URI` is configured, treating the explicit URI as selection of ModelStreamer
- avoid probing ModelStreamer partial local caches with InstantTensor and triggering a mutated model retry/reinitialization
- document the updated eligibility rule and cover remote, local, and Hugging Face-style model URIs

## Motivation

With an object-store `MX_MODEL_URI`, vLLM can expose a ModelStreamer cache containing the safetensors index before every referenced shard is materialized. InstantTensor currently runs before ModelStreamer, consumes that index as a complete local checkpoint, then fails on missing shards. Because the failure occurs while applying the lazy weight iterator, ModelExpress conservatively reinitializes the model. That extra initialization can change vLLM compile-cache state and cause a P2P destination to compute different torch.compile/AOT cache keys from its source.

`MX_MODEL_URI` is already the ModelStreamer activation signal, so this change does not enumerate or special-case URI schemes.

## Tests

- `pytest tests/test_instant_tensor_strategy.py tests/test_model_streamer_strategy.py tests/test_vllm_loader.py -q` (130 passed)
- `pre-commit run`

## Summary by CodeRabbit

* **Documentation**
  * Clarified that InstantTensor is available only when `MX_MODEL_URI` is unset.
  * Documented that configuring `MX_MODEL_URI` selects the ModelStreamer loading path.
  * Updated deployment guidance for InstantTensor defaults and loading behavior.

* **Bug Fixes**
  * InstantTensor now skips loading when a model URI is configured, allowing the appropriate alternative loader to handle the model.

* **Tests**
  * Added coverage for cloud, local, and model-reference URI configurations.


AndyDai-nv:
Closing this approach after design review. MX_MODEL_URI being set is not a reliable signal that InstantTensor should be disabled: it may refer to a complete local directory or another setup where InstantTensor remains appropriate. We will keep strategy selection explicit and document MX_INSTANT_TENSOR=0 for object-storage ModelStreamer deployments instead.

PR #734, opened 2026-09-08
docs: disable InstantTensor for object-store streaming

AndyDai-nv:
## Summary

- document `MX_INSTANT_TENSOR=0` for vLLM ModelStreamer deployments backed by S3, GCS, or Azure
- explain that InstantTensor expects a complete local checkpoint and can otherwise probe a partially materialized ModelStreamer cache before the streaming strategy runs
- keep InstantTensor recommended for complete local checkpoints, including init-container-synchronized weights
- update the checked-in vLLM S3 and Azure manifests with the safe setting

## Context

InstantTensor intentionally precedes ModelStreamer in the loading strategy chain. When a RunAI ModelStreamer cache contains a safetensors index but not every referenced shard, InstantTensor can begin applying weights and then fail on a missing shard. The resulting model reinitialization can prevent torch.compile artifacts from being reused by a P2P target.

We intentionally keep this as explicit deployment configuration rather than inferring eligibility from whether `MX_MODEL_URI` is set: that variable may also identify a complete local model or another environment where InstantTensor is appropriate.

## Validation

- `pre-commit run` (all hooks passed, including YAML validation)

## Summary by CodeRabbit

* **Documentation**
  * Added guidance to set `MX_INSTANT_TENSOR=0` when using vLLM with ModelStreamer object-storage URIs.
  * Clarified that this prevents incomplete, partially materialized local caches from being treated as complete checkpoints.
  * Updated deployment and troubleshooting instructions for Azure Blob and S3-backed ModelStreamer configurations.
  * Updated Kubernetes examples and manifests with the required setting for lazy checkpoint materialization.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -61,6 +61,8 @@ P2P transfers weight tensors, not the whole model repository. Targets still need
 
 Verify `MX_MODEL_URI` and storage access from inside the runtime container. The prefix must contain the expected safetensors and index/configuration files. Direct ModelStreamer does not use the ModelExpress server.
 
+For vLLM object-storage URIs, set `MX_INSTANT_TENSOR=0`. If logs show InstantTensor reporting weight files referenced by the safetensors index as missing, it has inspected ModelStreamer's partially materialized local cache before ModelStreamer ran. Disabling InstantTensor avoids the failed partial load and subsequent model reinitialization. Complete checkpoints synchronized to local disk can continue to use InstantTensor.
+
 For SGLang, keep the model identity in `--model-path` and the object URI in `MX_MODEL_URI`. Passing the object URI as `--model-path` selects SGLang's native loader. Use the checked-in [vLLM](../examples/model_streamer_k8s/client/vllm/README.md) or [SGLang](../examples/model_streamer_k8s/client/sglang/README.md) examples as the baseline.
 
 ## Kubernetes rollout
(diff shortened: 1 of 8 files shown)


zhengluo-nv (dismissed):

docs/TROUBLESHOOTING.md
```diff
@@ -61,6 +61,8 @@ P2P transfers weight tensors, not the whole model repository. Targets still need
 
 Verify `MX_MODEL_URI` and storage access from inside the runtime container. The prefix must contain the expected safetensors and index/configuration files. Direct ModelStreamer does not use the ModelExpress server.
 
+For vLLM object-storage URIs, set `MX_INSTANT_TENSOR=0`. If logs show InstantTensor reporting weight files referenced by the safetensors index as missing, it has inspected ModelStreamer's partially materialized local cache before ModelStreamer ran. Disabling InstantTensor avoids the failed partial load and subsequent model reinitialization. Complete checkpoints synchronized to local disk can continue to use InstantTensor.
```
zhengluo-nv, line 64: Relying on customer to set this is not ideal. 

> When a RunAI ModelStreamer cache contains a safetensors index but not every referenced shard, InstantTensor can begin applying weights and then fail on a missing shard.

Can we actually solve this problem?

zhengluo-nv (changes requested):

AndyDai-nv:
closed for 775

PR #735, opened 2026-09-08
feat(refit): reuse manifest metadata

KavinKrishnan:
## Summary

Reuses stable manifest metadata and transfer plans across warm refits. Version-specific content digests are refreshed without rebuilding the transfer plan.

The PR also adds trainer and receiver timing so refits report manifest work, discovery, transfer, reconstruction, and installation separately.

## Changes

- Reuse trainer manifests while tensor registration and layout stay unchanged.
- Reuse receiver transfer plans while endpoints, addresses, dtypes, shapes, and sharding stay unchanged.
- Refresh content digests without rebuilding a valid plan.
- Fetch fallback replicas only when the selected source fails.
- Remove released manifests and an extra JSON conversion.
- Report timing, manifest bytes, and cache hits for successful and failed refits.

## Validation

- Full Python client suite: 1931 passed, 4 skipped.
- GCP Qwen3-30B-A3B: 30 warm refits completed without restart, OOM, or refit errors.
- All warm manifest and transfer-plan lookups hit the cache.
- Warm receivers fetched 0 manifest bytes, with 4.0 ms median source discovery.
- Digest-enabled validation completed four warm refits with no digest mismatch and reused the transfer plan each time.

## Dependency

None outstanding. #704 has merged, and this branch is rebased on top of it.

## Review

Rebased onto current `main` and conflict-free. Conflicts were with #704 and were resolved by keeping its 100 MB manifest receive limit on the relocated fetch, and by keeping both its large-manifest test and this branch's staging-failure test.

Addressing @zhengluo-nv:

- **One timing API** (0dd601a). Owned work goes through `refit_span`; `modelexpress_rl.timing` is reduced to cycle ownership, and the generic helpers now live beside `refit_span` as `add_refit_duration`, `add_refit_metadata` and `set_refit_cold`. About twenty call sites converted. `refit_span` gained two things so nothing was lost: it yields a dict for counters only knowable once the work has run, and takes `duration_key` so several spans sharing one normalized stage can still say which of them was expensive.
- **Replica health is per slot again** (57d1107). The lazy fetching had started pairing replicas at a shared offset, so a fleet whose healthy replicas sat at different indices in different slots yielded no candidate even though a complete healthy set existed. Health is resolved per slot again without giving up laziness: a slot's replicas are walked only as far as the requested candidate index. Covered by `test_generator_assembles_healthy_replicas_from_different_offsets`, which fails against the offset-aligned selection and passes with the fix.

Earlier in review, both CodeRabbit findings were addressed in 6645ddf, along with a third instance of the same problem: a staging failure emitted no timing record at all, since the recorder is handed forward on the staged handle and a staging failure happens before there is one.

### DIFF
--- a/modelexpress_client/python/modelexpress_rl/inference/client.py
+++ b/modelexpress_client/python/modelexpress_rl/inference/client.py
@@ -15,7 +15,7 @@ from typing import Any
 import grpc
 from modelexpress import auth, envs
 from modelexpress.client import _get_server_url
-from modelexpress.refit.timing import RefitTimingRecorder
+from modelexpress.refit.timing import RefitTimingRecorder, refit_span
 
 from modelexpress_rl import envs as rl_envs
 from modelexpress_rl import timing
@@ -344,7 +344,7 @@ class ModelExpressGeneratorClient:
             try:
                 with timing.active(recorder):
                     if self._runtime.initial_version_id is not None:
-                        with timing.refit_span("control_discovery"):
+                        with refit_span("control_discovery"):
                             chain = self._resolve_replay_chain(version.version_id)
                         update = (
                             self._runtime.session.stage(chain[0])
@@ -352,7 +352,7 @@ class ModelExpressGeneratorClient:
                             else self._runtime.session.stage_chain(chain)
                         )
                     else:
-                        with timing.refit_span("control_discovery"):
+                        with refit_span("control_discovery"):
                             ready = self._get_ready_version(version.version_id)
                         update = self._runtime.session.stage(ready)
             except BaseException:
--- a/modelexpress_client/python/modelexpress_rl/inference/client.py
+++ b/modelexpress_client/python/modelexpress_rl/inference/client.py
@@ -17,7 +17,8 @@ from modelexpress import auth, envs
 from modelexpress.client import _get_server_url
 from modelexpress.refit.timing import RefitTimingRecorder
 
-from modelexpress_rl import envs as rl_envs, timing
+from modelexpress_rl import envs as rl_envs
+from modelexpress_rl import timing
 from modelexpress_rl.version import WeightVersionRef
 
 from .. import refit_pb2, refit_pb2_grpc
@@ -340,19 +341,27 @@ class ModelExpressGeneratorClient:
                 version_id=version.version_id,
                 rank=rl_envs.LOCAL_RANK,
             )
-            with timing.active(recorder):
-                if self._runtime.initial_version_id is not None:
-                    with timing.refit_span("control_discovery"):
-                        chain = self._resolve_replay_chain(version.version_id)
-                    update = (
-                        self._runtime.session.stage(chain[0])
-                        if len(chain) == 1
-                        else self._runtime.session.stage_chain(chain)
-                    )
-                else:
-                    with timing.refit_span("control_discovery"):
-                        ready = self._get_ready_version(version.version_id)
-                    update = self._runtime.session.stage(ready)
+            try:
+                with timing.active(recorder):
+                    if self._runtime.initial_version_id is not None:
+                        with timing.refit_span("control_discovery"):
+                            chain = self._resolve_replay_chain(version.version_id)
+                        update = (
+                            self._runtime.session.stage(chain[0])
+                            if len(chain) == 1
+                            else self._runtime.session.stage_chain(chain)
+                        )
+                    else:
+                        with timing.refit_span("control_discovery"):
+                            ready = self._get_ready_version(version.version_id)
+                        update = self._runtime.session.stage(ready)
+            except BaseException:
+                # Nothing else will report this cycle: the recorder is handed on
+                # through the staged handle, and staging failed before there was
+                # one. A refit that died on the wire is exactly the case the
+                # stage split exists to explain.
+                timing.emit(recorder, logger)
+                raise
             self._active_handle = StagedWeightHandle(
                 client=self,
                 version_id=version.version_id,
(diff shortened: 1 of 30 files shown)


modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -294,20 +299,28 @@ def stage_weight(self, *, version: WeightVersionRef) -> StagedWeightHandle:
...
-                    else self._runtime.session.stage_chain(chain)
-                )
-            else:
-                ready = self._get_ready_version(version.version_id)
-                update = self._runtime.session.stage(ready)
+            recorder = timing.start_cycle(
+                version_id=version.version_id,
+                rank=rl_envs.LOCAL_RANK,
+            )
+            with timing.active(recorder):
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in 6645ddf. Staging now emits on failure, and release emits from a finally block. a4d881a adds regression coverage for the release-failure path.

modelexpress_client/python/modelexpress_rl/inference/engines/vllm/installer.py
```diff
@@ -213,8 +214,13 @@ def parameter_layout(self) -> dict[str, tuple[tuple[int, ...], torch.dtype]]:
...
         self._process_and_commit(tensors)
-        _update_mla_absorbed_weights(self._model, quantized=self._is_quantized)
-        torch.cuda.synchronize(self._device)
+        # Derived-weight fixups plus the synchronize that makes the whole
+        # install observable. Separate from the per-layer stages because it is
+        # paid once and does not scale with the number of layers, so folding it
+        # in would make those look worse than they are on small models.
+        with refit_span("post_install"):
+            _update_mla_absorbed_weights(self._model, quantized=self._is_quantized)
+            torch.cuda.synchronize(self._device)
```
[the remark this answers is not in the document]
KavinKrishnan: Fixed in 6645ddf. Checkpoint MLA fixups and CUDA synchronization now run inside post_install. a4d881a adds an assertion that checkpoint installation records that span.

modelexpress_client/python/modelexpress_rl/inference/methods/full_tensor.py
```diff
@@ -92,15 +118,45 @@ def prepare(
                 self._active_plan is not None
                 and self._active_fingerprint == inputs.physical_fingerprint
             )
+            timing.record_cold(not reusable)
+            manifests = [item.transport.manifest for item in inputs.sources]
+            manifest_digests = tuple(item.manifest_digest for item in inputs.sources)
             if not reusable:
+                started = time.perf_counter()
```
zhengluo-nv: Can we use the existing `modelexpress.refit.timing` API consistently? Prefer `refit_span()` for synchronous work owned by the current function, and reserve explicit `perf_counter()` plus `record_measured()` for durations produced by another layer or reused for throughput or legacy metrics. The generic duration-recording helper should probably live beside `refit_span()` in `modelexpress.refit.timing`, leaving the RL module responsible only for cycle ownership.

modelexpress_client/python/modelexpress_rl/inference/source/trainer.py
```diff
@@ -56,88 +58,171 @@ def candidates(self, version: WeightVersion) -> Iterator[ResolvedSource]:
...
+                logger.warning(
+                    "no trainer source published for required slot %s",
+                    source_slot_id,
+                )
+                return
+            ordered_slots.append(ordered)
+
+        seen = set()
+        candidate_count = max((len(slot) for slot in ordered_slots), default=1)
+        for offset in range(candidate_count):
```
zhengluo-nv: Replica fallback is selected by aligned offset before reachability is checked. If slot A has only `a0` healthy and slot B only `b1` healthy, this loop tries `(a0, b0)` and `(a1, b1)` and yields no candidate even though the complete healthy set `(a0, b1)` exists. Please resolve usable replicas independently per source slot before assembling candidate sets, as the previous flow did, while retaining lazy fallback fetching. I reproduced this against the current head with a focused two-slot test.

modelexpress_client/python/modelexpress_rl/inference/methods/full_tensor.py
KavinKrishnan: Done in 0dd601a. Owned work now goes through `refit_span`; `modelexpress_rl.timing` is reduced to cycle ownership (`start_cycle`, `active`, `emit`), and the generic helpers moved beside `refit_span` in `modelexpress.refit.timing` as `add_refit_duration`, `add_refit_metadata` and `set_refit_cold`. About twenty call sites across the inference and trainer paths converted.

Two additions were needed so nothing was lost in the move. `refit_span` now yields a dict for counters only knowable once the work has run, which is what several of the hand-rolled timers were actually for -- a published shard count, a cache outcome. And it takes `duration_key` to file its own duration under a metadata name: several spans share one normalized stage, and the stage total cannot say whether a slow warm refit is waiting on the wire or on the CPU, which is what `manifest_fetch_s` versus `manifest_hash_s` answers.

`add_refit_duration` is now the exception you described rather than the default: the transport's own wire and reconstruct measurements, and the publisher's manifest generation and serialization. I also added `add_refit_metadata` for the two sites that were filing sizes and cache outcomes through a zero-second duration, which worked and also left those stages reporting more spans than they had run.

modelexpress_client/python/modelexpress_rl/inference/source/trainer.py
KavinKrishnan: Confirmed and fixed in 57d1107 -- thanks for reproducing it, that was a real regression I introduced with the lazy fetching.

Health is resolved per slot again. `_SlotReplicas` walks a slot's replicas only as far as the requested candidate index, so the independence you asked for is back without giving up the lazy fetch: a healthy fleet still fetches exactly one manifest per slot, and a degraded one fetches only as far as its first healthy replica. Candidate assembly then takes the k-th *usable* replica of each slot rather than the k-th published one, so your `(a0, b1)` case is now the first candidate yielded. A slot with fewer healthy replicas cycles its own rather than ending the search; a slot with none is still terminal.

Added `test_generator_assembles_healthy_replicas_from_different_offsets` covering exactly your two-slot scenario. I checked it fails against the offset-aligned selection ("no usable trainer source for required slot rank:1") and passes with the fix.

zhengluo-nv (approved):

KavinKrishnan:
/ok to test [57d1107](https://github.com/ai-dynamo/modelexpress/pull/735/commits/57d11076e6069a92412e27011911c5a0fd80dcdc)

PR #740, opened 2026-09-09
feat(refit): add P2P-first active refit

zhengluo-nv:
#738 has merged and this branch is based on `main`, so this PR contains only the active-refit P0 follow-up.

## Why

An active generator should install an exact requested version from an existing generator when possible, while retaining canonical S3 as the durable fallback. Generator P2P must transfer the live post-PWAL representation; trainer inputs remain load-time tensors that vLLM processes through PWAL before they become runnable weights.

## What changed

- object-storage generators with compatible runtime-tensor support default to exact-version `GENERATOR` P2P, then canonical `OBJECT_STORAGE`; either order can be configured explicitly
- generator P2P stages the donor's complete post-PWAL runtime tensor set and copies it into the recipient's existing live storage only at the engine safe point
- generator P2P reuses the inference loader's rank-local NIXL manager for publication and receive buffers instead of creating a second agent; the loader remains the lifecycle owner
- trainer NIXL remains a separate load-time tensor method and continues through vLLM's PWAL installation path, preserving trainer-to-generator reshard refit
- model-quantized vLLM workers and workers using FP8 KV cache declare runtime-tensor generator P2P unavailable; when S3 is configured they select canonical object storage before staging or engine mutation
- if generator preparation is unavailable, the session resolves and synchronously reconstructs the complete S3 lineage from the full HF root through the target deltas
- canonical S3 bootstrap preserves the root artifact's source identity, allowing reconstructed checkpoints to pass installation validation
- a successful generator P2P update does not start background checkpoint rebuilding; a later peer miss reconstructs the requested version from S3 in the foreground
- staging owns source selection and fallback, while installation remains a separate safe-point operation
- adds a controlled Dynamo + vLLM E2E environment for full-HF plus S3-delta publication, active refit, generator fallback, desired-version advancement, and replacement cold start
- renames the existing full-tensor example to `dynamo_vllm_reshard_refit` to distinguish it from the new S3-delta lifecycle example

## Testing

- exact PR head `0683d2e77c2b593bb492b7b2295f38bec3e46eee`: focused generator/runtime/NIXL/vLLM tests — `105 passed`
- current committed tree: complete Python suite — `1959 passed, 33 skipped`
- exact PR head: staged pre-commit hooks passed; E2E shell and Python scripts passed syntax checks
- Dynamo + vLLM functional E2E with two generators:
  - published a fresh full-HF `v1` root and XOR deltas `v2` and `v3` to S3
  - for every version, the first generator reconstructed the canonical S3 chain and the second generator selected `GENERATOR` P2P
  - inference succeeded after each worker update and both workers reported the requested serving version
  - deleting one generator produced a replacement that read desired `v3`, cold-loaded it from the surviving generator through `desired_version_p2p`, and returned with both workers reporting `v3`
  - the cold P2P transfer moved 339 runtime tensors / 1.20 GB in 0.22 s (44.4 Gbit/s reported by MX)
  - each vLLM rank used one loader-owned NIXL agent; receive buffers were registered on that same transport
- earlier Dynamo + vLLM B200 E2E at pre-squash commit `97bc1e09b354e7e07b76a8863a71874986b99d95` validated the trainer-to-generator reshard lifecycle: 1.503 GB transferred through NIXL, vLLM installed version `d6541093`, the Control API reported that version, and deterministic inference matched after resume

## Qualification limits

- the new S3-delta E2E used TP1 workers; TP>1, multinode local-rank checkpoint coordination, and rank-failure behavior are not qualified by this run
- inference smoke tests and serving-version checks are functional validation, not numerical parity or output-token version attribution
- post-PWAL generator P2P is intentionally unavailable for model-quantized configurations and FP8 KV cache until their derived runtime state can be updated safely
- multi-version S3 replay performs one atomic preparation attempt; transient whole-chain retry is deferred, and failure is recovered and surfaced before engine mutation
- the reported transfer rate is one functional-run observation, not a performance benchmark


modelexpress_client/python/modelexpress_rl/inference/session.py
```diff
@@ -161,7 +215,12 @@ def stage_chain(self, versions: tuple[WeightVersion, ...]) -> SessionUpdate:
             plans = []
             for version in versions:
                 try:
-                    plan = next(self._planner.plans(version))
+                    plan = next(
+                        self._planner.plans(
+                            version,
+                            source_kind=source_kind,
+                        )
+                    )
```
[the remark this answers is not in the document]
zhengluo-nv: Good catch. After re-checking, this is a real distinction: single-version object-storage staging gets repeated plans, while multi-version replay currently performs one atomic `prepare_chain()` call. A transient S3 HEAD/GET failure can trigger this path, and `preparation_failed()` restores the checkpoint-store state before the error is surfaced. I am deferring whole-chain retry to a follow-up rather than broadening this PR; the failure occurs before engine mutation, so the caller can safely retry the weight update.

modelexpress_client/python/modelexpress_rl/inference/client.py
```diff
@@ -353,21 +351,32 @@ def apply_weight(self, staged: StagedWeightHandle) -> Any:
...
             if staged._update.released:
                 raise RuntimeError("staged weight has already been released")
-            assert self._runtime is not None
+            runtime = self._require_runtime()
+            was_applied = staged._update.applied
+            if not was_applied:
+                runtime.unpublish_runtime_tensors()
             try:
-                result = self._runtime.session.apply(staged._update)
+                result = runtime.session.apply(staged._update)
```
anson627: Could we add a donor-side read lease/drain mechanism, or retain an immutable version-specific snapshot, before overwriting published runtime tensors? Unpublishing stops discovery/heartbeat, but leaves the NIXL agent and registered memory alive, it does not wait for peers that already fetched the descriptors. A peer pulling version V can therefore overlap this donor’s installation of V+1 and receive mixed-version bytes. The peer path checks tensor metadata but does not verify the transferred contents against a version-specific digest, so that result could be accepted as V

Also Could we add a concurrency test that pauses a peer read, starts a donor refit, and verifies that the read either completes against unchanged V or fails safely?

anson627:
we should probably validate the new GPU paths before enabling P2P by default? one useful e2e test case would be: install V through a generator peer, remove the peer for the next update, reconstruct V+1 from the full S3 lineage, and compare deterministic inference against a directly loaded reference after each resume

modelexpress_client/python/modelexpress_rl/inference/client.py
zhengluo-nv: Make sense. Will follow up in the next PR.

Currently, it exposes an env variable to enable the p2p transfer in cold start refit and active refit. So this does not block the end to end testing

```
MX_GENERATOR_SOURCE_ORDER: GENERATOR,OBJECT_STORAGE
```

anson627: OK, I will give it a try

anson627 (approved):

anson627:
approved with follow-up issue https://github.com/ai-dynamo/modelexpress/issues/752

nv-hwoo (approved):

PR #743, opened 2026-09-10
fix(client): keep retrying weight-source publication through a server outage

daii-0818:
## Motivation

When a source cannot be published for longer than `MX_PUBLISH_TIMEOUT_SECS` (default 30 minutes), `PublisherThread` **permanently gives up**: it stops the thread and, via `cleanup_fn`, tears down the worker manifest gRPC server. The weights are still resident and the worker is otherwise healthy, but it can never serve as a P2P source again without a process restart.

Blast radius: a single metadata-server outage longer than the timeout disqualifies every worker that finishes loading during the window. We hit this operationally while bringing up a multi-node P2P weight-transfer deployment: after a metadata-server blip at the wrong moment, the only way to recover the affected pods' source status was to restart them, with the full weight-load cost that implies.

The timeout semantics made sense for one-shot, best-effort publications (artifacts, reshard rendezvous), where spinning forever on a publication nobody depends on is waste. A weight source is the opposite: the publication is the point, the payload is small, and it succeeds on its own as soon as the server returns.

## Changes

- `PublisherThread` gains a `retry_publish_forever` flag. When set, `_publish_timed_out` no longer stops the thread after the timeout; it logs a single "retrying indefinitely" notice and keeps the per-tick retry cadence (the existing in-call 3-attempt backoff inside `_publish_metadata_to_server` still applies per tick). Rejection-triggered re-registration state resets the overdue latch along with the existing timeout state.
- Weight-source constructions opt in:
  - `publish_metadata_and_ready` — covers both the P2P-path payload (worker manifest served by the worker's gRPC server and the legacy dual-write payload, and is shared by the sglang and vLLM engines through the load strategy;
  - the sglang TransferEngine heartbeat in .
- One-shot publishers (artifact lifecycle / artifact transfer / reshard rendezvous) keep the bounded give-up unchanged.

## Testing

 — 30 passed, including four new cases: an overdue forever-publisher keeps trying without stopping, the overdue notice logs exactly once, a forever-publisher recovers (publishes and heartbeats READY) once the server returns, and a bounded publisher still gives up at the timeout. , , ,  all pass (the one failing  case fails identically on clean main — missing  in the local venv, unrelated).
EOF
)

## Summary by CodeRabbit

* **Reliability**
  * Metadata publication now continues retrying beyond the configured timeout, helping resident weight sources remain available when the service temporarily cannot be reached.
  * Publication automatically recovers when connectivity returns, while retaining bounded retry behavior where indefinite retries are not enabled.

* **Configuration**
  * Added the preferred `MX_REDIS_URL` environment variable. When both variables are set, it takes precedence over the legacy `REDIS_URL` setting; existing fallback behavior remains supported.


modelexpress_client/python/modelexpress/engines/sglang/loader.py
```diff
@@ -515,6 +515,10 @@ def publish_fn() -> str:
                     if envs.MX_ARTIFACT_READY_URL.strip()
                     else None
                 ),
+                # The weights stay resident; keep retrying publication through
+                # a metadata-server outage instead of giving up after
+                # MX_PUBLISH_TIMEOUT_SECS.
+                retry_publish_forever=True,
```
zhengluo-nv: Instead of using a boolean, we can use a retry timeout, where 0 means retry forever

zhengluo-nv:
Made a few changes. We should use MX_PUBLISH_TIMEOUT_SECS=0 to set infinite retry. We should not override this env variable.

zhengluo-nv:
/ok to test [5d7a61f](https://github.com/ai-dynamo/modelexpress/pull/743/commits/5d7a61f70ec1342f28433d95b4c6ad4475504db2)

PR #745, opened 2026-09-10
feat(metrics): split a P2P source attempt into the phases the strategy owns

yixinh-nv:
`mx_p2p_transfer_seconds` is one interval per source attempt, so ninety seconds
there cannot tell an RDMA read from a slow registration or a peer that took a
minute to answer the handshake. This splits the attempt into seven phases, each
recorded from the `with`-block `RdmaStrategy` already wraps around that call —
ten added lines in the strategy, nothing in the NIXL manager or any adapter
changed. `metadata` is the GetMetadata/manifest fetch that ran before the
transfer span and was never measured; the other six sum to less than the
transfer by construction.

| Family | Type | Labels | Answers |
| --- | --- | --- | --- |
| `mx_p2p_source_attempt_phase_seconds` | Histogram | `policy`, `phase`, `outcome`, `scheme` | Which part of the transfer took the time, and which phase a failed attempt died in? |

Phases: `metadata`, `prepare`, `register`, `handshake`, `receive`, `finalize`, `release`.
Two panels join the P2P clients row under the transfer panels.

## Measured on hardware

Two H200 pods on one nebius node, `Qwen2.5-7B-Instruct` (15.25 GB), SGLang, NIXL
over RDMA. The source loaded natively in 37.62 s and published; the target pulled
from it in **2.36 s** end to end.

| Phase | Mean | Share of transfer |
| --- | ---: | ---: |
| `metadata` | 8.66 ms | — (before the span) |
| `prepare` | 4.56 ms | 0.2% |
| `register` | 1.653 s | 72.8% |
| `handshake` | 181 ms | 8.0% |
| `receive` | 431 ms | 19.0% |
| `finalize` | 3.3 µs | 0.0% |
| `release` | 118 µs | 0.0% |
| **transfer** | **2.271 s** | Σ phases = 99.92% |

The RDMA read itself was 0.43 s at 283 Gbps. Most of the "transfer" is
`register`: bringing up the NIXL agent and registering 201 memory regions, of
which the `ibv_reg_mr` calls were 0.29 s. That is the distinction the single
interval could not make.

phases table

## Overhead: before and after

Same source, same 7B, same image; the only difference is which client is
pip-installed on the target — `main` at this branch's merge-base (`644d53dd`,
no phase metrics) or this branch. Metrics enabled on both. Times are the
loader's own `load_model() COMPLETE` and `Total receive time` log lines.

| Pull | `main` receive / load | branch receive / load |
| --- | ---: | ---: |
| 1 | 2.05 s / 2.14 s | 2.06 s / 2.15 s |
| 2 | 2.07 s / 2.15 s | 2.05 s / 2.14 s |

The branch is within **10 ms** of `main` on every pull, and the spread within
each variant is the same size as the gap between them. Seven `perf_counter`
pairs and seven histogram observations per attempt are not measurable against a
two-second transfer.

### DIFF
--- a/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
+++ b/modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
@@ -167,11 +167,12 @@ class RdmaStrategy(LoadStrategy):
             )
 
             try:
-                source_worker = self._fetch_worker_metadata(
-                    ctx,
-                    mx_source_id,
-                    worker_id,
-                )
+                with selection_metrics.time_source_attempt_phase(policy, "metadata"):
+                    source_worker = self._fetch_worker_metadata(
+                        ctx,
+                        mx_source_id,
+                        worker_id,
+                    )
             except Exception as e:
                 logger.warning(
                     f"[Worker {ctx.global_rank}] Failed to fetch metadata for worker {worker_id}: {e}. "
@@ -480,11 +481,14 @@ class RdmaStrategy(LoadStrategy):
         source_worker_id: str,
     ) -> LoadResult:
         """Receive fully-processed weights via RDMA from an existing source."""
+        policy = configured_policy_label()
         try:
-            result = ctx.adapter.prepare_rdma_target(result)
-            result = ctx.adapter.before_rdma_receive(result)
+            with selection_metrics.time_source_attempt_phase(policy, "prepare"):
+                result = ctx.adapter.prepare_rdma_target(result)
+                result = ctx.adapter.before_rdma_receive(result)
             self._receive_from_peer(result, ctx, source_worker, mx_source_id)
-            return ctx.adapter.after_rdma_receive(result)
+            with selection_metrics.time_source_attempt_phase(policy, "finalize"):
+                return ctx.adapter.after_rdma_receive(result)
         except StrategyFailed:
             raise
         except Exception as e:
@@ -499,7 +503,9 @@ class RdmaStrategy(LoadStrategy):
     ) -> None:
         """Receive fully-processed tensors via RDMA from the detected source."""
         receive_start = time.perf_counter()
-        register_tensors(result, ctx)
+        policy = configured_policy_label()
+        with selection_metrics.time_source_attempt_phase(policy, "register"):
+            register_tensors(result, ctx)
 
         is_p2p = bool(source_worker.worker_grpc_endpoint)
         remote_agent_name = None
@@ -525,11 +531,12 @@ class RdmaStrategy(LoadStrategy):
                 # Claimed before the dial so a fetch that fails part-way, leaving
                 # metadata that lands later, is still released.
                 remote_agent_name = source_worker.agent_name
-                ctx.nixl_manager.fetch_remote_and_wait(
-                    remote_agent_name=source_worker.agent_name,
-                    ip=host,
-                    port=int(port_str),
-                )
+                with selection_metrics.time_source_attempt_phase(policy, "handshake"):
+                    ctx.nixl_manager.fetch_remote_and_wait(
+                        remote_agent_name=source_worker.agent_name,
+                        ip=host,
+                        port=int(port_str),
+                    )
                 nixl_fetch_time = time.perf_counter() - nixl_fetch_start
                 logger.info(
                     f"[Worker {ctx.global_rank}] [TIMING] P2P NIXL metadata fetch: "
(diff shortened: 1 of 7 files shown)


modelexpress_client/python/modelexpress/metrics.py
```diff
@@ -740,6 +794,49 @@ def observe_transfer_seconds(self, policy: str, outcome: str, seconds: float) ->
...
+        can classify it alone. A phase that raises is still recorded -- a
+        receive that hits the transfer deadline is the reading someone comes
+        looking for.
+        """
+        start = time.perf_counter()
+        outcome = "ok"
+        try:
+            yield
+        except BaseException:
+            outcome = "error"
```
nv-hwoo, line 850: what is the use of `outcome` when erroring out?

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -525,11 +531,12 @@ def _receive_from_peer(
...
-                    ip=host,
-                    port=int(port_str),
-                )
+                with selection_metrics.time_source_attempt_phase(policy, "handshake"):
+                    ctx.nixl_manager.fetch_remote_and_wait(
+                        remote_agent_name=source_worker.agent_name,
+                        ip=host,
+                        port=int(port_str),
+                    )
                 nixl_fetch_time = time.perf_counter() - nixl_fetch_start
```
nv-hwoo: i see we're already measuring times for these blocks. can we choose one to avoid duplicate measurements?

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -549,9 +556,10 @@ def _receive_from_peer(
                 # Loaded here rather than inside receive_from_source so this method
                 # holds the name it is responsible for releasing.
                 add_start = time.perf_counter()
-                remote_agent_name = ctx.nixl_manager.add_remote_agent(
-                    source_worker.nixl_metadata
-                )
+                with selection_metrics.time_source_attempt_phase(policy, "handshake"):
```
nv-hwoo: this one and the one above are both "handshake" phase. is this intended?

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
```diff
@@ -480,11 +481,14 @@ def _load_as_target(
         source_worker_id: str,
     ) -> LoadResult:
         """Receive fully-processed weights via RDMA from an existing source."""
+        policy = configured_policy_label()
```
nv-hwoo, line 487: Q: could you explain what `policy` is?

nv-hwoo (commented):
Overall looks good. Can we also compare e2e numbers before/after the change to make sure we're not accidentally introducing any overhead?

modelexpress_client/python/modelexpress/metrics.py
yixinh-nv: It is what lets the dashboard say which phase a failed attempt died in, and it keeps a receive that ran to the transfer deadline out of the `ok` mean — the docstring now says so (4302c92b).

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
yixinh-nv: Agreed and fixed in 4302c92b — the context manager now hands its measured duration back on the yielded handle and the log line reads that, so each span has one clock.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
yixinh-nv: Intended: it is one phase with two transports (P2P fetches the peer's NIXL metadata from the peer, centralized loads the copy the MX server holds) and exactly one branch runs per attempt — added a comment saying so in 4302c92b.

modelexpress_client/python/modelexpress/load_strategy/rdma_strategy.py
yixinh-nv: It is the source-selection policy label (`random`, `rendezvous_hash`, `load_aware`, ...) that every P2P client family already carries, so a phase duration can be read against the policy that picked the peer — commented at the assignment in 4302c92b.

yixinh-nv:
Added an *Overhead: before and after* section to the description — `main` at the merge-base vs this branch on the same source and 7B, two pulls each, within 10 ms on every pull and inside each variant's own spread.

nv-hwoo (approved):

PR #746, opened 2026-09-11
fix(crd): bound uint32 fields in OpenAPI schema

kevin85421:
## Summary

- cap the generated OpenAPI schemas for `fileCount`, `chunkCount`, and `nodeRank` at `u32::MAX`
- keep the Rust and protobuf fields as unsigned 32-bit integers while representing them as bounded `int64` values in Kubernetes

## Motivation

This follows up on the unsigned integer TODO in #722. Kubernetes CRD OpenAPI schemas support `int32` and `int64` formats, but not `uint32`. The existing non-negative `int64` schemas accept values above `u32::MAX`, which the Kubernetes API server can store but ModelExpress cannot deserialize into the corresponding Rust fields.

These fields are all part of the CR status, and the ModelExpress server should be their only writer, so an overflow should not occur during normal operation. However, defining the upper bounds is still useful because it keeps Kubernetes OpenAPI validation consistent with the Rust and protobuf `u32` contract and prevents the API server from storing values that the server cannot deserialize.

Using `format: int64` with a minimum of `0` and a maximum of `4294967295` represents the full `u32` range and rejects incompatible values at API validation time.

## Testing

- `cargo test -p modelexpress-server --bin crdgen`
- `cargo run --quiet -p modelexpress-server --bin crdgen | cmp - examples/crds.yaml`
- verified that the example and Helm CRD manifests are identical
- `python3 -m pre_commit run`

* Manually patch a CR status to `u32_max + 1`.
  image

### DIFF
--- a/examples/crds.yaml
+++ b/examples/crds.yaml
@@ -120,17 +120,20 @@ spec:
                       chunkCount:
                         description: Number of transfer chunks in the sealed artifact manifest.
                         format: int64
+                        maximum: 4294967295.0
                         minimum: 0.0
                         type: integer
                       fileCount:
                         description: Number of files in the sealed artifact manifest.
                         format: int64
+                        maximum: 4294967295.0
                         minimum: 0.0
                         type: integer
                       nodeRank:
                         default: 0
                         description: Distributed node rank that owns this node-scoped artifact.
                         format: int64
+                        maximum: 4294967295.0
                         minimum: 0.0
                         type: integer
                       totalSize:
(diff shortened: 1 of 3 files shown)


kevin85421:
cc @yixinh-nv a simple follow up PR of #722. Thanks!

kevin85421:
CI failures seem to be unrelated.

yixinh-nv (approved):
LGTM. Schema bounds match the u32 contract, both generated manifests are in sync, and the rolling-update CI failure is a 600s rollout timeout unrelated to this change.

kevin85421:
cc @yixinh-nv would you mind enabling the CI again? Thanks!

kevin85421:
Thanks @yixinh-nv!

PR #747, opened 2026-09-11
feat(refit): support per-tensor FSDP wire dtypes

yixinh-nv:
The FSDP publisher currently stages every floating tensor in BF16. Models that retain selected tensors in FP32 therefore lose precision during publication, even when inference stores those tensors in FP32. Casting the received values back to FP32 cannot recover the lost bits.

This adds `FSDPTrainerContext(wire_dtype_overrides=...)`, a mapping from exact state-dict names to transfer dtypes. Ordinary tensors continue using BF16 by default. Selected tensors can use FP16, BF16, or FP32.

### Generic framework integration

The implementation contains no model-name checks or architecture-specific tensor rules. The calling framework supplies its model's precision policy, so the same API can handle routing biases, normalization parameters, scaling tensors, and other precision exceptions across FSDP/DTensor models. For example, PrimeRL can construct the mapping using its existing `keep_in_fp32_for_weight_transfer()` hook. The motivating case is GLM-5's FP32 router selection bias.

Model registration, tensor-layout compatibility, and destination installation still need their own validation. This API covers unquantized tensor dtypes; packed FP8/FP4 weights and their scaling metadata remain outside its scope.

### Behavior

- Allocate and register persistent staging buffers in each selected dtype.
- Keep per-tensor dtype and element size in manifests, and calculate total bytes from the served buffers.
- Apply the same policy to `IN_PLACE`, requiring the source dtype to match.
- Copy the override mapping when creating the adapter; reject unknown names, non-floating overrides, unsupported dtypes, and source dtype changes before overwriting an existing buffer.
- Preserve the default BF16 behavior for existing callers. No server/protobuf change is required.

### Validation

- 69 focused CPU tests passed across the FSDP adapter/publisher, trainer client, vLLM installer unit tests, and slice planner.
- New tests cover exact FP32 preservation, default BF16 casting, FP16 overrides, scalar/matrix tensors, manifest byte counts, warm buffer reuse, caller-map mutation, and invalid input.
- A CPU probe using a real 256-element GLM-5 router-bias tensor changed from 256 rounded values under BF16 staging (maximum absolute error 0.015540123) to zero changed values with the FP32 override, including a warm perturbed update. Only distributed-rank queries, the CUDA allocation scope, and registration were substituted; allocation dtype, tensor copies, and manifest construction used the actual adapter code.
- A local combination with #735 passed 72 focused tests after resolving the overlap in manifest byte accounting. This PR targets current `main` and does not require #735.
- Applicable pre-commit checks passed; Rust hooks skipped because no Rust files changed.

GPU/RDMA execution, installed full-model equality, generation agreement, and performance are pending. The CPU probe does not establish those results.

## Summary by CodeRabbit

- **New Features**
  - Added per-tensor transfer precision overrides for FSDP refit operations, supporting FP16, BF16, and FP32.
  - Transfer metadata now reports payload sizes based on each tensor’s selected precision.
  - Added validation for tensor names and supported floating-point conversions.

- **Documentation**
  - Added configuration guidance covering precision overrides, validation, buffer handling, manifest accounting, and conversion limitations.

- **Bug Fixes**
  - Improved handling of tensor buffers, scalar tensors, and source tensors whose precision changes between operations.


KavinKrishnan:
I rebased this onto current `main` (`e27f086`) and pushed the result to
`kavink/rebase-747-onto-main` in this repo, in case it saves you the merge. I
have not touched this PR's branch or opened a competing one — take it, ignore
it, or tell me to open it as a PR, whichever you prefer. Your authorship is
preserved on the commit.

All 23 tests in `test_refit_fsdp_adapter.py` pass on the rebased head.

Two conflicts, both from `main` gaining the manifest caching and timing work
after this branch was cut. The import hunk was mechanical. The second one was
not, so please check it:

`_staged()` on `main` computes `total_bytes` from a single
`wire_elsize = torch.empty((), dtype=WIRE_DTYPE).element_size()`. That
assumption is exactly what this PR removes, so I kept `main`'s cache-hit and
timing structure and substituted your per-shard computation inside it:

```python
total_bytes = sum(
    math.prod(s.local_shape) * s.served_tensor.element_size()
    for s in shards
)
```

My reasoning is that with per-tensor overrides the shards no longer share one
wire width, so a single element size would under-count a manifest containing
FP32 tensors. If you intended `total_bytes` to stay a BF16-equivalent figure
for the metric's sake, my resolution changes its meaning and should be reverted.

Context on why this matters to us: PrimeRL
[#3487](https://github.com/PrimeIntellect-ai/prime-rl/pull/3487) carries the
consumer side, and GLM-5 needs the FP32 router selection bias preserved on the
wire. That PR works against released ModelExpress because it names
`wire_dtype_overrides` only when the installed client accepts it, but GLM-5
itself stays blocked until this lands.

yixinh-nv:
@KavinKrishnan thanks, took it. I re-signed your rebased commit (identical tree, authorship and sign-off preserved) and force-pushed it to this PR's branch as `ecc218f8`. GitHub now reports the PR mergeable and `/ok to test` is posted for the new head.

Your `total_bytes` resolution is correct and is exactly what this branch computed before #735 landed. The manifest should report the bytes actually served, so with per-tensor overrides it has to be summed per shard from each served buffer's element size. A single BF16 element width would under-count any FP32 tensor. Nothing to revert.

KavinKrishnan:
@yixinh-nv agreed on `total_bytes`, and thanks for re-signing the rebase — nothing reverted there.

I pushed `804c9cb1` to this branch with the three remaining review asks, all cosmetic:

- `_create_trainer_adapter` now takes one construction path. The FSDP branch was early-returning with kwargs while Megatron fell through to the shared `adapter_type(...)` call; both now select an `adapter_type` and pass engine-specific kwargs to one call site.
- Deleted the orphan `"""Trainer engine selection."""` literal left below the real module docstring in `engines/__init__.py`.
- `FSDPTrainerContext.wire_dtype_overrides` defaults to `MappingProxyType({})` rather than `dict`, so the mutable default behind an immutable-looking `Mapping` on a frozen dataclass is now actually immutable. The adapter's `dict(... or {})` copy already made it safe; this makes it true at the declaration.

Two other asks from the review were already resolved on your rebased head, so I left them alone: the `classic_cuda_alloc` import no longer moves into the torch group (that churn is gone from the diff), and `served_tensor` is a property falling back to `source_tensor`, so `_stage()` can never read it before assignment.

129 focused tests pass locally (FSDP adapter and publisher, Megatron adapter and selection, fused wire, batch install, replica merge, slice plan, transfer, envs), and all 55 CI checks are green on the new head.

modelexpress_client/python/modelexpress_rl/train/context.py
```diff
@@ -14,6 +22,11 @@ class TrainerEngineContext:
 class FSDPTrainerContext(TrainerEngineContext):
     """Select FSDP/DTensor tensor capture and geometry."""
 
+    wire_dtype_overrides: Mapping[str, torch.dtype] = field(
+        default_factory=lambda: MappingProxyType({})
+    )
+    """Exact state-dict names to transfer as FP16, BF16 or FP32; other tensors use BF16."""
+
```
zhengluo-nv, line 30: Please make this an exhaustive wire-dtype contract instead of an override with an implicit BF16 fallback. Wire precision is part of the tensor representation; a typo, renamed key, or newly added floating tensor would otherwise silently publish as BF16 and can create numerically plausible divergence.

I suggest requiring `wire_dtypes: Mapping[str, torch.dtype]`, validating at bind time that its keys exactly match every floating state-dict entry, and using direct lookup rather than a default. Frameworks can still generate the complete map from their model-specific precision policy. Since this API is unreleased, this is the right time for the breaking change.

zhengluo-nv: We will do this in the following PRs. Merging to unblock progress

PR #748, opened 2026-09-11
feat(refit): add the NCCL M2N collective weight-transfer path

nicolasnoble:
Adds the NCCL M2N collective path for refit: a control plane in the server that brokers collective groups, epochs and bootstrap ids, and a Python data plane that reshards weights from a trainer mesh to a generator mesh.

This should supersede #659, #661, #665 and #666, which are cumulative branches, not a chain. Yixin's commits keep his authorship.

The client asks for the lane set it wants. MX allocates the ids, tracks membership and epochs, and does not check how the source side is split. The earlier revision worked the layout out from a source partition count and rejected any trainer count that was not a multiple of it, which I think duplicates a check the client already does, and turns away layouts that look valid.

Lane communicators are created non-blocking, and creating one puts the others back into ncclInProgress, where they will not accept collectives. The cache now polls the other live lanes back to success after each init, before the barrier that follows.

A transfer also has a deadline now. READY only means the group formed, so MX_NCCL_REFIT_TRANSFER_TIMEOUT_S bounds the reshard as well. On expiry the group is aborted and has to re-form at a new epoch.

The data plane calls nccl.m2n.reshard from the nccl-extensions distribution, published alongside nccl4py. It is an optional runtime dependency and I have left it out of pyproject.toml, for the same reason NIXL is not in there: it resolves per CUDA major through an extra, so pinning one picks the CUDA version for everyone who installs this client. Use nccl-extensions[cu12] or [cu13] to match your toolkit.

Reshard needs NCCL 2.30.7 or newer. Some CUDA images ship an older libnccl that gets loaded first, and the failure then happens inside the native call, so the guard checks the loaded library.

Testing: workspace suite and clippy clean. The collective service has integration coverage against real Redis and Lua through a live gRPC server. On four B200s, four ranks join through a live server and complete a 2x2 reshard byte-exact on both destination ranks. Two mismatch cohorts must fail for the passing one to count.

## Summary by CodeRabbit

* **New Features**
  * Added NCCL M2N collective refit support for coordinated trainer-to-generator weight transfers.
  * Added collective group membership, lane assignment, transfer tracking, readiness checks, epoch handling, and failure reporting.
  * Added Redis-backed control-plane coordination with idempotent transfer operations and worker registration expiry.
  * Added configurable streams, timeouts, polling, chunk sizes, and registration TTLs.

* **Documentation**
  * Documented architecture, deployment settings, installation requirements, and collective refit workflows.

* **Tests**
  * Added unit, GPU, and Redis-backed integration coverage for collective refit behavior.


modelexpress_client/python/modelexpress_rl/collective/comm.py
```diff
@@ -0,0 +1,432 @@
...
+
+            device_context = torch.cuda.device(device)
+
+        comm = None
+        try:
+            with device_context:
+                comm = communicator.Communicator.init(
+                    nranks=world_size,
+                    rank=rank,
+                    unique_id=_unique_id_from_bytes(unique_id),
```
yixinh-nv, line 359: The leader also rebuilds its id via `from_bytes`, but `ncclGetUniqueId` owns the bootstrap socket, so rank 0 never listens and peers get ECONNREFUSED. Keep the minted object for the leader.

modelexpress_client/python/modelexpress_rl/collective/client.py
```diff
@@ -0,0 +1,568 @@
...
+
+    def finish_weight_update(
+        self, version: str, operation_id: str | None = None
+    ) -> None:
+        half = self._require_round(version, "finish_weight_update")
+        try:
+            half.finish_weight_update(self.membership.broadcast_lane.lane_id)
+        except Exception as error:
+            self._report(operation_id, succeeded=False, message=repr(error))
+            half.abort()
```
yixinh-nv, line 443: If `_report` raises, `half.abort()` never runs. Abort first or use try/finally (same at line 539).

modelexpress_server/src/refit_collective/backend/redis/scripts/create_collective_transfer.lua
```diff
@@ -0,0 +1,50 @@
...
+
+local epoch = redis.call('HGET', KEYS[3], 'epoch')
+if not epoch then
+  return 'NOGROUP'
+end
+
+redis.call('HSET', KEYS[1],
+  'operation_id', ARGV[1],
+  'group_id', ARGV[2],
+  'epoch', epoch,
```
yixinh-nv, line 41: If the group epoch bumps after this, the op can never be reported or deleted, and the idempotency key (`SET` with no TTL) leaks forever. Let `delete` accept stale-epoch ops, or mark them `ABORTED`.

modelexpress_client/python/modelexpress_rl/collective/backend.py
```diff
@@ -0,0 +1,410 @@
...
+from .spi import LocalParamSpec, RefitCtx, resolve_specs
+from .types import ParamPlan, ReshardPlan
+
+logger = logging.getLogger("modelexpress_rl.collective.backend")
+
+DEFAULT_LAYER_GROUP = 0
+_M2N_CALL_LOCK = threading.Lock()
+
+
+def require_nccl_m2n() -> None:
```
yixinh-nv, line 35: This only checks the import; an older loaded `libnccl.so.2` still fails inside the native call. The `ncclGetVersion` >= 2.30.7 check from `test_collective_reshard.py` belongs here.

modelexpress_client/python/modelexpress_rl/collective/comm.py
nicolasnoble: i don't think this one is right, but tell me if i'm missing something. the bootstrap listener is a side effect of the `ncclGetUniqueId` call and is in the calling process. the python object we hand back doesn't own it, and there's no close or free for a `ncclUniqueId` in the api anyway. it's a 128 byte POD and we length check it in `rendezvous.py`.

`tests/gpu/test_collective_reshard.py` already exercises it: creates the id in the parent, passes only raw bytes to four spawned children, so every rank including rank 0 goes through `from_bytes`. green on real gpus. production should be easier, the process that made the id is rank 0 and stays up.

one thing i couldn't close: `nccl4py` isn't vendored here, so i only read `UniqueId.from_bytes` through its call sites. if you know it keeps state, say so and i'll redo this.

modelexpress_client/python/modelexpress_rl/collective/client.py
nicolasnoble: fixed in 4d79a3f9, both sites. abort first, since reporting is the network call and the abort isn't, so an unreachable control plane can't leave peers waiting in nccl. the report is wrapped now and the original failure is what propagates.

modelexpress_server/src/refit_collective/backend/redis/scripts/create_collective_transfer.lua
nicolasnoble: yep. the report script checks the epoch twice, once against the op and once against the group, so after a bump the new epoch fails one check and the old one fails the other. no value passes both.

one correction, delete checks whether the op is terminal and not its epoch, so letting it accept stale-epoch ops wouldn't reach it. `ABORTED` does work, and it's already treated as terminal in report, in delete and in the rust parser, with nothing writing it.

i'd rather keep a cross-language state machine change out of this pr, it's already large. tracking it as a follow-up, plus a ttl on the idempotency key separately.

modelexpress_client/python/modelexpress_rl/collective/backend.py
nicolasnoble: fixed in 49123c0d. `MIN_NCCL` and the version probe are in `backend.py` now and memoized, since `require_nccl_m2n` runs per parameter too. the gpu test and the bench harness import from there instead of each keeping a copy.

one thing i did on purpose: a probe that can't resolve the handle returns `None` and passes through. the import has already succeeded at that point, so failing there would ground the data plane on a loader layout we can't see through. tested both ways.

modelexpress_client/python/modelexpress_rl/collective/comm.py
yixinh-nv: You're right; I checked nccl4py, and the ID object doesn't own the listener, so my original diagnosis was incorrect.

modelexpress_client/python/modelexpress_rl/collective/client.py
yixinh-nv: Confirmed both paths now abort before reporting and preserve the original error, thanks.

modelexpress_server/src/refit_collective/backend/redis/scripts/create_collective_transfer.lua
yixinh-nv: Thanks for clarifying; please link the follow-up for stale-operation cleanup and the idempotency-key TTL.

modelexpress_client/python/modelexpress_rl/collective/backend.py
yixinh-nv: Confirmed the runtime version check is in place, with the unreadable-probe case intentionally left unchecked.

modelexpress_server/src/refit_collective/backend/redis/scripts/join_collective_group.lua
```diff
@@ -0,0 +1,260 @@
...
+    'plan_source_endpoint', '',
+    'plan_source_digest', '',
+    'created_at_unix_ms', ARGV[15])
+else
+  -- The first membership change from READY invalidates the old communicator
+  -- and opens a clean FORMING epoch. Further replacements in that same clean
+  -- epoch only acknowledge it; otherwise a full-cohort restart would bump once
+  -- per slot and no worker could share an epoch. Once any lane has published a
+  -- current-epoch bootstrap, another membership change must invalidate it.
+  membership_change_requires_bump = redis.call('HGET', KEYS[1], 'state') == 'READY'
```
ajcasagrande, line 106: [MEDIUM] A same-slot replacement admitted during a clean FORMING epoch evicts the original worker without bumping the epoch. The evicted worker's `await_ready` then still returns READY, and it enters the collective at the rank its replacement now holds.

**Mechanism** (all line-verified at `49123c0`):

1. **This line.** `membership_change_requires_bump` is `state == 'READY'`, else true only if some lane already carries `bootstrap_epoch == epoch`. In clean FORMING — no current-epoch bootstrap yet — it is **false**.
2. `:156-165` — a same-slot join with a different `worker_id` matches on role and ordinal, clears the plan source, then `if membership_change_requires_bump then changed = true end`. Under the false guard: **no bump**.
3. `:204-205` — the slot record is overwritten unconditionally: `HSET KEYS[2] ARGV[5] ARGV[6]..'|'..`. The original's participant record is gone; its registration key is still live and its waiter untouched.
4. `publish_group_bootstrap.lua:63-70` — readiness counts `HLEN` and walks `HVALS` of the participants hash: **current participants only**. The evicted worker is invisible to it, and the group reaches READY at the *same* epoch.
5. `rendezvous.py:595-598` — `if group.epoch != epoch: raise` / `if state == READY: return group`. The evicted worker's epoch still matches, so its wait succeeds.
6. `client.py:298-305` validates only lane-id coverage (`all_lane_ids - set(by_lane_id)`). `self._slot_id` is used at join (`:234`) and for the declared-lane mirror (`:245`) and **never again** — so the client initializes communicators from its stale membership, at the replacement's `rank_in_lane`.

I checked the two guards that could have rejected the replacement and neither fires: `DUPLICATE_WORKER` (`:171-180`) is scoped to *other* slots, and `CONFLICTING_ASSIGNMENT` (`:153-155`) compares role and ordinal only — both match by construction for a same-slot replacement.

**Window.** Every lane's rank 0 is a trainer (`lanes.rs:93`) and a leader publishes on join (`client.py:277,286`) before waiting (`:294`), so the window closes **when the first trainer joins**. Its width tracks leader arrival latency, not cohort size. In `run_e2e.sh` it is narrow — the generator's join sits behind full vLLM startup (`generator.py:57`→`:111`) while the trainer's is behind only a checkpoint load (`trainer.py:220-225`→`:258`), so trainers close it within seconds and the late generators are the *least* exposed cohort. The exposed case is a **delayed leader**: a slow checkpoint load, cold image pull, or rescheduled t0 holds the window open across every other join. Triggering it needs a replacement started while the original's registration is still live — K8s pod-replacement overlap in a termination grace period, or an autoscaler liveness false positive.

**Consequence, not overstated.** Most likely loud: a duplicate rank makes `Communicator.init` hang or fail, bounded by `MX_NCCL_REFIT_COMM_INIT_TIMEOUT_S` (300s), after which `_join_and_bootstrap`'s `except BaseException` aborts the group. I am not claiming silent corruption — the tail risk is a bootstrap that fails less cleanly.

Graded MEDIUM rather than higher because it needs a specific orchestrator race, the window is bounded, and the failure is loud. Higher than this review's LOWs because it is a live hole in the fencing contract rather than a lost diagnostic — and `docs/NCCL_M2N_REFIT.md:128` states that contract as "a membership change bumps the group epoch", which this membership change does not.

**Caveat: source-derived, not executed** — no Redis in my environment, so I could not reproduce it end to end. The control flow is quoted above so you can check the reading directly.

modelexpress_client/python/modelexpress_rl/collective/backend.py
```diff
@@ -0,0 +1,464 @@
...
+
+logger = logging.getLogger("modelexpress_rl.collective.backend")
+
+DEFAULT_LAYER_GROUP = 0
+_M2N_CALL_LOCK = threading.Lock()
+
+#: The reshard entry points were added in this NCCL release. An older library
+#: fails inside the native call rather than at import, so importability alone
+#: does not establish the runtime is usable.
+MIN_NCCL = (2, 30, 7)
```
ajcasagrande, line 39: [LOW] This floor is above what the repo's own committed lock provides, and the CPU-only unit tests silently depend on the host's libnccl.

`require_nccl_m2n()` (`:67-99`) imports `nccl.m2n` and then reads the version of the **real** `libnccl.so.2` mapped into the process (`loaded_nccl_version`, `:41-65`), raising below this floor. It is called from `_reshard` at `:123` — the ordinary data path the new CPU-only unit tests exercise. Those tests stub `nccl.m2n` into `sys.modules` (`test_collective_backend.py:60-64`) but do **not** stub `loaded_nccl_version`, so the host's libnccl decides whether they pass. That's at odds with what the modules say about themselves (`types.py`: "testable on a laptop"; `rendezvous.py`: "Torch-free and NCCL-free by construction").

Measured both ways at `49123c0`:

```
# through the committed lock (torch 2.10.0 -> nvidia-nccl-cu12 2.27.5)
$ uv run --frozen --extra dev python -m pytest tests/ -q
32 failed, 2083 passed, 34 skipped
  NcclUnavailableError: the libnccl this process loads is 2.27.5, and reshard needs 2.30.7

# the way CI installs (fresh resolve -> torch 2.14.0 / nvidia-nccl-cu13 2.30.7)
$ uv pip install -e ".[dev]" 'protobuf==5.27.2' && pytest tests/ -q
2115 passed, 34 skipped
```

**Graded LOW because CI is green** — `python-test` does `uv pip install -e ".[dev]"`, a fresh resolve that ignores the lock. What's left is that anyone running the suite through the repo's own lock (the natural `uv run` path, and what `.agents/skills/bump-version` maintains) sees 32 failures unrelated to their change; and the green resolve lands on **exactly 2.30.7**, the floor with zero margin, so one patch version of drift turns CI red for reasons unrelated to the code under test.

Cheapest fix: stub `loaded_nccl_version` in the unit tests. They already stub `nccl.m2n`, and `TestNcclVersionFloor` tests the probe directly, so no coverage is lost.

Note `uv.lock` is not in this PR's diff — the incompatibility comes from the new floor meeting a pre-existing lock, which is still this PR's to own or to note.

ajcasagrande (commented):
## Review

**Nothing here blocks merge.** One MEDIUM correctness hole I'd want tracked before this runs under a real orchestrator, three LOWs, two NITs.

Reviewed `b2ec150..49123c0` (52 files, +13,296) in a worktree at the PR head, not against the current branch. Note up front that this is 13k lines and I have one pass on it — see "What I did not check" at the bottom, which is load-bearing.

### Findings

| | Where | What |
|---|---|---|
| MEDIUM | `join_collective_group.lua:106` | A same-slot replacement in a clean FORMING epoch evicts a live worker without bumping the epoch; the evicted worker's `await_ready` still returns READY and it enters the collective at its replacement's rank |
| LOW | `backend.py:39` | `MIN_NCCL` floor is above what the repo's committed lock provides; the CPU-only unit tests silently depend on the host's libnccl |
| LOW | `service.rs:103` | A comma in a `slot_id` permanently wedges a group — the delimiter allow-list omits the one separator `encode_lanes` uses |
| LOW | `rendezvous.py:595` | `EpochChangedError` absorbs two conditions retrying cannot fix, and one makes this PR's own `disagreeing_slots` diagnostic unreachable client-side |
| NIT | `client.py:221` | Unused parameter |
| NIT | `NCCL_M2N_REFIT.md:239` | State diagram names an RPC and two states that do not exist |

### Verified correct — you don't need to defend these

Ran at `49123c0`:
- `cargo clippy --workspace --all-targets` — clean, zero warnings.
- `cargo test -p modelexpress-server --lib refit_collective` — 31 passed.
- Full Python suite in a CI-mirror env (`uv pip install -e ".[dev]"`, resolves torch 2.14.0 / nvidia-nccl-cu13 2.30.7) — **`2115 passed, 34 skipped`**.

Read and checked:
- **`LaneLayout::new` structural validation** — duplicate lane ids, unspecified kind, unknown slots, duplicate slot in a lane, slot on no lane, >1 broadcast lane, lane with no trainer at rank 0: each rejected with a named error rather than left to time out, each with a test.
- **`group_id_for`** — order-insensitive over the expected-slot lists and the lane vector, order-*sensitive* within a lane (that is the rank assignment), and separated across the role boundary so moving a name between roles cannot collide.
- **Lua KEYS arithmetic** in all five scripts (`#KEYS - 4`, `for i = 5, #KEYS`, `KEYS[4+i]`) matches what the Rust pushes in `join_group`, `publish_bootstrap`, `report_transfer`, `refresh_group_state`.
- **Terminal-state immutability** in `report_collective_transfer.lua` — a late failure cannot regress COMPLETE, a late success cannot erase FAILED, and a failure bumps the epoch and clears every lane in the same transaction.
- **`publish_bootstrap`'s TOCTOU is closed** — the rank-0 slot is picked outside the script, but the script re-verifies the leader's worker_id, joined epoch and live registration, and rank-0 identity is fixed by the immutable declared lane order.
- **Two-sided op ordering** in `backend.py` — `_all_bulk` is sorted by `canonical()` on both sides, so a trainer's partition-filtered subset on lane `p` is exactly the subsequence the generator issues there; layer grouping partitions that order without reordering it.
- `participant_from_record` rejects the retired 5-field format; `decode_lanes` rejects extra fields and non-numeric ids; `hex_encode`/`hex_decode` round-trip 128 bytes; `operation_idempotency_key` is length-prefixed so the concatenation is unambiguous.
- `server.rs` wiring is additive — `add_optional_service`, absent rather than degraded for non-Redis backends.

One correction to that list, since it matters more than the additions: I had originally cleared the epoch-bump semantics as "walked by hand ... behave as the comments claim". That walk confirmed each comment matched its code without asking whether the comment's *assumption* held, and the MEDIUM below is exactly where it doesn't. The clean-FORMING replacement case is removed from the cleared list.

### What I did not check — please weight the above accordingly

- **No Redis in my environment**, so `workspace-tests/tests/refit_collective_service_redis.rs` (887 lines) never ran. The MEDIUM and the `EpochChangedError` LOW are both **source-derived, not executed** — I've quoted the control flow in full in each so you can check my reading, but treat them as readings.
- **No GPU and no nccl-extensions runtime**, so `tests/gpu/`, the bench harness, and the whole example directory are unverified.
- Of the four areas I considered highest-risk — Lua epoch/readiness under concurrent joins, two-sided op ordering, rendezvous sequencing, and the Redis integration suite — the MEDIUM came out of the *first* one. I would not read the absence of further findings in the other three as evidence they're clean; it mostly reflects where I spent the pass.

modelexpress_server/src/refit_collective/service.rs
```diff
@@ -0,0 +1,393 @@
...
+    for (field, slots) in [
+        ("spec.expected_trainer_slots", &spec.expected_trainer_slots),
+        (
+            "spec.expected_generator_slots",
+            &spec.expected_generator_slots,
+        ),
+    ] {
+        for slot in slots {
+            required(slot, field)?;
+            delimiter_free(slot, field, &['\0', '\n', '\r', '|'])?;
```
ajcasagrande, line 103: [LOW] This allow-list omits the one separator `encode_lanes` actually uses, so a comma in a `slot_id` permanently wedges the group.

`encode_lanes` (`redis.rs:216-233`) stores the lane set as `lane_id|kind|trainers.join(",")|generators.join(",")`, and its doc comment asserts "Slot ids in this path are role-ordinal strings, so none of the three separators can occur inside one." Three of them are enforced here; `,` is not. `split_csv` (`redis.rs:282-287`) splits on it.

Reproduction of the exact encode/decode pair:

```
stored lanes field     = "0|1|t,0|g0"
decode_lanes trainers  = ["t", "0"]
expected_trainer_slots = ["t,0"]     # stored newline-joined, survives intact
```

`read_group_once` (`redis.rs:363-368`) then calls `LaneLayout::new(decode_lanes(...), ...)`, which returns `LaneError::UnknownSlot` for `"t"` → `Internal`. `join_collective_group` never decodes, so **the join succeeds and the group is written to Redis**; every subsequent `GetCollectiveGroup`, `PublishGroupBootstrap` and `ReportCollectiveTransfer` for that group then fails `Internal`, with no recovery short of manual Redis surgery.

**Deliberately graded LOW on reachability**: every caller in this PR builds slot ids as `f"t{i}"`/`f"g{i}"`, and no CLI flag, config file or dataset field produces a comma. It is reachable only by an integrator choosing its own naming — which the proto invites ("Logical, version-scoped contribution identifier"). Flagged as a one-character addition to a validation list that already exists, not as a live defect.

modelexpress_client/python/modelexpress_rl/collective/rendezvous.py
```diff
@@ -0,0 +1,673 @@
...
+                    raise GroupNotReadyError(
+                        group_id,
+                        _missing_slots(group) if group is not None else [],
+                        timeout_s,
+                    ) from error
+                time.sleep(
+                    max(0.0, min(poll_interval_s, deadline - time.monotonic()))
+                )
+                continue
+            if group.epoch != epoch:
```
ajcasagrande, line 595: [LOW] This check fires before anything else is inspected, so `EpochChangedError` — whose message is "the cached communicator and plan must be rebuilt", i.e. *retry* — absorbs two conditions where retrying is futile.

**(a) A real plan-digest disagreement.** `join_collective_group.lua:185-187` bumps the epoch on *every* join whose digest differs from the group's:

```lua
if redis.call('HGET', KEYS[1], 'plan_digest') ~= ARGV[9] then
  changed = true
end
```

A cohort that genuinely disagrees therefore churns the epoch on each join, and every waiter lands here. The self-defeating part: `_missing_slots` formats exactly this case at `:643-646` (`"plan digest disagreement on slot {slot}"`), but `_missing_slots` only runs on the **timeout** path, which the epoch bump pre-empts. The server populates `disagreeing_slots` correctly and the client's own waiter can never surface it — while `refit_collective.proto:129-133` says the field "exists so that a cohort which disagrees is diagnosable instead of merely stuck in FORMING with no field to read".

Corroborating, though not proof: `disagreeing_slots` and that formatter have **no test anywhere in the repo** (`grep -rn disagreeing_slots` returns only the two production read sites and three write sites), while `_missing_slots`' other branches *are* covered (`test_collective_rendezvous.py:470-483`, `:551`). An untested feature isn't automatically broken, but the gap is specific to this branch.

**(b) A registration lapsing during FORMING.** `refresh_collective_group.lua:36-52` bumps the epoch whenever it sweeps an expired registration, unguarded — while `join_collective_group.lua:106-115` deliberately guards the identical sweep, and says why ("a full-cohort restart would bump once per slot and no worker could share an epoch"). `refresh` runs on the **read** path: `read_group()` calls it on every `GetCollectiveGroup`, which `await_ready` polls every 0.25s per waiter.

LOW because the rendezvous fails either way — what's lost is the diagnosis, and in (a) specifically a field this PR added for that purpose. Nothing hangs; the deadline still bounds it.

**Caveat: source-derived, not executed** — no Redis available, so neither path was reproduced end to end.

modelexpress_client/python/modelexpress_rl/collective/client.py
```diff
@@ -0,0 +1,584 @@
...
+            LaneDeclaration(
+                self._source_partition_count,
+                "BROADCAST",
+                tuple(self._trainer_slots),
+                tuple(self._generator_slots),
+            )
+        )
+        return lanes
+
+    def _join_and_bootstrap(
```
ajcasagrande, line 221: [NIT] `source_partition` is never referenced in this method's body — confirmed zero occurrences between the signature and the next method. Both call sites (`:425`, `:524`) pass it.

docs/NCCL_M2N_REFIT.md
```diff
@@ -0,0 +1,493 @@
...
+### 5.4 States
+
+```mermaid
+stateDiagram-v2
+  direction LR
+  state "CollectiveGroup" as G {
+    [*] --> FORMING
+    FORMING --> READY: all expected slots joined,<br/>every lane bootstrapped,<br/>every registration live
+    READY --> FORMING: membership change (epoch += 1)
+    READY --> RELEASING: DeleteCollectiveGroup
```
ajcasagrande, line 239: [NIT] This diagram names an RPC and two states the implementation does not have.

- **No `DeleteCollectiveGroup` RPC exists.** `refit_collective.proto` declares seven RPCs on `RefitCollectiveService`, and this document's own RPC listing ~20 lines below doesn't include it either. `docs/ARCHITECTURE.md`, added in this same PR, states flatly: "There is no group-delete RPC, symmetrically with workers never creating one."
- **Nothing ever writes `RELEASING` or `ABORTED`** (`:247-248`). Every Lua script writes only `'READY'` or `'FORMING'`; grepping the scripts for those two strings returns comparisons only, no assignment. The transfer deadline is client-side and reports `succeeded=false`, which the Lua records as **FAILED**.

Graded NIT because this document is banner-marked "**Status: design.**" at line 7, so forward-looking content is legitimate. Worth a line anyway because the reference doc shipped in the same PR contradicts it, leaving a reader two documents and no way to tell which is current.

PR #753, opened 2026-09-11
perf(refit): batch initial layout capture

yixinh-nv:
## Problem and change

Layout capture currently invokes the engine's `load_weights` once per converted source tensor. Loaders that build model-wide parameter dictionaries and expert mappings on entry repeat that setup thousands of times during the first refit.

Pass all lazy weights through one loader call. Each lazy tensor retains its original published-source identity and conversion/view chain. If an `UnsupportedReshard` interrupts the bulk attempt, discard its newly recorded copies and unattributed-write count, then retry per source to preserve precise failure diagnostics. Unexpected engine errors still propagate, and loader hooks are restored on every exit.

This is a first-capture optimization based directly on main. It is independent of bounded staging in #749 and does not cache layouts across refits.

## Validation

- 149 focused Python tests passed: geometry, MoE/fused reconstruction, slice/transfer planning, vLLM installer/adapter, and generator/runtime.
- New regressions cover one loader initialization, unsupported failures at the beginning/middle/end without duplicate records, converted-source attribution, and hook restoration on unexpected errors.
- Pre-commit, Ruff, and `git diff --check` passed. Rust hooks skipped because there are no Rust changes.
- CPU/meta benchmark: 4,096 sources, 512 destination parameters, PyTorch 2.14.0, three alternating-order samples per arm. Both arms produced identical capture results.

| Arm | Loader calls | Raw capture seconds | Median |
| --- | ---: | --- | ---: |
| Previous per-source path | 4,096 | 0.669285, 0.667772, 0.674371 | 0.669285 s |
| Bulk capture | 1 | 0.007415, 0.007917, 0.007520 | 0.007520 s |

The approximately 89x improvement is specific to this synthetic model-wide-lookup benchmark. It does not establish vLLM, GPU-transfer, or end-to-end refit performance. An unsupported bulk capture incurs an extra attempt before the existing diagnostic fallback.

Reproduce from `modelexpress_client/python` with the package dependencies installed:

```bash
PYTHONPATH=. python benchmarks/bench_refit_capture.py --sources 4096 --parameters 512 --repeats 3
```

## Results

- All **128 receiver ranks** completed their initial geometry capture, each recording **3,279 copies, zero unsupported sources, and zero unattributed writes**.
- Capture completion appeared **28–68 seconds after the orchestrator logged that all inference engines were paused**.
- In the previous run with per-source capture, only four receiver ranks recorded capture completion before the failed attempt ended; those observations were **795–1,038 seconds after the same pause milestone**, with the same copy and unsupported/unattributed counts.

These are wall-clock log milestones that include discovery, scheduling and preparation before capture; the runs also differ by the separate integration fixes, so this is **not an isolated capture A/B timing or an end-to-end refit speedup**. The prior attempt failed later during installation, and the latest observation establishes capture completion only. 

## Summary by CodeRabbit

- **Performance**
  - Improved refit resharding geometry capture by processing compatible weight sources in a single batch, reducing repeated loading work.

- **Reliability**
  - Added automatic fallback to individual source processing when batch operations are unsupported.
  - Preserved detailed diagnostics identifying affected sources while allowing unexpected errors to surface normally.

- **Documentation**
  - Documented bulk geometry capture behavior and fallback handling.


modelexpress_client/python/modelexpress/refit/reshard/geometry.py
```diff
@@ -324,18 +324,23 @@ def capture_weights(
...
-                model.load_weights([(name, tensor)])
-            except UnsupportedReshard as exc:
-                unsupported.append(source)
-                unsupported_reasons[source] = str(exc)
+        # Large MoE loaders rebuild model-wide parameter/expert maps on entry.
+        # LazyWeights carry their own source identity across one bulk invocation.
+        copies_start, unattributed_start = len(recorder.copies), recorder.unattributed
+        try:
+            model.load_weights(list(weights.items()))
+        except UnsupportedReshard:
```
zhengluo-nv: Do we know when will this fail and throw UnsupportedReshard?

yixinh-nv: UnsupportedReshard is raised for untraceable loader operations such as LazyWeight arithmetic, and the per-source retry identifies the affected tensors.

KavinKrishnan:
@yixinh-nv I pushed `7166daab` to this branch addressing the four rollback findings from the landing review. All 55 CI checks are green on the new head and the geometry suite is 26 passing.

**Recorder state ownership.** The call site was reaching into the recorder to truncate `recorder.copies` and restore `recorder.unattributed`, which means a third accumulator added to `_BakeRecorder` later would be silently missed. That moved into `_BakeRecorder.attempt()`, a context manager that snapshots and restores `copies`, `unattributed` *and* `current` on any exception, so the invariant lives with the state it protects.

**The bulk exception was discarded.** This was the one with teeth. `except UnsupportedReshard:` bound nothing, so if the bulk attempt failed and the per-source retry then passed for every source, `unsupported` stayed empty and `capture_weights` returned a plan as though nothing had happened — masking exactly the order-dependent loader incompatibility that only shows up in bulk mode. The cause is now logged at warning before the retry runs.

**Stamp restoration.** You were right that this needed confirming rather than changing. The retry does run against pristine stamps: `_install_stamps` replaces each `param.weight_loader` once and every stamp resets `recorder.current` in its own `finally`, so a failed bulk attempt leaves no stamp state behind. `attempt()` now restores `current` explicitly as well, and the new test asserts every loader is its original object on exit.

**The lost invariant.** "It never produces an incorrect partial plan" is restated in both the module design note and `docs/ARCHITECTURE.md`, with the reason it still holds — the interrupted attempt rolls back every accumulator, so the retry starts from exactly the state the bulk attempt began with.

New test `test_bulk_only_failure_is_reported_even_when_every_source_retries_clean` covers the masking case directly: a model that raises `UnsupportedReshard` only on multi-weight calls, asserting the plan is complete, `unsupported` is empty, the bulk cause is in the log, and the loaders are restored.

zhengluo-nv (approved):

PR #760, opened 2026-09-14
docs(examples): add vLLM MTP manifest

AndyDai-nv:
## Summary
- add a single-GPU Qwen3.5-0.8B MTP manifest following the existing vLLM P2P example
- document the source-to-target scale-up workflow and clarify that only target-model weights use P2P
- link the new manifest from the Kubernetes P2P example README

## Validation
- parsed both YAML documents with PyYAML
- rendered with `kubectl apply --dry-run=client --validate=false`
- passed the repository pre-commit hooks
- validated Qwen3.5-0.8B on two H200 nodes: 273 tensors / 1.64 GB transferred over NIXL RDMA in 0.051s; target inference succeeded with MTP enabled

## Summary by CodeRabbit

- **New Features**
  - Added a Kubernetes deployment option for serving Qwen3.5-0.8B with vLLM.
  - Enabled ModelExpress peer-to-peer weight transfer and speculative decoding.
  - Added an HTTP service on port 8000 with GPU, RDMA, shared memory, and model-cache configuration.
  - Documented the new deployment option in the Kubernetes examples.


examples/p2p_transfer_k8s/client/vllm/vllm-mtp.yaml
```diff
@@ -0,0 +1,144 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Single-GPU vLLM deployment with ModelExpress P2P weight transfer and MTP.
+# Uses Qwen3.5-0.8B so the full source -> target flow fits on two GPUs.
```
zhengluo-nv, line 5: Can we add a larger model? Since GLM 5.2 MTP is fixed, can we add that as an example

AndyDai-nv: Updated the example to use the official `zai-org/GLM-5.2-FP8` checkpoint. The manifest now requests 8 GPUs per replica with TP=8, FP8 KV cache, and 5 MTP speculative tokens, following the official vLLM recipe. I also updated the resource prerequisites and README entry.

zhengluo-nv (approved):

PR #767, opened 2026-09-15
refactor(refit): receive generator P2P directly into live tensors

### LINKED ISSUE #765: Refit P2P stages a full second copy of model weights, requiring ~2x model size of free GPU memory
## Problem

Refit over P2P allocates a full second copy of the model's runtime tensors as NIXL receive buffers, so a replica needs roughly 2x the model size in free GPU memory before a peer refit can succeed without hitting OOM. 

This is the refit path only. Cold-start P2P loads into the model's own storage and is unaffected.

zhengluo-nv:
## Summary

- retain the exact donor read lease during staging without moving tensor bytes
- receive generator-to-generator NIXL data directly into the registered live post-PWAL tensors at the engine safe point
- remove the second model-sized GPU receive buffer and the subsequent device-to-device copy
- preserve trainer and object-storage staging behavior

Fixes #765.

## Failure semantics

Source fallback remains safe before live mutation begins. A failure before NIXL submission republishes the unchanged serving version. Once the direct receive starts, a transfer failure marks the update as having potentially mutated the engine; the worker must be reset or restarted instead of falling through with mixed weights.

## Testing

- focused post-review regression suite: `120 passed, 7 skipped`
- broader refit/reshard unit suite: `639 passed, 1 skipped`
- `pre-commit run --all-files`
- manual nscale Dynamo + vLLM E2E with two TP1 workers:
  - active refit passed for `v1 -> v2 -> v3`
  - exercised S3 reconstruction for the first worker and generator P2P through `RuntimeTensorNixlUpdateMethod` for the second worker
  - inference succeeded after each installed version
  - a replacement worker cold-started at desired `v3` through P2P
  - observed replacement transfer: 1.20 GB in 0.049 s (197.3 Gbps); this is an observed functional-run measurement, not a performance benchmark

Qualification is currently limited to two TP1 workers. This run does not qualify TP>1, multinode behavior, numerical parity, or production performance.

## Dependency

Builds on the donor read-lease lifecycle merged in #757.

## Summary by CodeRabbit

* **New Features**
  * Runtime weight refits can transfer data directly into live model tensors, reducing staging overhead.
  * Refits now report transfer and installation performance metrics.
  * Preparation can reserve peer resources or reconstruct checkpoints before applying changes.

* **Bug Fixes**
  * Added stricter validation for peer transfer metadata, tensor names, sizes, and data types.
  * Improved cleanup of transfer resources after failures.
  * Runtime state is fenced when a refit fails after live data mutation.

* **Documentation**
  * Clarified refit preparation, transfer behavior, supported configurations, and recovery requirements.


modelexpress_client/python/modelexpress_rl/inference/nixl_staged_transfer.py
```diff
@@ -643,75 +685,67 @@ def stage_peer(
...
+        )
+        host, port, remote_agent_name = self._peer_endpoint(manifest)
+        try:
+            self._manager.fetch_remote_and_wait(
+                remote_agent_name=remote_agent_name,
+                ip=host,
+                port=port,
+                timeout_seconds=self._timeout,
+            )
+            on_transfer_start()
```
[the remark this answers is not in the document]
zhengluo-nv: Addressed in 6331b40d. The callback now runs inside `NixlTransferManager.receive_from_source()` immediately before NIXL submission, after setup and descriptor preparation. A setup failure therefore leaves `installation_started` false; the caller also republishes the unchanged serving version after such a pre-transfer failure. Regression tests cover both boundaries.

zhengluo-nv:
Addressed both findings from the latest CodeRabbit review in `6331b40d`:

- `on_transfer_start` now runs immediately before NIXL submission, so descriptor/setup failures do not mark the live engine as mutated.
- `apply_weight` captures the current serving version and republishes it when an update fails before transfer submission. Once submission starts, the runtime remains unpublished and the engine is fenced as uncertain.

Added regression coverage for both boundaries. The focused suite passes with `120 passed, 7 skipped`, and `pre-commit run --all-files` passes.

modelexpress_client/python/modelexpress_rl/inference/methods/runtime_tensor.py
```diff
@@ -44,33 +62,70 @@ def capabilities(self) -> MethodCapabilities:
...
+        try:
+            read.metrics.update(
+                self._transfer.receive_peer(
+                    tensor_read=read.tensor_read,
+                    destination_tensors=read.tensors,
+                    on_transfer_start=read.mark_transfer_started,
+                )
+            )
+        finally:
+            read.tensor_read.close()
```
anson627: Can we invalidate the handle on failure or reacquire and validate the lease before retrying? A pre-transfer failure closes the lease but leaves _active_read reusable, and retrying apply_weight(staged) can read donor memory without a valid lease and accept another version’s weights

anson627 (approved):

modelexpress_client/python/modelexpress_rl/inference/methods/runtime_tensor.py
zhengluo-nv: Will update

zhengluo-nv: Fixed in 2044182b. Every consumed TensorReadLease is now cleared. A retry after a pre-transfer failure reacquires and revalidates the same exact mx_source_id and worker_id before NIXL runs; if transfer submission already started, the staged read is not retryable and the engine remains fenced. Added regressions for both boundaries. The focused parent suite passes 83 tests, and the combined parent plus stacked-child suite passes 120 tests with 7 skips.

nv-hwoo (approved):

PR #770, opened 2026-09-15
fix(client): skip zero-byte tensors in NIXL transfers

### LINKED ISSUE #762: Kimi INT4 P2P loading falls back to S3: zero-byte tensor descriptors cause NIXL_ERR_NOT_FOUND
Kimi-K2.7-Code replica startup selects RDMA, fetches the source manifest, then fails in `prepXferDlist` with `NIXL_ERR_NOT_FOUND` on every TP rank and falls back to S3. The manifest contains zero-byte quantization metadata tensors whose null addresses have no registered NIXL storage.

### Environment

- Two separate nodes, each with 8 NVIDIA B200 GPUs, on the same InfiniBand fabric; TP=8 per replica.
- Model: `moonshotai/Kimi-K2.7-Code`, revision `74797c9c62378b951a1f6fcf5c4631024e9b8bef`, native INT4 snapshot (~595 GB).
- Runtime: Dynamo 1.3.0, vLLM 0.23.0, ModelExpress 0.5.1.
- Installed `nixl_transfer.py` matches upstream commit `e27f08685b2a545dac4a905d068abe241ae75e82` (apart from a trailing newline).

### Two-node reproduction

1. Mirror the full pinned checkpoint to an S3 prefix. Configure ModelExpress server/Redis discovery and working RDMA connectivity between the two nodes.
2. Start one worker with the following loader settings (substitute the S3 URI and server address):

```bash
export VLLM_PLUGINS=modelexpress
export MX_INSTANT_TENSOR=0
export MX_MODEL_URI=s3://BUCKET/PREFIX
export MX_SERVER_ADDRESS=SERVER:8000
export MODEL_EXPRESS_URL="$MX_SERVER_ADDRESS"
export MX_RDMA_NIC_PIN=auto
export MX_METRICS_ENABLED=1
export MX_TRANSFER_TIMEOUT=1800
export UCX_RNDV_SCHEME=get_zcopy
export UCX_RNDV_THRESH=0

python3 -m dynamo.vllm \
  --model moonshotai/Kimi-K2.7-Code \
  --revision 74797c9c62378b951a1f6fcf5c4631024e9b8bef \
  --served-model-name moonshotai/Kimi-K2.7-Code \
  --load-format modelexpress \
  --model-loader-extra-config '{"memory_limit":8589934592,"concurrency":8}' \
  --tensor-parallel-size 8 --trust-remote-code --dtype bfloat16 \
  --gpu-memory-utilization 0.70 --max-model-len 4096
```

3. Wait for all eight source ranks to finish S3 loading, publish P2P metadata, and become Ready. We allocated 64 CPU cores and 256 GiB host RAM per worker; the bounded loader memory avoids an unrelated S3 staging OOM.
4. Start an identically configured second worker on the other node, sharing the same ModelExpress discovery service and Dynamo namespace.
5. Observe all eight receiver ranks selecting RDMA, then failing before transfer and falling back to ModelStreamer.

Representative output (agent suffix omitted):

```text
prepXferDlist: failed to prepare the descriptors for any of the specified or potential backends for agent 'mx-auto-worker3-...'
[Worker 3] Strategy rdma failed, trying next: RDMA receive failed: NIXL_ERR_NOT_FOUND
```

Each rank's source manifest contained 1,975 tensors and 76,573,250,156 bytes, including 240 empty `g_idx`/sort-index metadata tensors (`addr=0`, `size=0`). Registration skips empty storage, but `receive_from_source()` still creates transfer descriptors for these tensors.

### Small regression reproduction

This CPU-only adapter test reproduces the descriptor construction bug without loading Kimi or requiring RDMA hardware. Run it in an environment with the ModelExpress Python client, torch, and pytest installed. The mock rejects zero-byte ranges at the NIXL preparation boundary; the actual hardware failure above was observed separately.

```python
"""NIXL transfer contract for quantized models with empty metadata tensors."""

from types import SimpleNamespace
from unittest.mock import Mock
import pytest
import torch
from modelexpress.nixl_transfer import NixlTransferManager, ManifestMismatchError


def transfer(tensors, sizes=None):
    manager = NixlTransferManager(agent_name="empty-test", device_id=0)
    manager._accelerator_backend = Mock(nixl_mem_type="VRAM")
    agent = Mock()
    agent.check_xfer_state.return_value = "DONE"
    descriptors = []

    def prepare(**kwargs):
        values = kwargs["xfer_list"]
        assert all(size > 0 for _, size, _ in values), "NIXL rejects zero-byte ranges"
        descriptors.append(values)
        return object()

    agent.prep_xfer_dlist.side_effect = prepare
    manager._agent = agent
    manager._tensors = tensors
    source = [
        SimpleNamespace(
            name=name,
            addr=t.data_ptr(),
            size=(sizes or {}).get(name, t.numel() * t.element_size()),
            dtype=str(t.dtype),
            device_id=0,
        )
        for name, t in tensors.items()
    ]
    result = manager.receive_from_source(
        b"", source, remote_agent_name="source", require_exact_match=True
    )
    return result, descriptors


def test_empty_metadata_preserves_count_without_zero_byte_nixl_ranges():
    (nbytes, count, _), descriptors = transfer(
        {"weight": torch.ones(4), "indices": torch.empty(0, dtype=torch.int32)}
    )
    assert (nbytes, count) == (16, 2)
    assert len(descriptors) == 2
    assert all(len(d) == 1 for d in descriptors)


def test_all_empty_manifest_needs_no_nixl_transfer():
    result, descriptors = transfer({"indices": torch.empty(0, dtype=torch.int32)})
    assert result == (0, 1, 0.0)
    assert descriptors == []


def test_empty_tensor_still_rejects_source_size_mismatch():
    with pytest.raises(ManifestMismatchError, match="size mismatch"):
        transfer({"indices": torch.empty(0, dtype=torch.int32)}, {"indices": 4})
```

Run `pytest -q test_empty_tensor_transfer.py`. Unpatched source: **2 failed, 1 passed**. With the change below: **3 passed**.

nv-hwoo:
## Problem

Quantized models such as Kimi include empty metadata tensors. Registration skips these tensors, but the receiver still builds zero-byte NIXL descriptors for them. NIXL cannot resolve their unregistered addresses, causing `NIXL_ERR_NOT_FOUND` and fallback to S3.

Fixes #762.

## Fix

Validate and count empty tensors before excluding them from NIXL transfer descriptors. Preserve strict manifest checks and receive metrics.

| Case | Result |
|---|---|
| Mixed empty and nonempty tensors | Count all matches; transfer only nonempty tensors |
| All matched tensors are empty | Return `(0, matched_tensors, 0.0)` without preparing a transfer |
| Invalid manifest | Preserve existing mismatch errors |

## Where to start

- [`receive_from_source()`](https://github.com/ai-dynamo/modelexpress/blob/c813b8458dd84cf85c9c21169e6244e6bb631ce9/modelexpress_client/python/modelexpress/nixl_transfer.py#L815): review validation ordering, the separate match count, and the all-empty return.
- [`test_pool_registration.py`](https://github.com/ai-dynamo/modelexpress/blob/c813b8458dd84cf85c9c21169e6244e6bb631ce9/modelexpress_client/python/tests/test_pool_registration.py): regression coverage for mixed/all-empty tensors, mismatches, transfer indices, and outcome metrics.

## Test and Validation

- Focused CPU suite covering transfer, registration, manifest validation, and peer lifecycle: **160 passed**.
- Issue #762's CPU reproducer: **3 passed**, previously 2 failed and 1 passed.
- Applicable pre-commit hooks on changed files: **passed**.
- Two-node Kimi/RDMA reproduction: **not run**; it requires the full checkpoint and a two-node, TP=8 GPU deployment. Local tests use mocked NIXL.


nv-hwoo:
Closing since https://github.com/ai-dynamo/modelexpress/pull/769 includes the fix

Issue #782, opened 2026-09-17
P2P refit leaves host KV-cache scales stale despite matching GPU tensors


anson627:
## Bug and evidence

P2P refit updates GPU KV-cache scales but leaves cached host floats and CPU mirrors stale for `nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4`.

Tested with two B200 workers (one GPU each, TP=1), native vLLM + ModelExpress generator APIs, without runtime patches:

- Before refit: **0/18** host/GPU scale mismatches on either worker.
- After S3 refit: **0/18** mismatches on the donor.
- After P2P refit: **11/18** mismatches on the receiver, although **all 761 GPU tensor hashes match the donor**.
- Both workers report the new serving version; weight addresses are preserved. No OOM.

Example: `model.layers.5.mixer.attn._k_scale`:

| State | GPU tensor | `_k_scale_float` / `_k_scale_cpu` |
|---|---:|---:|
| Receiver before refit | 0.03173828125 | 0.03173828125 |
| Donor after S3 refit | 0.0316685251891613 | 0.0316685251891613 |
| Receiver after P2P refit | 0.0316685251891613 | **0.03173828125** |

The consistency assertion stops the test before post-refit resume/inference; output-quality impact was not measured.

## Reproduction

Environment: ModelExpress 0.7.0, vLLM 0.27.0, PyTorch 2.13.0+cu130. Model revision: `bee7596271d1495f6992ae224aefde4410e816b8`.

1. Load the model on a donor from object storage, then load a receiver through ModelExpress P2P. Use TP=1, max model length 4096, max sequences 8, GPU memory utilization 0.70. Verify matching GPU tensor hashes and host scales initially.
2. Stop serving on both workers. Publish an embedding-only XOR delta (~64 MiB payload) for `backbone.embeddings.weight`; retain checkpoint quantization scales unchanged.
3. Initialize `ModelExpressGeneratorClient` with `source_order=(WeightSource.OBJECT_STORAGE,)` on the donor and `(WeightSource.GENERATOR,)` on the receiver, with no object-storage fallback on the receiver.
4. Call `stage_weight(version=WeightVersionRef(new_version))`, then `apply_weight(staged)` on the donor; repeat for the same version on the receiver.
5. Synchronize CUDA, compare all runtime GPU tensor hashes, and run this audit inside each worker against the live model:

```python
for name, module in model.named_modules():
[...]

PR #783, opened 2026-09-17
fix(vllm): refresh host attention scales after eager RL peer refit

anson627:
P2P RL refits overwrite runtime attention-scale tensors while leaving vLLM's Python scalars and CPU mirrors stale. This change extracts the refresh from #711 into a shared helper and uses it after eager-mode runtime tensor installation.

The eager refit path updates q/k/v host mirrors in place and invalidates FlashInfer's cached BMM and output-scale scalars together so the next forward recomputes them. Cold loading retains its existing validation and cold-cache requirement. Peer-refit eligibility remains as introduced by #778: quantization, FP8 KV-cache dtype, and execution mode do not gate peer selection. The new host-scale refresh runs only with `enforce_eager=True`; graph-mode refits retain the existing direct-copy behavior without this refresh. Captured scalar updates and graph recapture remain outside this fix.

Related to #782.

### Draft limitation: retained quantized KV cache

Tracked separately in #784 for a follow-up PR. This PR is scoped to applying the host-mirror refresh from #711 to eager RL refits.

This change does not clear stored KV entries or change the user's keep/clear policy. The invalidated FlashInfer caches are launch scalars, not KV token storage. However, it does not yet enforce compatibility when an async RL transition retains quantized KV entries and changes their K/V scales. Those entries would still be encoded with the previous scales. Supporting that case requires keeping the scales fixed while entries are retained or rejecting an incompatible transition before mutation, consistently across S3 and P2P. CUDA graph recapture is outside this change's scope.

### Validation

- 73 targeted cold-load, RL installer, and runtime eligibility tests passed.
- 117 surrounding runtime-transfer, generator, and weight-transfer tests passed; local gRPC tests required localhost binding outside the sandbox.
- Applicable pre-commit checks and `git diff --check` passed. Rust hooks skipped because there are no Rust changes.
- Regression coverage includes repeated eager refits, per-head scales, storage preservation, invalid scale state, destination validation, and quantized peer eligibility in both eager and graph modes.
- B200/vLLM end-to-end validation confirmed stale host-scale issue #782 is fixed for peer refit with model_config.enforce_eager=True. All scales match, and resumed inference passes

## Summary by CodeRabbit

* **New Features**
  * Improved warm model refits for quantized models and FP8 KV caches when eager execution is enabled.
  * Host-side attention scales now refresh automatically after runtime weight transfers, with related cached values invalidated for consistency.

* **Bug Fixes**
  * Added validation to reject invalid scale states before engine updates.
  * Preserved live tensor storage during warm refits.

* **Documentation**
  * Updated architecture and refit guidance to describe eager-mode requirements and fallback behavior.


modelexpress_client/python/modelexpress_rl/inference/engines/vllm/__init__.py
```diff
@@ -10,6 +10,21 @@
 from .context import VllmGeneratorContext
 
 
+def _supports_runtime_tensor_p2p(vllm_config: object) -> bool:
```
nv-hwoo: i think this quantization check came back by mistake

anson627: nice catch! fixed

modelexpress_client/python/modelexpress/engines/vllm/host_quantization.py
```diff
@@ -0,0 +1,187 @@
...
+                    module_label,
+                    float_name,
+                    previous,
+                    value,
+                )
+            setattr(module, float_name, value)
+
+            cpu_mirror = getattr(module, f"_{scale_name}_scale_cpu", None)
+            if isinstance(cpu_mirror, torch.Tensor):
+                cpu_mirror.fill_(value)
```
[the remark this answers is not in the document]
nv-hwoo: Flushing the caches will be up to the RL orchestrator

nv-hwoo:
/ok to test [04803b5](https://github.com/ai-dynamo/modelexpress/pull/783/commits/04803b52e024d836b2766249a66a388e275802d6)

nv-hwoo (approved):

PR #788, opened 2026-09-18
feat(server): terminate TLS on the gRPC listener

wseaton:
Previously the gRPC listener served plaintext only, two issues: 
- ServiceAccount bearer tokens crossed the pod network in the clear w/o a service mesh technology in the mix
- and a clusters in regulated environments that mandates a TLS profile for every TLS server had no way to run one

Adds a `tls` config section and flags (`--tls-cert-file`, `--tls-key-file`, `--tls-min-version`, `--tls-cipher-suites`, `--tls-groups`). 

tonic's own TLS is rustls-only, so a build linking OpenSSL cannot use it; so instead the listener is wrapped and handed to tonic as a stream of negotiated connections. The handshake comes from the build's backend, OpenSSL (`tls-openssl`) or rustls with ring (`tls-rustls`, the default).

Settings normalized to use OpenSSL spelling on both backends. Minimum version accepts `TLS1.2` and `VersionTLS12`, and a mixed 1.2/1.3 cipher list configures the two generations independently.

The follow up PR for Client TLS, Helm chart changes, and the end-to-end tests is #789.

Tests: `--lib tls::` on both backends (24 rustls, 13 OpenSSL), `modelexpress-common --lib tls::` (7), `modelexpress-server --lib config::` (32). fmt and clippy clean across the TLS feature matrix.


wseaton:
addressed coderabbits concern about fail open TLS configuration, also changed configuration errors to be typed

